PR #5, opened 2020-11-18
Updating the readme.md to add subscribeProxy

jherr:
It's not a huge use case, but a complete solution has non-React subscriptions to the store, so it's good to show that it's covered. :)


drcmda:
it's subscribe though, not subscribeProxy. it was changed last minute. i will add this to the svg as well.

PR #8, opened 2020-11-19
vanilla support

### LINKED ISSUE #6: web components and esm
hey hey! i've been looking for a mobx replacement because it's so bloated: and valtio looks really elegant and promising!

unfortunately, i can't consider valtio because my apps are esm-only and use true web components — no react — i'm currently using `lit-element`

are there any plans to decouple valtio from react, so it can be used in any application, like mobx?

to alleviate the incompatibility, in addition to the react-decoupling work, the esm distribution would need to be modified to include `.js` extensions on all non-bare-specifier esm imports (so that library can work natively in browsers), eg:
- `import { createMutableSource, useMutableSource } from './useMutableSource'`  
    &nbsp; &nbsp; *becomes*
- `import { createMutableSource, useMutableSource } from './useMutableSource.js'`

:wave: chase

dai-shi:
close #6 

Now, we should be able to do:
```js
import { proxy, subscribe, snapshot } from 'valtio/vanilla'
```

@chase-moskal 
Not sure how this works natively in browser. It still depends on proxy-compare and it needs to be adjusted too?

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 5094,
-    "minified": 2272,
-    "gzipped": 973,
+    "bundled": 5109,
+    "minified": 2263,
+    "gzipped": 968,
     "treeshaked": {
       "rollup": {
         "code": 61,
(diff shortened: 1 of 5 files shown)


dai-shi:
It should work if one use bundlers anyway. Merging this.

PR #9, opened 2020-11-19
fix: array length for direct assignment

### LINKED ISSUE #7: Assigning directly to an array by index does not change array length
## Expected behavior

When assigning directly to an array position, the array length should be updated, a la the native JavaScript array behavior. Any components depending on the array (for example, iterating over its items) should be re-rendered.

**Example of index assignment with a native Array**

```
const arr = [];
arr[0] = 'foo';
arr.length; // 1
arr[6] = 'bar';
arr.length; // 7
```

## Observed behavior

When assigning to an array position on a proxy store, the length is not updated, and dependent components do not re-render. However, an update notification is sent to `subscribe`.

Reproduction: https://codesandbox.io/s/valtio-array-length-bug-kl8d2?file=/src/App.js

Uncomment line 44 to "fix" the problem manually by assigning the correct array length. This triggers the component re-render, too.

Observe also that the `subscribe` callback, when it logs the current proxy state, shows the new third item present but the length is still `2`.

a-type:
Fixes #7 : array length not updating on index assignment.

It appears that creating an empty array using `Object.create` from the prototype of the original doesn't create a proper array, just an Object which inherits from the Array prototype. Real arrays have special behaviors.

See also https://stackoverflow.com/questions/22022058/new-array-vs-object-createarray-prototype

The added tests check for the special indexing properties of an Array: a simple case of adding a sequential item, and an edge case of adding a non-contiguous one.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,7 +37,10 @@ const createProxy = <T extends object>(initialObject: T = {} as T): T => {
       listeners.forEach((listener) => listener(nextVersion))
     }
   }
-  const proxy = new Proxy(Object.create(initialObject.constructor.prototype), {
+  const emptyCopy = Array.isArray(initialObject)
+    ? []
+    : Object.create(initialObject.constructor.prototype)
+  const proxy = new Proxy(emptyCopy, {
     get(target, prop, receiver) {
       if (prop === MUTABLE_SOURCE) {
         if (!mutableSource) {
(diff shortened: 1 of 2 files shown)


a-type:
Apologies, new laptop - didn't set up my Git properly. May push a forced update shortly to properly do the commit.

dai-shi (approved):
Thanks for working on this!
I read the SO, and this should be only the way to make a new array properly. A bit unfortunate JS, but can't be helped. Good to go.

PR #12, opened 2020-11-20
Update readme with suspense

theomjones:
Awesome package!

I know the readme image is pretty but it's not accessible and quite hard to read compared to a traditional readme, maybe it could be put to the bottom as the alternate view?

Anyway, here's an update to alt text!


dai-shi (approved):
thaaaanks!

PR #15, opened 2020-11-20
support undefined constructor

### LINKED ISSUE #11: How can I "get" current state?
Really enjoying this library! I came across an issue when I need to add extra data to a state (and persist previous one) object i.e.

```ts
assetState.textures = { ...assetState.textures, ...sheet.textures };
```

Doing something like this results in the following error

```
Uncaught TypeError: Cannot read property 'prototype' of undefined
    at proxy (index.js:46)
    at Object.set (index.js:125)
    at index.js:137
    at Array.forEach (<anonymous>)
    at proxy (index.js:136)
    at Object.set (index.js:125)
    at index.js:137
    at Array.forEach (<anonymous>)
    at proxy (index.js:136)
    at Object.set (index.js:125)
```

Error points to this line of code

```ts
const emptyCopy = Array.isArray(initialObject) ? [] : Object.create(initialObject.constructor.prototype);
```

I log out `assetState` prior to that line that throws errors and it seems it is created correctly, I see things like `[[Handler]]` etc..

dai-shi:
close #11.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 4953,
-    "minified": 2186,
-    "gzipped": 944,
+    "bundled": 4973,
+    "minified": 2188,
+    "gzipped": 941,
     "treeshaked": {
       "rollup": {
         "code": 61,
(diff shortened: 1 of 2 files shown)


dai-shi:
https://codesandbox.io/s/sweet-resonance-xz5pd?file=/src/App.js
@IljaDaderko Fixed your last codesandbox example. I'm not sure if your whole use case with pixi.js works or works with good performance, though. Thanks for your reproduction anyway. It's fundamental.

IljaDaderko:
Awesome! Thank you and yeh, I already re-wrote it since there is no need for all of this to be in a state, nevertheless at least it helped to make valtio a bit more "tanky" 😄

dai-shi:
OK, let's merge this then.

dai-shi:
Published: https://www.npmjs.com/package/valtio/v/0.3.1

PR #16, opened 2020-11-20
feat: freeze snapshot

dai-shi:
Snapshots are shared in the tree, so it shouldn't be mutated.
As states are OK to mutate, it's sometimes confusing.
This PR freeze snapshots to avoid unexpected behavior.

Unfortunately error message is something simple like "TypeError", which can't be modified.


Noitidart:
Couple questions:

1. Doesn't `Object.freeze` have performance overhead? I thought because of this people usually only do this in dev.
2. I am taking a snapshot, to just remove the `proxy` and then I need to mutate it, but I'm getting errors thrown (haven't tested in prod). Is it possible to opt out of freezing in dev too?

Noitidart:
Got answer on discord. Have to use `_.cloneDeep`, `snapshot` is more performant because it's read-only.

And SO shows no difference in perf - https://stackoverflow.com/questions/8435080/any-performance-benefit-to-locking-down-javascript-objects

PR #19, opened 2020-11-21
feat: batch updates

dai-shi:
This batches several updates into one. This changes the update behavior in timing.


dai-shi:
This might raise other issues, but let's see how it goes. Please file an issue if anyone finds any.

Issue #20, opened 2020-11-21
In a subscription handler, if state is touched (but not changed) it leads to an infinite loop.


wilcoschoneveld:
Previous discussions: https://github.com/pmndrs/valtio/pull/13
Related test: https://github.com/pmndrs/valtio/blob/master/tests/subscribe.test.tsx#L53

I'm thinking about a similar kind of API like seen in [immer](https://github.com/immerjs/immer):

```
subscribe(state, draft => {
    // mutate draft here
})
```

you could check if the `draft` was modified after the handler execution and determine if (and which) listeners should be called.

(Another thought: a similar technique might be possible in React components with the `useProxy` hook, allowing the batching of `draft` updates by calling listeners _after_ rendering components (instead of waiting for the next eventloop tick with `Promise.resolve()`), but I don't yet see how this could be implemented)

Aslemammad:
@wilcoschoneveld  Could you reproduce the problem in codesandbox?

dai-shi:
I think it's nothing related with `subscribe`. (Besides, I'm a bit hesitant to change the api. It's basically from the useMutableSource spec.)

```js
const state = proxy({ count: 0, obj: {} })

state.count = 0 // this is expected not to trigger updates, correct?

state.obj = {} // can we trigger updates because it's a new object?
```

@wilcoschoneveld Do you get the idea? Let's check if it's updated with `Object.is`.

wilcoschoneveld:
> I'm a bit hesitant to change the api. It's basically from the useMutableSource spec.

I understand!

> ```js
> const state = proxy({ count: 0, obj: {} })
> 
> state.count = 0 // this is expected not to trigger updates, correct?
> 
> state.obj = {} // can we trigger updates because it's a new object?
> ```

Yes that makes sense!

PR #23, opened 2020-11-22
skip setting property if not changed

dai-shi:
https://github.com/pmndrs/valtio/issues/20#issuecomment-731644189

The fix is rather simple.

@wilcoschoneveld How does it look?

### DIFF
--- a/tests/subscribe.test.tsx
+++ b/tests/subscribe.test.tsx
@@ -47,7 +47,6 @@ describe('subscribe', () => {
 
   it('should not re-run subscription if no change', async () => {
     const obj = proxy({ count: 0 })
-
     const handler = jest.fn()
 
     subscribe(obj, handler)
--- a/tests/subscribe.test.tsx
+++ b/tests/subscribe.test.tsx
@@ -45,8 +45,10 @@ describe('subscribe', () => {
     expect(() => subscribe(obj.count, handler)).toThrow()
   })
 
-  it.skip('should not re-run subscription if no change', async () => {
+  it('should not re-run subscription if no change', async () => {
     const obj = proxy({ count: 0 })
+    await Promise.resolve()
+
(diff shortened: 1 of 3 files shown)


tests/subscribe.test.tsx
```diff
@@ -45,8 +45,10 @@ describe('subscribe', () => {
     expect(() => subscribe(obj.count, handler)).toThrow()
   })
 
-  it.skip('should not re-run subscription if no change', async () => {
+  it('should not re-run subscription if no change', async () => {
     const obj = proxy({ count: 0 })
+    await Promise.resolve()
```
wilcoschoneveld, line 50: I would like to discuss this line, right now the `subscribe` handler is called immediately if you subscribe in the same tick as creating the proxy, but it is not called immediately if you subscribe in the next tick. I feel like either behaviour is fine as long as it's consistent. What do you think?

wilcoschoneveld: Here is an example of a situation where the subscribe handler is called and it might even be desired in this case:
https://codesandbox.io/s/react-forked-p9dy0?file=/src/App.js

However if by default the subscribe handles are not called directly the user could manually call it if desired.

dai-shi: > I feel like either behaviour is fine as long as it's consistent. What do you think?

Yeah, that would be less confusing.

When I added this L50 in the test, I thought it's unfortunate too.
I mean we should avoid notifying updates in `proxy()`. Maybe, adding `initializing` flag would help.

dai-shi: c049b42 should be better.

wilcoschoneveld: looks good!

wilcoschoneveld: thanks!

PR #24, opened 2020-11-23
unsupported map/set, improve delete property

dai-shi:
Map/Set are not supported, but they should be at least usable as non tracking objects.

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -4,8 +4,8 @@ const VERSION = Symbol()
 const LISTENERS = Symbol()
 const SNAPSHOT = Symbol()
 
-const isObject = (x: unknown): x is object =>
-  typeof x === 'object' && x !== null
+const isSupportedObject = (x: unknown): x is object =>
+  typeof x === 'object' && x !== null && !(x as any)[Symbol.iterator]
 
 const proxyCache = new WeakMap<object, object>()
 let globalVersion = 0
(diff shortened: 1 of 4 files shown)


src/vanilla.ts
```diff
@@ -4,8 +4,8 @@ const VERSION = Symbol()
 const LISTENERS = Symbol()
 const SNAPSHOT = Symbol()
 
-const isObject = (x: unknown): x is object =>
-  typeof x === 'object' && x !== null
+const isSupportedObject = (x: unknown): x is object =>
+  typeof x === 'object' && x !== null && !(x as any)[Symbol.iterator]
```
dai-shi, line 8: Hm, this doesn't detect `WeakMap`...

dai-shi: Oh, and array has an iterator...

dai-shi:
This is not perfect (as `WeakMap` is not detected), but we have a larger issue (proxy-compare only supports plain objects). Meanwhile, this could be an intermediate solution.

tests/basic.test.tsx
```diff
@@ -27,7 +27,7 @@ it('simple counter', async () => {
   await findByText('count: 1')
 })
 
-it('no extra re-renders', async () => {
+it.only('no extra re-renders', async () => {
```
dai-shi, line 30: 🤦‍♂️

PR #25, opened 2020-11-23
fix: snapshots have to be plain objects

dai-shi:
continuing from #24 

As of now, proxy-compare does only support plain objects and arrays, we can't use prototypes.
This is to work around by copying keys.

I'm not confident that test cases cover everything. Help wanted.

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -7,7 +7,10 @@ const SNAPSHOT = Symbol()
 const isSupportedObject = (x: unknown): x is object =>
   typeof x === 'object' &&
   x !== null &&
-  (Array.isArray(x) || !(x as any)[Symbol.iterator])
+  (Array.isArray(x) || !(x as any)[Symbol.iterator]) &&
+  !(x instanceof WeakMap) &&
+  !(x instanceof WeakSet)
+// XXX all unsupported objects are not listed here. is there a better way?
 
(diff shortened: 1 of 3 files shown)


dai-shi:
`Class TicTac` still seems to be working. https://codesandbox.io/s/valtio-tic-tac-forked-p14i1?file=/src/index.js

(Note: This example doesn't do well for re-renders. Modified version: https://codesandbox.io/s/valtio-tic-tac-forked-nwp42?file=/src/index.js )

src/vanilla.ts
```diff
@@ -50,10 +53,21 @@ export const proxy = <T extends object>(initialObject: T = {} as T): T => {
...
-          ? []
-          : Object.create(target.constructor?.prototype || null)
+        const snapshot: any = Array.isArray(target) ? [] : {}
         snapshotCache.set(receiver, { version, snapshot })
+        Reflect.ownKeys(Object.getPrototypeOf(target) || {}).forEach((key) => {
+          if (key === Symbol.unscopables) return
+          try {
+            const value = target[key]
+            // if (typeof value === 'function') {
+            //   snapshot[key] = value.bind(target)
```
dai-shi, line 63: I couldn't make a test case that would require this bind.

dai-shi:
I still don't think it's optimal and there must be various edge cases, but let's merge and release this. Feedbacks are welcome.

PR #26, opened 2020-11-23
Improve non-proxy error

Aslemammad:
@dai-shi, I think we should write tests for errors, What's your idea?

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -129,9 +129,22 @@ export const proxy = <T extends object>(initialObject: T = {} as T): T => {
   return p
 }
 
-export const getVersion = (p: any): number => p[VERSION]
+function assertProxy(p: any) {
+  if (typeof process === 'object' && process.env.NODE_ENV === 'production')
+    return
+
+  if (typeof p !== 'object' || !p[VERSION]) {
+    throw new Error('Please use proxy object')
+  }
+}
+
+export const getVersion = (p: any): number => {
+  assertProxy(p)
+  return p[VERSION]
+}
 
 export const subscribe = (p: any, callback: () => void) => {
+  assertProxy(p)
   let pendingVersion = 0
   const listener = (nextVersion: number) => {
     pendingVersion = nextVersion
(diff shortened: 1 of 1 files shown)


dai-shi:
> I think we should write tests for errors, What's your idea?

It's definitely desirable.

Aslemammad:
> > I think we should write tests for errors, What's your idea?
> 
> It's definitely desirable.

I'll work on it in an another PR.

Aslemammad:
@dai-shi How is it now?

src/vanilla.ts
```diff
@@ -4,6 +4,9 @@ const VERSION = Symbol()
 const LISTENERS = Symbol()
 const SNAPSHOT = Symbol()
 
+const isProduction =
+  typeof process === 'object' && process.env.NODE_ENV === 'production'
```
dai-shi, line 8:
```suggestion
  typeof process !== 'object' || process.env.NODE_ENV === 'production'
```
You failed to invert the logic. 😄

dai-shi (changes requested):
As far as I confirm with https://xem.github.io/terser-online/,
`isProduction` fails to minify the production code.
Please make it inline too.

src/vanilla.ts
Aslemammad: Yea, I forgot. thanks.

Aslemammad:
@dai-shi I made isProduction inline now.

dai-shi (approved):
Thanks for adjustment!

Aslemammad:
@dai-shi Thanks.

dai-shi:
What would happen if one does `useProxy({})`?
We'd want to invoke `getVersion()` at the beginning. (and we need another `if`. 🤷‍♂️ )

Aslemammad:
jest fail tests: Error: Please use proxy object Jest in useProxy

dai-shi:
That's good. Let's merge this in.
Thanks for your contribution!

Aslemammad:
Thanks for the merging.

PR #27, opened 2020-11-23
simplify snapshot without copying prototype props

dai-shi:
After releasing v0.4.1, I'm a bit uncomfortable because it seems pretty hard to maintain all those edge cases.
This will drop some use cases to use prototypes in snapshots.
The rationale behind it is we would not track function call results.


tests/class.test.tsx
```diff
@@ -141,94 +141,3 @@ it('inherited class without methods', async () => {
...
-  }
-
-  const obj = proxy(new CountClass())
-
-  const Counter: React.FC = () => {
-    const snapshot = useProxy(obj)
-    return (
-      <>
-        <div>count: {snapshot.count}</div>
-        <div>doubled: {snapshot.doubled()}</div>
```
dai-shi, line 163: This looks nice, but as snapshot.doubled wouldn't change, it will not trigger re-renders.
The only reason this example works is `snapshot.count` is used in the previous line.

dai-shi:
Tic Tac example
https://codesandbox.io/s/valtio-tic-tac-forked-tfbcy?file=/src/index.js
It's not very fancy, but internally it makes sense, and I expect it's less fragile.

dalcib:
> After releasing v0.4.1, I'm a bit uncomfortable because it seems pretty hard to maintain all those edge cases.
> This will drop some use cases to use prototypes in snapshots.
> The rationale behind it is we would not track function call results.

I understand that it introduces some complexities in the code, but with v0.4.1, the Tic Tac example with getters ( https://codesandbox.io/s/valtio-tic-tac-forked-p14i1?file=/src/index.js) was working like a charm, rerendering only the square that was clicked. Now, with this commit, the previous behaviour came back, and all squares rerender in every click.

I think Computed properties or Getters are a big win for a state management package, and one of the reasons for Mobx success, so it would be good to support it out of the box.

The Tic Tac valtio refactor to use subscribe (https://codesandbox.io/s/valtio-tic-tac-forked-nwp42) needs four abstractions (store, computed, subscribe and actions) instead of just one (a class with getters). Sometimes we need to access properties and getters in the same component, which leads to the situation to need two `useProxy`, one for the `store` and the other for the `computed`, which can cause some confusion. It makes the code harder to maintain, because you need to remember which property comes from the `store` and which one comes from the `computed`. 

So, it would be great if it would be possible to keep the features introduced in v0.4.1

dai-shi:
I wasn’t aware 1.4.1 solved rerender issue, which wasn’t the case in 1.4.0, correct?

Honestly, I didn’t think getters work as expected. Let me investigate it again.

dalcib:
Exactly.

v0.4.1 was working ok (just rerender the square clicked, not the others).
Verified with React Dev Tools.
OK (Rendering the clicked):
rrone
Not OK (Rerendering all):
rrall
v0.2.3 Not OK❌:  https://codesandbox.io/s/valtio-tic-tac-j0nut
v0.4.1 OK✔:  https://codesandbox.io/s/valtio-tic-tac-041-i9zqb
#27 Not OK❌: https://codesandbox.io/s/valtio-tic-tac-forked-tfbcy

dai-shi:
https://codesandbox.io/s/valtio-tic-tac-040-rv1nl
Confirmed this is not OK in v0.4.0.

This is crazy, so nice! I didn't expect this behavior while making v0.4.1 (I know usage tracking didn't work and wanted to fix it.)
So, the getters or functions that read `this` read from proxies (by `createDeepProxy`) and the usage is tracked. This is not technically computed (like the result is not memoized), but it should cover many use cases in React.
I will add some more tests to check this behavior.

dai-shi:
Closing this as #29 will be a proper fix.

PR #29, opened 2020-11-24
support snapshot with prototype

dai-shi:
As discussed in #27, I need to fix it with keeping the behavior.
v0.4.1 is reported to fail in some use cases.
This should fix both v0.4.0 and v0.4.1 issues.

@a-type @dalcib

----
This allows following patterns.

```js
const state = proxy({
  count1: 1,
  doubled1() {
    return this.count1 * 2;
  },
}

class CountClass {
  count2 = 1;
  count3 = 1;
  doubled2() {
    return this.count2 * 2;
  }
  get doubled3() {
    return this.count3 * 2;
  }
}

const instance = proxy(new CountClass())
```


dalcib:
I converted a real (small) application from `mobx` to `valtio` using this build.
The changes requested were minimal ([commit](https://github.com/cefiti/cefiti/commit/8fb8f47c74b35c037eb576401c28ddea4491e2a6)). The code now is cleaner than before and everything is working well. 
The application has two classes as the state ([store](https://github.com/cefiti/cefiti/blob/valtio/src/store.ts) and [uiStore](https://github.com/cefiti/cefiti/blob/valtio/src/uistore.ts)) with getters and methods.

The only extra change needed is the following:
In `mobx` I was using the handler methods as properties to avoid the bind in React:
```javascript
//mobx-store.js
class store {
    ...
    handleChanges = (e) => {this.prop = e.target.value;}
}

//component.js
const comp = () => <input value={store.prop} onChange={store.handleChanges} >
```
It didn't work with `valtio`. It generates `this` is `undefined` inside the method. I had to change the code to:
```javascript
//valtio-store.js
class store {
    ...
    handleChanges(e) { this.prop = e.target.value;}
}

//component.js
const comp = () => <input value={snapshot.prop} onChange={(e) => store.handleChanges(e)} >
```

Even with this change, the result is very good, with a plus of reducing the bundle size.

dai-shi:
@dalcib That's great news. I'm not sure how to support class fields with arrow functions. Let's revisit it later. It might be difficult, though.

I'd also like to get a feedback from @a-type if possible.

dai-shi:
https://codesandbox.io/s/valtio-tic-tac-issue29-b47p0?file=/src/index.js
You already tried it, but here's the same tic tac example.

a-type:
Sorry I've been a bit distracted lately with work. I'll take a look at the code soon and maybe build and link it to my project to see how that goes.

Is the idea to support custom classes for direct read/write of properties including within invocation of methods? Trying to catch up on the context.

dai-shi:
@a-type 
Thanks for your updates. Don't worry. I'm sure this is better than v0.4.1 and you said your app didn't work v0.4.1, so I will merge it later and release it. Please take your time and try the released version.

> Is the idea to support custom classes for direct read/write of properties including within invocation of methods?

The idea is to support class methods (or object methods) for direct **read** of properties with `this`. This works for getters too. (Not sure what you mean by "within invocation of methods")

I will also try to create a small examples to illustrate the usage.

a-type:
I was curious if a method which writes to a property would trigger a change (if it didn't already - I guess that might already work).

I'm all for more magic!

a-type:
I was curious if a method which writes to a property would trigger a change (if it didn't already - I guess that might already work).

I'm all for more magic!

dai-shi:
> I was curious if a method which writes to a property would trigger a change

useProxy returns a snapshot which is an immutable object. So, invoking a snapshot method doesn't sound nice. Besides, we shouldn't mutate state in render.

If you mean something like this:
```js
const snapshot = useProxy(state)
useEffect(() => {
  snapshot.increment()
})
```
If you don't use `this`, it's probably possible.
```js
const state = proxy({
  count: 0,
  increment() {
    state.count++ // works
    this.count++ // doesn't work if it's from a snapshot
  }
})
```
To avoid confusion, we should recommend mutating from state method, instead of snapshot method:
```js
useEffect(() => {
  // snapshot.increment() // not recommended
  state.increment() // good
})
```

dai-shi:
https://codesandbox.io/s/react-typescript-forked-ns15i?file=/src/App.tsx
What works:
- object method
- class method
- class get method

object get method doesn't work because we don't copy the getter itself. (should we support this in later PRs?)

dalcib:
Excellent. Thanks.
By the way, the setters in Classes are also working.
https://codesandbox.io/s/react-typescript-forked-gno5d?file=/src/App.tsx
The setters are not working in Objects, as expected.

I think it would be good in the future to have getters and setters working on Objects, with the same behaviour as pure javascript, maintaining parity between Classes and Objects.

dai-shi:
Yes, setters in classes work for `state` aka proxy (but not for `snapshot`). I think it worked in previous versions too.

btw, if we could copy the object getters in proxy as functions, we can evaluate them when creating snapshot, which become essentially computed (only evaluated once.) This will be a very interesting characteristic, technically. not sure if developer friendly.

PR #35, opened 2020-11-30
fix: object getters returns snapshot undefined

dai-shi:
Issue reported on Twitter: https://twitter.com/dai_shi/status/1333202394809700354

> There seems to be a bug in handling object getters. If we return non-objects, it's not undefined.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 6546,
-    "minified": 3224,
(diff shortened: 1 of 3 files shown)


itaditya:
It's working! https://codesandbox.io/s/user-todos-with-valtio-check-fix-g7nuu?file=/src/App.js

Thanks @dai-shi

dai-shi:
Thanks for checking. I will merge and release.

PR #37, opened 2020-12-01
add useLocalProxy react hook

steve8708:
via https://codesandbox.io/s/hopeful-merkle-ziu1h?file=/src/App.js:105-278

only change I made was added TS typings and allow the argument to be a function in case intializing the state object is non trivial and you don't want that running every render when only needed on first render

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -101,7 +101,7 @@ subscribe(state, () => {
 
 ##### Use it locally in components
 
-You can use it locally in components.
+You can use it locally in components. [Note](./src/utils.ts#L5-L15)
 
 ```jsx
 import { useLocalProxy } from 'valtio/utils'
--- a/readme.md
+++ b/readme.md
@@ -101,7 +101,7 @@ subscribe(state, () => {
 
 ##### Use it locally in components
 
-You can use it locally in components. [Note](./src/utils.ts#L5-L15)
+You can use it locally in components. [Notes](./src/utils.ts#L5-L15)
 
 ```jsx
 import { useLocalProxy } from 'valtio/utils'
--- a/readme.md
+++ b/readme.md
@@ -98,3 +98,14 @@ subscribe(state, () => {
   const obj = snapshot(state) // A snapshot is an immutable object
 })
 ```
+
+##### Use it locally in components
+
+You can use it locally in components.
+
+```jsx
+import { useLocalProxy } from 'valtio/utils'
+
+function Foo() {
+  const [snapshot, state] = useLocalProxy({ count: 0, text: 'hello' })
+```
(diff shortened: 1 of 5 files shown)


dai-shi:
Hi, I was thinking about adding useLocalProxy too. But, not in the core. I would put it in utils because we want to keep the core api minimal and it's just a wrapper function around the core.

```js
import { useLocalProxy } from 'valtio/utils';
```

I can merge this, and continue the work on it on my end, if you are ok with it.

steve8708:
Thanks @dai-shi - makes sense and that works for me, and happy to update JSX Lite to use `import { useLocalProxy } from 'valtio/utils';` or whatever the final API is when ready 👍

dai-shi:
Sounds good. I think I can add commits to this branch. Let me try.

src/index.ts
```diff
@@ -53,4 +53,13 @@ const useProxy = <T extends object>(p: T): T => {
   return createDeepProxy(currSnapshot, affected, proxyCache)
 }
 
-export { proxy, subscribe, snapshot, useProxy }
+const useLocalProxy = <T extends object>(init: T | (() => T)) => {
+  const ref = useRef<T>()
+  if (!ref.current) {
+    const initialObject = typeof init === 'function' ? (init as any)() : init
```
Aslemammad, line 59: Just for the suggestion, Can we use `(init as () => T)` instead of `(init as any)`?

Aslemammad (changes requested):

src/index.ts
steve8708: ah yes ofc - I think @dai-shi is going to take over this PR but went ahead and threw that update in bc I agree `any` is ugly

dai-shi:
Created a simple example: https://codesandbox.io/s/react-typescript-forked-qye5u?file=/src/App.tsx

readme.md
```diff
@@ -98,3 +98,14 @@ subscribe(state, () => {
   const obj = snapshot(state) // A snapshot is an immutable object
 })
 ```
+
+##### Use it locally in components
+
+You can use it locally in components. [Note](./src/utils.ts#L5-L15)
+
```
Aslemammad, line 105: Now it's great, but can't we put the notes here, I don't have any problem with this, but just a question.

Aslemammad (commented):
I feel great about it, I put a comment in the readme.

readme.md
dai-shi: I just feel a bit long..

Aslemammad: That's ok, no problem.

dai-shi:
Let's merge this in.

dai-shi:
https://www.npmjs.com/package/valtio/v/0.4.5

Issue #45, opened 2020-12-10
How can I use controlled Components with valtio?


thelinuxlich:
With `useState` you can have a controlled component with no extra worries, but with valtio proxied variables, if I map the `state.variable` to the `value` property of the component and the `snapshot.variable` to the `onChange` event, weird behavior happens(textareas, for example, when you type the cursor drops to the end of the text on every change)

dai-shi:
Right, just yesterday, I noticed this behavior in jotai and know valtio is the same. Jotai is more severe because it tries to replace useState usage, and will fix it soon.

For valtio, I'm not sure what to do. This is causing with (async) batching mutations. Maybe, add a new function to trigger updates in sync, but it's hardly a nice solution. Hm, that's almost the same as forceUpdate. (and it doesn't work, OK, I need to fix something. will work on it.)

dai-shi:
OK, here's the workaround.
https://codesandbox.io/s/suspicious-darwin-dv99h?file=/src/App.js

I guess this is the best for now. It's unfortunate to have such a pitfall, though.

thelinuxlich:
Might be worth looking how react-easy-state does this, it uses proxies too and they fixed the caret positioning issue https://github.com/RisingStack/react-easy-state/pull/144

dai-shi:
Yeah, exporting batch is another option.
I'm a bit concerned about it though. In addition to exporting another function, we'd need almost always require `batch()` and it'd become looks like immer's `produce()`. I really think the default behavior should be batching automatically.

(which could lead, valtio is best fit with react-hook-form. 🤔 )

dai-shi:
Just to mitigate the forceUpdate hack, we could provide a util. (This is still unfortunate.)
```js
import { proxy, useProxy } from 'valtio'
import { useSyncUpdate } from 'valtio/utils'

const state = proxy({ text: 'hello' })

const TextBox = () => {
  const snapshot = useProxy(state)
  const update = useSyncUpdate()
  const onChange = (e) => {
    update(() => {
      state.text = e.target.value
    })
  }
  return <input value={snapshot.text} onChange={onChange} />
}
```

but there might be another way if we create a new util.

thelinuxlich:
at least developer experience wise this is better :)

thelinuxlich:
[...]

PR #46, opened 2020-12-11
fix: render from parents or something with uMS emulation hack

dai-shi:
ref #45 

This allows `forceUpdate` workaround.

Example:
```js
const state = proxy({ text: "hello" });

const TextBox = () => {
  const [, forceUpdate] = useReducer((c) => c + 1, 0);
  const snapshot = useProxy(state);
  const onChange = (e) => {
    state.text = e.target.value;
    forceUpdate();
  };
  return <input value={snapshot.text} onChange={onChange} />;
};
```
https://codesandbox.io/s/suspicious-darwin-dv99h

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,7 +19,7 @@ const useProxy = <T extends object>(p: T): T => {
   useEffect(() => {
     lastAffected.current = affected
   })
-  const getChangedSnapshot = useMemo(() => {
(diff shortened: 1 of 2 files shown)


dai-shi:
I will try another run for #45, but this fix is good anyway, which should be more consistent with uMS behavior.

PR #47, opened 2020-12-11
Update readme.md

jon301:


dai-shi (approved):
Thanks for the suggestion! Let me merge it for now as it explains the usage. We might change it in the future, if we had better one or needed restructuring.
Thanks.

PR #49, opened 2020-12-13
support suspense error and better typing

dai-shi:
I think it shouldn't replace the state property when a promise is resolved.


dai-shi:
TypeScript version of this: https://twitter.com/0xca0a/status/1329023210768265216
https://codesandbox.io/s/react-typescript-forked-z38kv?file=/src/App.tsx

PR #54, opened 2020-12-22
feat: improve re-render behavior and embed useSubscription technique in uMS emulation

dai-shi:
What I did in #46 was more or less a hack and introduced unnecessary invocation of render function.
This PR is a rewrite of uMS emulation to deal with the fix in #46, while keeping the render behavior the same as before. I needed to embed useSubscription, otherwise it'd be doing the same thing twice.
Note: We use the useSubscription technique which does invoke render function several times before committing.


dai-shi:
Make sure the codesandbox in #46 works. (Hey, can we make a test for this??)
https://codesandbox.io/s/nifty-wescoff-3xkpu
Seems good.

dai-shi:
https://codesandbox.io/s/cool-bird-8k43m?file=/src/App.js
Checking forceUpdate workaround works.

PR #56, opened 2020-12-24
fix: better render invocation with scoped subscription

dai-shi:
While I was working on #54, I noticed a slight unexpected behavior. A hook can call render function several times when another hook bails out.
Turns out we should use scope subscription instead. This PR fixes it and adds the test spec for the expected behavior.


dai-shi:
https://codesandbox.io/s/damp-river-8tkvu
Checked the example in #46 still works.

PR #57, opened 2020-12-29
Use useState instead of useRef in util

istarkov:
Gives nothing, just
More readable (- 2 `if`s)
Less size (minified)
Uses "native" React mechanics to init something one time.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -43,9 +43,9 @@
     "gzipped": 1118
   },
   "utils.js": {
-    "bundled": 3391,
-    "minified": 1141,
-    "gzipped": 645,
+    "bundled": 3356,
+    "minified": 1089,
+    "gzipped": 624,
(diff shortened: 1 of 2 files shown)


dai-shi:
Thanks for the suggestion, but I think useState does more job than useRef, so performance-wise useRef should be better (never confirmed though). I'm not sure about the readability, because I'm more used to lazy initialization with useRef. 🤔

istarkov:
Have no strong opinion, switched to useState in my code from useRef for 1 time init after reading long long thread somewhere what you can and can't do in render after suspense :-) 
_(as I remember one time init is allowed case, btw less things I need to hold in my head the better I feel :-))_

dai-shi:
Yeah, one time lazy init should be fine in CM.
Let's keep this as is, because that was the original purpose of this hook to hide the lazy init of useRef.

PR #59, opened 2020-12-30
fix: do not reuse weak map and avoid lazy initialization

dai-shi:
- do not reuse weak map for deepChangedCache (safer)
- useLayoutEffect instead of useRef lazy init (`p` can be changed, but it would re-render anyway)

I didn't see any actual issues, but they are theoretical issues.


dai-shi:
https://www.npmjs.com/package/valtio/v/0.5.1

PR #60, opened 2020-12-30
Properly check for value for promise result

zvictor:
Valtio currently does not work well with falsy values. This PR is a small improvement to allow falsy values to not be ignored.

Use case:
```js
const state = proxy({ user: user().catch(e => null) })
```

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 9899,
-    "minified": 4642,
-    "gzipped": 1702,
+    "bundled": 9901,
+    "minified": 4644,
+    "gzipped": 1706,
     "treeshaked": {
       "rollup": {
         "code": 137,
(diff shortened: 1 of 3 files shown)


zvictor:
The problem can be seen in action at

https://github.com/zvictor/faugra/blob/dcb41d832b78afadadd63438dec91d63288078d3/examples/with-proxied-authentication/pages/index.js#L16-L30

If `user === null`, the `snapshot` object doesn't get updated.

dai-shi (approved):
Oh, nice catch! I missed that.
Would you like to add a test case in `tests/async.test.tsx` ?

zvictor:
there you go :)

dai-shi (approved):
Very nice. Thanks for your contribution!

dai-shi:
Published: https://www.npmjs.com/package/valtio/v/0.5.2

PR #62, opened 2021-01-04
feat: ref api

### LINKED ISSUE #61: 💊 [feature request]: valtio should support non-serializable data, too
Valtio needs something that allows users to put non proxied data into the store. Currently it cannot handle foreign classes and objects, since accessors to parents/children end up proxying the entire system because Valtio cannot differentiate and just climbs through the nested graph. This is shaping up to be a bigger concern because right now it can only hold serializable data.

A suggestion would be to introduce some kind of hint that makes it only check reference equality if the users wants it so.

```jsx
import { proxy, ref } from 'valtio'

const state = proxy({
  status: true,
  city: 'Berlin',
  body: ref(document.body), // < will not be proxied
  audio: ref(new Audio('/hello.mp3')), // < will not be proxied
})

// I guess it needs to be consistent, so overwrites probably need it, too ...
<div onClick={() => state.audio = ref(new Audo('/chime.mp3'))} />
```

This wouldn't work with SSR or localCache, but at least we have the ability to throw complex foreign state into the store in local scenarios.

dai-shi:
close #61 

## Summary

`ref` will mark an object so that it won't be wrapped by proxies when used as a (nested) property. 

## Usage

```js
import { proxy, ref } from 'valtio'

const state = proxy({
  obj: {},
  body: ref(document.body), // this is not tracked
})
```

## Caveat 1

This `ref` is for tracking mutations only.

For render optimization (= `useProxy(state)` use case), objects could wrapped by proxies if they are plain objects, but that would only happen on demand, so it shouldn't be a problem. It doesn't influence the state object either.

## Caveat 2

```js
const state = proxy(ref(obj)) // this is not supported, it's nonsense
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
-<img src="logo.svg">
+<img src="logo.svg" alt="valtio">
 <br />
-<br/>
+<br />
 
 <code>npm i valtio</code> makes proxy-state simple
 
@@ -107,6 +107,20 @@ You can subscribe a component to state without causing render, just stick the su
 function Foo() {
   const ref = useRef(state.obj)
   useEffect(() => subscribe(state.obj, () => ref.current = state.obj), [state.obj])
+  // ...
+```
+
+##### Avoid state properties to be wrapped with proxies
+
+See: https://github.com/pmndrs/valtio/pull/62 for more information.
+
+```js
+import { proxy, ref } from 'valtio'
+
+const state = proxy({
+  count: 0,
+  dom: ref(document.body),
+})
 ```
 
 ##### Dev tools
--- a/readme.md
+++ b/readme.md
@@ -112,7 +112,7 @@ function Foo() {
 
 ##### Avoid state properties to be wrapped with proxies
 
-See: https://github.com/pmndrs/valtio/pull/62 for more information.
+See https://github.com/pmndrs/valtio/pull/62 for more information.
 
 ```js
 import { proxy, ref } from 'valtio'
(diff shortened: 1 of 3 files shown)


dai-shi:
@drcmda Would you be able to try this build and see if it works as expected?

drcmda:
@dai-shi works! :-D blocks inner mutation, but ref equality still holds.

dai-shi:
@drcmda 
Great to hear that. Would you help me adding this feature in readme.md? Any suggestion?

dai-shi:
Published: https://www.npmjs.com/package/valtio/v/0.6.0

StarpTech:
This feature is very helpful. Without bypassing the `EventEmitter` the code won't work when subscribing to an event.

```js
import { proxy, ref } from 'valtio';

export const state = proxy({
  isConnecting: false,
  signals: ref(new EventEmitter()),
  connection: null
});
```

The following code will results in:

```
Uncaught TypeError: 'set' on proxy: trap returned falsish for property 'Notification'
```

```js
import { proxy, ref } from 'valtio';

export const state = proxy({
  isConnecting: false,
  signals: new EventEmitter(),
  connection: null
});

// anywhere
const { connection, signals } = useSnapshot(state);
useEffect(() => {
 signals.on('Notification', fn);
 return () => {
   return signals.off('Notification', fn);
 };
}, [signals)
```

dai-shi:
@StarpTech You might want to try using `state.signals` instead inside useEffect. If it doesn't work, `ref()` would be unavoidable.

StarpTech:
Yes, it works. I'm wondering why it happened at all. It feels like a bug. See https://github.com/GoogleChrome/proxy-polyfill/issues/20#issuecomment-240027418

StarpTech:
Removing `ref()` will result in an infinite loop.

PR #63, opened 2021-01-05
Add missing } in readme

LookRain:


dai-shi:
I think it's intentional (by @drcmda ) but the styles are not consistent throughout the readme.

dai-shi:
Let's leave it as is for now, because closing `}` doesn't make sense without using `ref` anyway.

PR #65, opened 2021-01-14
Fix/follow ref in snapshot

dai-shi:
Follow-up #62 

It turns out that the "Caveat 1" in #62 can be a real issue, not quite surprisingly...

So, this fixes the behavior and the Caveat 1 is gone.

The issue is reported in discord like this:
> Is `ref` supposed to make it possible to store arbitrary React components inside of valtio state? I'm running into an issue when storing certain components
https://codesandbox.io/s/qo9nh

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 10076,
-    "minified": 4707,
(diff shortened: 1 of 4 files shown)


dai-shi:
Published: https://www.npmjs.com/package/valtio/v/0.6.1

PR #70, opened 2021-01-25
fix: preserve type in ref wrapper

tsumo:


dai-shi (approved):
Nice catch! Looks good.

dai-shi:
Published: https://www.npmjs.com/package/valtio/v/0.6.3

Thanks for your contribution!

PR #71, opened 2021-01-31
feat: proxyWithComputed util

dai-shi:
object getters are somewhat supported but not very handy.
This adds a util function to make it useful with dependency tracking.
This is for expert users: https://github.com/pmndrs/valtio/pull/71#issuecomment-770474003

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -134,31 +134,32 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
 export const proxyWithComputed = <T extends object, U extends object>(
   initialObject: T,
   computedFns: { [K in keyof U]: (snap: NonPromise<T>) => U[K] }
-): T & U => {
-  const getProxy = () => p
+) => {
   const NOTIFIER = Symbol()
   Object.defineProperty(initialObject, NOTIFIER, { value: 0 })
-  const notifyUpdate = () => ++(getProxy() as any)[NOTIFIER]
   ;(Object.keys(computedFns) as (keyof U)[]).forEach((key) => {
     const fn = computedFns[key]
-    let prevComputed: U[typeof key]
+    let computedValue: U[typeof key]
     let prevSnapshot: NonPromise<T> | undefined
     let affected = new WeakMap()
     const wrappedFn = function () {
-      const state = getProxy()
-      const snap = snapshot(state)
-      if (!prevSnapshot || isDeepChanged(prevSnapshot, snap, affected)) {
+      const nextSnapshot = snapshot(p)
+      if (
+        !prevSnapshot ||
+        isDeepChanged(prevSnapshot, nextSnapshot, affected)
+      ) {
         affected = new WeakMap()
-        prevComputed = fn(createDeepProxy(snap, affected))
-        if (prevComputed instanceof Promise) {
-          prevComputed.then((v) => {
-            prevComputed = v
-            notifyUpdate()
+        computedValue = fn(createDeepProxy(nextSnapshot, affected))
+        if (computedValue instanceof Promise) {
+          computedValue.then((v) => {
+            computedValue = v
+            ++(p as any)[NOTIFIER] // HACK notify update
           })
+          // XXX no error handling
         }
-        prevSnapshot = snap
+        prevSnapshot = nextSnapshot
       }
-      return prevComputed
+      return computedValue
     }
     Object.defineProperty(initialObject, key, {
       get: wrappedFn,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -126,9 +126,11 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
  * @example
  * import { proxy } from 'valtio'
  * import { computed } from 'valtio/utils'
- * const state = proxy({ count: 1 })
- * const getDoubled = computed(state, snap => snap.count * 2)
- * const doubled = getDoubled()
+ * const doubledGetter = computed(snap => snap.count * 2)
+ * const state = proxy({
+ *   count: 1,
+ *   get doubled() { return doubledGetter.apply(this) },
+ * })
  */
 export function computed<T extends object, U>(fn: (snap: NonPromise<T>) => U) {
   let prevComputed: U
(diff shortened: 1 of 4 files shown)


thelinuxlich:
I think if computed returns a function instead of a reactive value, kinda defeats the purpose

dai-shi:
I was hoping to assign the function as object getters. It turned out really hard. I might spend some more time, but I think this is still for experts. Usually, it's not needed.

dai-shi:
@thelinuxlich 
I noticed I did something unnecessary.
The new commit is much better, only I need to add a small change in core. `.bind(p)`

I hoped we could do something like this:
```js
const state = proxy({
  count: 0,
  get doubled: computed(snap => snap.count * 2),
})
```
But, it seems impossible.

thelinuxlich:
What if proxy accepted a optional second parameter as a object with the computeds?

```js
const state = proxy({
  count: 0 }, {
  doubled: computed(snap => snap.count * 2)
}
})
```

dai-shi:
I guess that style could allow `Object.defineProperty()`. So, it may work. I would do it in utils, though.
```js
import { proxyWithGetters } from 'valtio/utils'
```

dai-shi:
Here you go.
Now, if it does work, I don't think we need to separate functions as there would be no other use cases.
So, ideally:
```js
import { proxyWithComputed } from 'valtio/utils'
```

thelinuxlich:
Actually, since you expect only computeds in the second object, you should be able to hide the computed wrapper and do it internally, right?

```js
import { proxyWithComputed } from 'valtio/utils'

const state = proxyWithComputed({
    count: 0 
  }, {
    doubled: snap => snap.count * 2
  }
})
```

thelinuxlich:
Also, is it possible to support Promises as the common Valtio properties? (async computed)

AjaxSolutions:
> Actually, since you expect only computeds in the second object, you should be able to hide the computed wrapper and do it internally, right?

If there is the second object, then we could use just _proxy_.  The second object would imply that this is _proxyWithComputed_.

thelinuxlich:
@AjaxSolutions I agree but @dai-shi insists this should be separate

src/utils.ts
```diff
@@ -118,58 +118,53 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
 }
 
 /**
- * computed
+ * proxyWithComputed
  *
- * This is to create a function to return a computed value
- * with dependency tracking.
+ * This is to create a proxy with initial object and additional object,
+ * which sepcifies functions for computed values with dependency tracking.
```
AjaxSolutions, line 124: Typo: sepcifies -> specifies

thelinuxlich:
I saw your changes, excellent! I'm gonna provoke you with one additional feature, computed setters: 

```js
import { proxyWithComputed } from 'valtio/utils'

const state = proxyWithComputed({
    firstName: 'foo',
    lastName: 'bar'
  }, {
    fullName: {
       get(snap) {
           return snap.firstName + ' ' + snap.lastName
       },
      set(newValue, state) {
         const names = newValue.split(' ')
         state.firstName = names[0]
         state.lastName = names[names.length - 1]
      }
     }
  }
})
```

dai-shi:
> you should be able to hide the computed wrapper and do it internally, right?

yes, done. much cleaner with type inference.

> async computed

done. this is hacky. think it as experimental.

> this should be separate

yes, for several reasons.
1. we want to keep api as simple as possible
2. this means we don't want to encourage ordinary users to use it (I may need to add a note)
3. because, it not always good. in most use cases, render optimization works fine
4. it can be worse as computed comes with a cost (and that cost overlaps with render optimization)
5. so, even for expert users, use this carefully. the recommendation is to measure it and choose. use it blindly will cause performance drop.

If someone is keen on performance, my recommendation is not to depend on dependency tracking, but to manual optimize it and pre-compute.

The reasons I'm motivated to create this in utils are, a) of technical interest, b) useful in some edge cases, c) useful in vanilla without react, and d) that i might be wrong and someone can experiment with it.

dai-shi:
oh, you guys are so fast!

src/utils.ts
dai-shi: thanks!

dai-shi:
> computed setters

That's an interesting idea. I'd need to add object setters support firstly in core.

thelinuxlich:
Very cool, now Valtio has pretty much what other reactive frameworks have like Svelte and Vue. This deserves a major release bump.

thelinuxlich (approved):

src/utils.ts
```diff
@@ -114,3 +116,72 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
...
+ *
+ * This is to create a proxy with initial object and additional object,
+ * which specifies getters for computed values with dependency tracking.
+ * It also accepts optional setters for computed values.
+ *
+ * [Notes]
+ * This is for expert users and not recommended for ordinary users.
+ * Contradictory to its name, this is costly and overlaps with useProxy.
+ * Do not try to optimize too early. It can worsen the performance.
+ * Mesarement and comparison will be very important.
```
AjaxSolutions, line 131: Sorry, I'm the typo guy: Mesarement -> Measurement

dai-shi: That's actually helpful. Can you make it as a suggestion?

src/utils.ts
```diff
@@ -121,35 +121,51 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
...
  * This is to create a proxy with initial object and additional object,
- * which specifies functions for computed values with dependency tracking.
+ * which specifies getters for computed values with dependency tracking.
+ * It also accepts optional setters for computed values.
+ *
+ * [Notes]
+ * This is for expert users and not recommended for ordinary users.
+ * Contradictory to its name, this is costly and overlaps with useProxy.
+ * Do not try to optimize too early. It can worsen the performance.
+ * Mesarement and comparison will be very important.
```
AjaxSolutions, line 131:
```suggestion
* Measurement and comparison will be very important.
```

PR #74, opened 2021-02-02
Better computed docs

thelinuxlich:


readme.md
```diff
@@ -173,6 +173,25 @@ import { proxyWithComputed } from 'valtio/utils'
...
+  doubled: snap => snap.count * 2
+})
+```
+
+Computed values accept custom setters too:
+
+```js
+import { proxyWithComputed } from 'valtio/utils'
+
+const state = proxyWithComputed({
```
dai-shi, line 185:
```suggestion

// Computed values accept custom setters too:

const state2 = proxyWithComputed({
```

Let's make it as short as possible.

readme.md
```diff
@@ -173,6 +173,25 @@ import { proxyWithComputed } from 'valtio/utils'
...
+  firstName: 'Alec',
+  lastName: 'Baldwin'
+}, {
+  fullName: {
+    get(snap) { return snap.firstName + ' ' + snap.lastName },
+    set(newValue, state) {
+       const names = newValue.split(' ')
+       state.firstName = names[0]
+       state.lastName = names[names.length - 1]
+    }
```
dai-shi, line 195:
```suggestion
    get: (snap) => snap.firstName + ' ' + snap.lastName,
    set: (state, newValue) => { [state.firstName, state.lastName] = newValue.split(' ') },
```

You might want to deal with middle name, but let's do this.

dai-shi (commented):
Thanks for the suggestion. Two comments from me.

thelinuxlich:
We are missing a async computed example but I don't know if it would belong in this section or in the React Suspense section.

dai-shi:
We can keep it secret for now. (It's pretty experimental.)
If you would like to add it, please add it to the source code comment.
Keeping readme small is crucial for us.

dai-shi (approved):

PR #84, opened 2021-02-12
Add dependency order information for computeds of computeds

thelinuxlich:
And it would be cool if eslint-plugin-valtio can warn about this too!


dai-shi (approved):
Looks good, small, and contained.

(We may want to use `<overview>` if readme gets bigger.)

PR #87, opened 2021-02-15
Add information about storing vanilla functions in state object

thelinuxlich:


readme.md
```diff
@@ -168,6 +168,21 @@ function Foo() {
   const [snapshot, state] = useLocalProxy({ count: 0, text: 'hello' })
 ```
 
+### Adding vanilla functions to the state object
+
+For organization' sake, you might want to add functions in your state object. Just be aware that these functions won't have any reactivity, as vanilla Javascript, they work through explicit invocation.
+
```
AjaxSolutions, line 174: I like this [example](https://twitter.com/0xca0a/status/1329023210768265216) better.

thelinuxlich: I don't see what you mean, in that example the user is assigning to a reactive field the result of a fetch, a Promise.

AjaxSolutions: OK I see your point, but I prefer the _addCity_ example in this [video](https://youtu.be/XnlECbwcJZ0?t=461).  I'd rather document Jack Herrington's approach, but both can be documented if you feel strongly about it.

thelinuxlich: I made a small change, but I feel the example should be kept simple and concise.

dai-shi (commented):
I have two comments on this.
- we'd like to keep readme simple and concise as you said, and this is slightly different from  others. this is more like best practice or one of it, using the library. I agree it should be documented somehow, but it we were to add this here, there would be so many others to add.
- I'm still considering which to recommend a) using `state` in `fetch` and allowing both `snapshot.fetch()` and `state.fetch()`, or b) using `this` in `fetch` and only allowing `state.fetch()`.

Meanwhile, I would be more comfortable to create a wiki page and link there from readme.

thelinuxlich:
Why use this in fetch when state is accessible?

thelinuxlich:
Instead of this we could have a section explaining what types of variables aren't wrapped with proxies(Function, Set), probably within the _ref_ documentation

dai-shi:
> Why use this in fetch when state is accessible?

Because it looks more self organized? You could even create the object before `proxy()`.

dai-shi:
> Instead of this we could have a section explaining what types of variables aren't wrapped with proxies(Function, Set, Map, Date), probably within the _ref_ documentation

That sounds good. I still like to use `<overview>` as it gets long.

How do you like the idea of wiki pages for this (this issue) kind of best practices?

thelinuxlich:
I never used the overview tag, do you mean summary and details? IMO makes some important info hidden, like React-Easy-State docs https://github.com/RisingStack/react-easy-state

dai-shi:
It would depend. It may hide other basic info less visible, otherwise.

yes, details/summary. why did I think overview???

thelinuxlich:
I think the docs need just a ToC. I will close this PR.

PR #88, opened 2021-02-19
chore(tests): add simple tests for ref rerender and subscription

### LINKED ISSUE #66: Add test for `ref`
We added `ref` in #62 and #65, but there are no tests for it.

YPAzevedo:
Adds some simple tests for `ref` that was introduced in #62 and #65  fixes #66

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -367,5 +367,6 @@ it('should not trigger re-render when mutating object wrapped in ref', async () 
   await findByText('count: 0')
 
   fireEvent.click(getByText('button'))
+  await Promise.resolve()
   await findByText('count: 0')
 })
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -1,6 +1,6 @@
 import React, { StrictMode, useRef, useEffect, useState } from 'react'
 import { fireEvent, render, waitFor } from '@testing-library/react'
-import { proxy, useProxy } from '../src/index'
+import { proxy, ref, useProxy } from '../src/index'
 
 it('simple counter', async () => {
   const obj = proxy({ count: 0 })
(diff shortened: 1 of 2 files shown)


tests/basic.test.tsx
```diff
@@ -344,3 +344,28 @@ it('render from outside', async () => {
...
+  const { getByText, findByText } = render(
+    <StrictMode>
+      <Counter />
+    </StrictMode>
+  )
+
+  await findByText('count: 0')
+
+  fireEvent.click(getByText('button'))
+  await findByText('count: 0')
```
dai-shi, line 370: I guess this check is weak. It doesn't guarantee that the count keeps `0`, does it?

dai-shi: can you double check if this fails if you remove `ref`? I might be wrong though.

tests/subscribe.test.tsx
```diff
@@ -83,4 +83,16 @@ describe('subscribe', () => {
...
+  it('should not call subscription for objects wrapped in ref', async () => {
+    const obj = proxy({ nested: ref({ count: 0 }) })
+    const handler = jest.fn()
+
+    subscribe(obj, handler)
+
+    obj.nested.count += 1
+
+    await Promise.resolve()
+    expect(handler).toBeCalledTimes(0)
```
dai-shi, line 96: this is probably ok.

dai-shi (commented):
Thanks for your work!
overall it looks good.
the tests cover #62.
they don't cover #65. (it's tricky, probably needs some discussions.)

tests/basic.test.tsx
YPAzevedo: yeah, it does still pass without ref 🤔 what would be your recommendation here to make the test stronger?

YPAzevedo: added the wait to make the assertion of findByText, wait for the state update to be flushed. now it fails with no ref :)

dai-shi: Yeah, I wasn't so sure about the best way.
Promise.resolve() is a nice hack, even though it depends on some implementation details. We are doing it in other tests, anyway.
Good to go. And hope someone has a better idea in the future.

dai-shi (approved):

tests/basic.test.tsx
YPAzevedo: Yeah, thanks! i'll see if theres a better solution and update this and the other tests! 👍

PR #90, opened 2021-02-22
Feat/nested computed

### LINKED ISSUE #85: Computed fields can't be added to nested objects
This would enable a computed to belong to the same namespace(i.e object) of a reactive property, for better organization

https://codesandbox.io/s/valtio-simple-counter-forked-z9mzx?file=/src/App.js

dai-shi:
This is to implement nested computed proposed in #85.

- Core change: always freeze snapshot except for having object getters. Previously, it doesn't freeze in production. (This is more in line with immer v8.)
- Implement nested computed, fixes #85 
- Detect and error overwriting with computeds, resolves https://github.com/pmndrs/eslint-plugin-valtio/issues/5
- Some refactors

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -191,4 +191,12 @@ const state2 = proxyWithComputed({
     get: (snap) => snap.firstName + ' ' + snap.lastName,
     set: (state, newValue) => { [state.firstName, state.lastName] = newValue.split(' ') },
 })
+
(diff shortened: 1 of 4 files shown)


dai-shi:
On second thought, I don't think nobody can maintain this code. Let's try something more valtio-like simple solution.

PR #92, opened 2021-02-22
chore(test): add another test case for ref

YPAzevedo:
adds test for the missing use case (i hope so) in #65

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -370,3 +370,31 @@ it('should not trigger re-render when mutating object wrapped in ref', async () 
   await Promise.resolve()
   await findByText('count: 0')
 })
+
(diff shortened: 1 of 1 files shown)


YPAzevedo:
@dai-shi i'm assuming this will cover that last test case. Let me know if not

dai-shi (approved):
Looks good. Thanks for your contribution!

We may want to have separate `tests/ref.test.tsx`? what do you think?

YPAzevedo:
> Looks good. Thanks for your contribution!
> 
> 
> 
> We may want to have separate `tests/ref.test.tsx`? what do you think?

Yes i agree, ill open another PR soon to move the ref related tests to a new file. 👌

PR #94, opened 2021-02-23
feat: addComputed

### LINKED ISSUE #85: Computed fields can't be added to nested objects

### LINKED ISSUE #86: A util method to add computed to state
Right now it's natural to add a reactive property to a proxified state object, what's missing is a util method for doing the same with computed:

```js
import { addComputed } from 'valtio/utils'
import state from './store/root' 

addComputed(state, { double(snap) { return snap.counter * 2 } })
```

dai-shi:
#90 trial made me think that valtio's computed model is different from other libraries concepts.
`proxyWithComputed` works with object getters because it is the same proxy object.
We can't support addComputed and nested with the same way.
So, here we add a new `addComputed`. Unfortunately, we can't infer typing nicely.

This should close #85 and close #86.

----

We put `addComputed` at first and then `proxyWithComputed` in readme. addComputed supports async well and we drop (undocumented / experimental) async from proxyWithComputed. However, async in addComputed is still experimental and not documented.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -191,4 +191,12 @@ const state2 = proxyWithComputed({
     get: (snap) => snap.firstName + ' ' + snap.lastName,
     set: (state, newValue) => { [state.firstName, state.lastName] = newValue.split(' ') },
 })
+
+// if you want a computed value to derive from another computed, you must declare the dependency first:
+const state = proxyWithComputed({
+  count: 1,
+}, {
+  doubled: snap => snap.count * 2,
+  quadrupled: snap => snap.doubled * 2
+})
 ```
--- a/readme.md
+++ b/readme.md
@@ -158,10 +158,46 @@ subscribe(state, () => {
 })
 ```
 
-#### Proxy with computed
+#### Computed values
 
-You can have computed values with dependency tracking. This is for experts.
-[Notes](./src/utils.ts#L121-L143)
+You can have computed values with dependency tracking.
+Dependency tracking in valtio conflicts with the work in useSnapshot.
+React users should consider using render functions (optionally useMemo)
+as a primary mean. Computed works well for some edge cases.
+
+##### `addComputed`
+
+This is to add new computed to an existing proxy state.
+It can add computed to different proxy state.
+
+```js
+import { addComputed } from 'valtio/utils'
+
+// create a base proxy
+const state = proxy({
+  count: 1,
+})
+
+// add computed to state
+addComputed(state, {
+  doubled: snap => snap.count * 2,
+})
+
+// create another proxy
+const state2 = proxy({
+  text: 'hello',
+})
+
+// add computed from state to state2
+addComputed(state, {
+  doubled: snap => snap.count * 2,
+}, state2)
+```
+
+##### `proxyWithComputed`
+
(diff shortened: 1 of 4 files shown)


dai-shi:
Should we name it `addComputed`?

thelinuxlich:
addComputed would be better IMO

thelinuxlich:
So this PR adds nested computeds and addComputed?

dai-shi:
More precisely, I concluded that the ideal nested computed is not possible. You can get the similar result using addComputed (with the optional argument, we need to document this.)

thelinuxlich:
I didn't saw a test demonstrating addComputed with get/set and nested, is the API different than proxyWithComputed then?

dai-shi:
yes, different. Let me add some tests later.

And regretfully, I wasn't able to support `set` in addComputed. It could be with a different property like get for `.foo` and  set for `.setFoo()`, but then it doesn't have to be a util. At least, not for this one.

thelinuxlich:
I see the optional third argument now. Why it can't be: 
```javascript
addComputed(
    state.math,
    {
      doubled: (snap) => computeDouble(snap.math.count),
    }
  )
```

thelinuxlich:
btw if in the future we're able to support computed getter/setter with `addComputed` might as well remove `proxyWithComputed` to have only one way of having computeds in the state.

dai-shi:
Oh, that was my desire as I wrote at first in #86.

> doubled: (snap) => computeDouble(snap.math.count),

No way to access this `snap`. It has to be `doubled: (snap) => computeDouble(snap.count),`, which is not "nested".

thelinuxlich:
Why a state property is not nested, does valtio stores it in a flattened path internally?

dai-shi:
Not sure what's you are asking. it's just js objects. you don't even know if there's only one parent.
```js
const obj1 = {}
const obj2 = {}
const child = {}
obj1.child = child
obj2.child = child
```
If you only get `child`, you don't know its parent.

thelinuxlich:
That's what I thought, so I can't understand why you can't add the computed property to the reference as the first parameter

dai-shi:
I guess you are confused with `proxyObject` and `targetObject`. The first argument is to create snapshot.

thelinuxlich:
Still confused lol

dai-shi:
Hm, not sure if I should merge this. Nobody is going to use it. I wish we could add this and drop `proxyWithComputed` but that's also a nice function with a cute hack.

Leaving it for a while..

thelinuxlich:
Yeah, I feel like if you were able to drop the third parameter and have `addComputed(state, obj_with_computeds)` would be a cool replacement to `proxyWithComputed`

dai-shi:
Didn't expect you say so. The third argument is for nesting and the big reason of developing `addComputed` over `proxyWithComputed`. addComputed doesn't support setter, which I'm personally fine, but it's one of the cute features in proxyWithComputed.

thelinuxlich:
well, at least it's a optional argument :)

PR #95, opened 2021-02-23
chore(test): move ref tests to new file

YPAzevedo:
Moving the tests for ref into a new file to keep code organized.


dai-shi (approved):
Thanks!

PR #100, opened 2021-03-07
feat: valtio/macro

dai-shi:
note: test requires build in advance. FIXME if possible.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -57,21 +57,21 @@
     }
   },
   "macro.js": {
-    "bundled": 1516,
-    "minified": 933,
-    "gzipped": 505
+    "bundled": 1520,
+    "minified": 937,
+    "gzipped": 508
   },
   "macro.module.js": {
-    "bundled": 1362,
-    "minified": 847,
-    "gzipped": 464,
+    "bundled": 1366,
+    "minified": 851,
+    "gzipped": 467,
     "treeshaked": {
       "rollup": {
-        "code": 740,
+        "code": 744,
         "import_statements": 251
       },
       "webpack": {
-        "code": 1801
+        "code": 1805
       }
     }
   }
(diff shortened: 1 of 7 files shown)


dai-shi:
> Error
Could not find path: '@babel/helper-module-imports'.

Hmm.

dai-shi:
https://codesandbox.io/s/react-typescript-forked-4lmr2?file=/src/App.tsx
Minimal example.

dai-shi:
Updated minimal example: https://codesandbox.io/s/react-typescript-forked-dfzh0?file=/src/App.tsx

dai-shi:
If we push this feature, we could make breaking changes to rename hooks??
useProxy -> useSnapshot
useBoundProxy -> useProxy

ref: https://twitter.com/dai_shi/status/1366308344999288837

drcmda:
sounds good, new major?

dai-shi:
@drcmda Would you try it in some of your projects and see how it practically works?

> new major?

We can do minor and add deprecation notice, then major.

@Aslemammad would be surprised with renaming for his eslint-plugin-valtio.

dai-shi:
> ofc i have no idea how any of this works, but it seems that would be optimal

I don't think it's easy to implement, or even possible.

proxy state is often defined in other files and imported, and also we can use a part of state.
```ts
import { state } from './somewhere' // is this proxy state or not?

  return <div>{state.foo.bar}</div> // should we subscribe to state or state.foo?
```

> another thing, if that macro is safe, doesn't cost much to ship - shouldn't it be the only default then? or are there downsides?

it only works with babel. requires some setup. 100% unsure if it's safe. 😓

dai-shi:
> can't it also just be useSnapshot?

I think it's too confusing to have the same name for different purposes / functionalities.

Let's go with:
```js
import { useSnapshot } from 'valtio'
import { useProxy } from 'valtio/macro'
```

dai-shi:
My misclick...

dai-shi:
I will make a follow-up PR.

PR #101, opened 2021-03-07
[BREAKING] rename to useSnapshot

dai-shi:
following up: https://github.com/pmndrs/valtio/pull/100#issuecomment-792231836

- old `useProxy` -> new `useSnapshot`
- old `useBoundProxy` -> new `useProxy` (only in `valtio/macro`)

This deprecates useProxy and useLocalProxy.

Well, it's not really "breaking", as old useProxy is still exported as deprecated.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 11521,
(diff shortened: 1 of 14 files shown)


dai-shi:
Let me release this to get more feedback. will see...

PR #103, opened 2021-03-07
fix typo

Fausto95:


readme.md
```diff
@@ -40,7 +40,7 @@ function Counter() {
   return (
     <div>
       {snap.count}
-      <button onClick={() => ++state.count}>+1</button>
+      <button onClick={() => ++snap.count}>+1</button>
```
dai-shi, line 43: ooookay, this is not typo. this is not in render. snapshot is immutable and mutation will throw an error.

PR #106, opened 2021-03-09
improve vanilla bundle

### LINKED ISSUE #105: subscribeKey doesn't work on vanilla version
When importing the vanilla version of valtio, subscribeKey doesn't work.  It always gives `Error('Please use proxy object')`.

```ts
import { proxy, subscribe, snapshot } from 'valtio/vanilla';
import { subscribeKey } from 'valtio/utils';

let state = proxy({
    name: 'Suzy',
});

subscribeKey(state, 'name', () => {});  // error
```

---

I was able to copy-paste the subscribeKey function into my file, instead of importing it, and then it worked.

I also added the `notifyInSync` variable which is missing from `subscribeKey`:

```ts
// this pasted into the above file, makes it work
export const subscribeKey = <T extends object>(
    proxyObject: T,
    key: keyof T,
    callback: (value: T[typeof key]) => void,
    notifyInSync?: boolean,
) => {
    let prevValue = proxyObject[key];
    return subscribe(proxyObject, () => {
        const nextValue = proxyObject[key];
        if (!Object.is(prevValue, nextValue)) {
            callback((prevValue = nextValue));
        }
    }, notifyInSync);
};
```

dai-shi:
close #105


dai-shi:
@cinnamon-bun Would you try the codesandbox build? ☝️

cinnamon-bun:
Confirmed, `subscribeKey` works from the vanilla build now.

I think I'm "using it wrong" though, because my subscriptions are not always triggering.  Mentioning it here in case it's related to the vanilla vs react bundling issue somehow.

I think if you replace part of a proxied object with a plain old normal javascript object, that new object does not become deeply proxied and so changes inside it won't trigger subscriptions

See comments in:
https://codesandbox.io/s/valtio-issue-105-pr-106-ujiih?file=/src/App.tsx

dai-shi:
https://codesandbox.io/s/react-typescript-forked-g8q4j?file=/src/App.tsx
Working well with counter example.

dai-shi:
@cinnamon-bun

> https://codesandbox.io/s/valtio-issue-105-pr-106-ujiih?file=/src/App.tsx

It's working as expected.

```js
subscribe(state.person1, () => {
    // this never runs at all
    console.log('> subscribe is running.  person1 =', state.person1);
})
```

this above is equivalent to the below.

```js
const person1ProxyObject = state.person1
subscribe(person1ProxyObject, () => {
    // this never runs at all
    console.log('> subscribe is running.  person1 =', state.person1);
})
```

So, if you assign a fresh object `state.person1 = {}`, the subscription is working on the original object, and `person1ProxyObject.name = 'foo'` will work.

Issue #34, opened 2020-11-27
[Umbrella] Improve tests


dai-shi:
I'd admit the current `tests` might not cover all use cases.
To make the library more stable for the future, adding tests would be great.
That would also clarify what is supported, and what is not supported but works for now.

Aslemammad:
Can you highlight the files that need to get improved?

dai-shi:
@Aslemammad I would say, basic.test.tsx, async.test.tsx, class.test.tsx and mapset.test.tsx. 😓 
You can create a new file or files, with totally new test design.

Aslemammad:
@dai-shi I'll take a look, thanks for the issue.

dai-shi:
@Aslemammad you are not working on this issue for now, are you? just confirming.

Aslemammad:
I improved some tests and sent some PRs before, and they got merged too. But for now, no, I'm not working on them.

Aslemammad:
No, actually, I'm not working.

dai-shi:
There's still room for improvement, but this umbrella issue is not very easy to contribute.
We may need to create smaller issues so that new comers can work on them more easily.

PR #107, opened 2021-03-09
Add notifyInSync argument to subscribeKey

cinnamon-bun:
Vanilla `subscribe` takes an optional argument, `notifyInSync`, which if `true` will disable batching of subscriptions.

This PR adds the same argument to `subscribeKey`, which just passes it along to `subscribe`.

(Note, I think there are no tests for this function for any of the vanilla functions)

As requested in https://github.com/pmndrs/valtio/issues/105#issuecomment-793200092

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -31,6 +31,7 @@ export const useLocalProxy = <T extends object>(init: T | (() => T)) => {
  *
  * The subscribeKey utility enables subscription to a primitive subproperty of a given state proxy.
  * Subscriptions created with subscribeKey will only fire when the specified property changes.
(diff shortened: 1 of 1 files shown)


dai-shi:
@cinnamon-bun Thanks for your contribution! I think it's safe to merge this as the change is simple.

> (Note, I think there are no tests for this function for any of the vanilla functions)

Right, there are some tests missing not only for vanilla: #34.
If you are interested, adding vanilla tests would be so helpful!

@tanhauhau also noticed vanilla tests are missing in his [video](https://www.youtube.com/watch?v=uZnO2G8pqn0).

Issue #108, opened 2021-03-10
No way to access resolved data of the async value from state


domlazic:
looking at how async values are implemented it looks like it's not possible to access the resolved data of the async value directly from the state, which could be useful if for example you want to update the value based on the current value.

for non-async values you can do this:

```javascript
const state = proxy({ count: 0 })
state.count = state.count + 1
```

but if `count` is an async value there doesn't seem to be a way of doing the equivalent

I was expecting this to be possible by doing something like this:

```javascript
const state = proxy({ count: Promise.resolve(0) })
state.count = state.count.then((currentCount) => currentCount + 1)
```

but it turns out that the value with which the original promise resolves is not returned from the internal promise handler, so the `currentCount` from above ends up being `undefined`.

is there a specific reason why the value is not returned from the internal promise handler?

**Update:**
I opened a PR with a fix, basically just returning the resolved value from the internal promise handler. I don't know how useful of a change this is so please close the PR if this change doesn't make sense

dai-shi:
Ah, nice catch. It's a use case I overlooked.
Oh, and you filed a PR, so nice.

PR #109, opened 2021-03-10
Async value fulfilled with data

domlazic:
see #108

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -5,9 +5,9 @@
     "gzipped": 1075
   },
   "vanilla.js": {
-    "bundled": 6394,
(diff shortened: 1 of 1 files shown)


src/vanilla.ts
```diff
@@ -140,6 +140,7 @@ export const proxy = <T extends object>(initialObject: T = {} as T): T => {
           .then((v) => {
             target[prop][PROMISE_RESULT] = v
             notifyUpdate()
+            return v
```
dai-shi, line 143: 👍 I was thinking of the same fix.

dai-shi (approved):
Very nice. Thanks for the test too.

Issue #110, opened 2021-03-10
Possible regression on 0.8.1


tommywalkie:
Tried using the newly released `0.8.1` version (via ES Modules) today and couldn't import any of Valtio methods.

[JSFiddle](https://jsfiddle.net/tommywalkie/vbn4kx3u/)

```js
// 0.8.0
import { subscribe as subscribeA } from 'https://cdn.skypack.dev/valtio@0.8.0'
// 0.8.1
import { subscribe as subscribeB } from 'https://cdn.skypack.dev/valtio@0.8.1'

console.log(subscribeA) // ✔️ works
console.log(subscribeB) // ❌ SyntaxError: import not found: subscribe
```

dai-shi:
Thanks for reporting! Sorry for the inconvenience.
Okay, I need to look for some other solutions that work with esm too. 🤔

dai-shi:
```js
import { subscribe as subscribeB } from 'https://cdn.skypack.dev/valtio@0.8.1/vanilla'
```
This seems to work. just FYI.

---

https://github.com/pmndrs/valtio/blob/a396bfdfde33a1ce93c1e9c24e387b068df3d50e/src/index.ts#L1

This doesn't seem to work with pika.

tommywalkie:
@dai-shi That's possible. Looking at https://github.com/snowpackjs/snowpack/issues/2774, the `zarr` package seems to run with similar issues.

dai-shi:
It turns out #112 didn't fix this issue at all.
https://jsfiddle.net/gpLesnzu/

Does anyone know how to work around this?

lauritzsh:
I wanted to try this library since it seems simple and flexible, while allowing writing to state outside React compared to Jotai. However, using Snowpack for development, I have a problem importing just following the examples posted. With

```react
import { proxy } from 'valtio';

export const test = proxy({ hello: 'world' });
```

I get `Uncaught SyntaxError: The requested module '../../../../_snowpack/pkg/valtio.js' does not provide an export named 'proxy'`. Some workarounds seems to be `import { proxy } from 'valtio/vanilla'` but that results in `Uncaught Error: Please use proxy object`. The only thing working for me is 

```react
import v from 'valtio';

export const test = v.proxy({ hello: 'world' });
```

If I do `import * as v from 'valtio';` then the structure is

![image](https://user-images.githubusercontent.com/1979410/114417788-2f996c80-9bb2-11eb-8ae4-7adda7df1a12.png)

[...]

PR #111, opened 2021-03-10
Consider using relative imports

tommywalkie:
Related to https://github.com/pmndrs/valtio/issues/110. Simply replaced `"valtio"` and `"valtio/vanilla"` imports with relative paths so it should make Valtio work with ES Modules CDNs (Snowpack/Skypack, JSDelivr/esm.run, esm.sh etc.) while they add support for/decide to not support named package submodules.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 5830,
-    "minified": 2636,
-    "gzipped": 1075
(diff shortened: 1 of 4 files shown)


dai-shi:
I understand what you mean. This is basically what it was before v0.8.1.
This reverts #106 and we need to fix #105 in some other ways.

tommywalkie:
Oh i see, didn't take `susbcribeKey` into account.

PR #120, opened 2021-03-21
drop deprecated functions for v1

dai-shi:
final work for v1.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,31 +1,6 @@
-import { useRef } from 'react'
-import { proxy, subscribe, snapshot } from './vanilla'
-import { useProxy } from 'valtio'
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -130,8 +105,7 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
  * This adds computed values to an existing proxy object.
  *
  * [Notes]
- * This is for expert users and not recommended for ordinary users.
- * Contradictory to its name, this is costly and overlaps with useSnapshot.
+ * This comes with cost and overlaps with useSnapshot.
```
AjaxSolutions, line 108: Minor: This comes with a cost....

dai-shi: Thanks as always!

PR #125, opened 2021-03-22
fix: addComputed with array.pop() use case

dai-shi:
This PR fixes a weird behavior reported in #124.
Because addComputed runs in sync, it can run get function with incomplete snapshot.
We give up running it sync, and delay one tick. This is not ideal, but until we found a better solution, this should be a safer solution.

Please note again that `addComputed` and `proxyWithComputed` are often not recommended with useSnapshot in react. Consider deriving state in render function as a primary mean.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -167,7 +167,7 @@ export const addComputed = <T extends object, U extends object>(
         targetObject[key] = value
       }
     }
-    subscribe(proxyObject, callback, true)
+    subscribe(proxyObject, callback)
     callback()
   })
 }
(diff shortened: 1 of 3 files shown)


dai-shi:
@opticode Would you make sure if this PR fixes your use case? You can locally install the build created by codesandbox ci.

opticode:
@dai-shi Checked and can confirm it works as expected now! Much appreciated 👏

dai-shi:
Merging this. I will wait for some other bug reports for a while and then make a new patch release.

PR #129, opened 2021-03-26
build: use esbuild plugin

aulneau:
Hi there!

I made a [PR for jotai](https://github.com/pmndrs/jotai/pull/378) a few days ago updating the build tooling to use esbuild as the typescript transpiler and @dai-shi requested I make PRs for [zustand](https://github.com/pmndrs/zustand/pull/343) and valtio, so here we are! The main reason for this change is because of how much faster esbuild is at compiling javascript from typescript:

Before
✨  Done in 24.54s.

After
✨  Done in 9.38s.

There are also some very minor bundle size changes with this update.

### General overview of changes
- added `esbuild` and `rollup-plugin-esbuild`
- replaced the `typescript` plugin in `createESMConfig` and `createCommonJSConfig` to use esbuild
- added `id.includes('@babel/runtime')` to the externals check [per this recommendation](https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers)
- updated the node version used in the codesandbox ci to 12 (due to esbuild requirement)
- added `tslib` as a dev dependency due to it being missing for the typescript plugin


dai-shi:
Hi, thanks for your work!
Would you also apply the same rollup config and concurrently technique in jotai here?

Two reasons:
- that would speed up even more
- we want to have the same structure for all three libs for better maintainability (which wasn't the case before.)

aulneau:
> Two reasons:
>
> - that would speed up even more
> - we want to have the same structure for all three libs for better maintainability (which wasn't the case before.)

Done! I actually think the speed element is the same, but I agree, it's better to have consistent structure between each.

thelinuxlich:
I hope that esbuild can replace babel macro in the future :)

rollup.config.js
```diff
@@ -47,21 +59,25 @@ function createCommonJSConfig(input, output) {
     external,
     plugins: [
       resolve({ extensions }),
-      typescript(),
+      getEsbuild('es2015'),
```
dai-shi, line 62:
```suggestion
```

according to https://github.com/pmndrs/zustand/pull/343#issuecomment-808745679 @aulneau

aulneau: done!

babel.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = (api, targets) => {
       ],
     ],
     plugins: [
+      '@babel/plugin-transform-runtime',
```
dai-shi, line 19: we don't need this?

aulneau: I might be wrong, but I think we do -> https://github.com/pmndrs/jotai/pull/378#discussion_r602957217

I set this as I wanted to get rid of the warning when bundling.

dai-shi (approved):
LGTM

Issue #138, opened 2021-04-04
ESM files deployed to NPM package contain import references to CommonJS sibbling JS files (missing `.module` suffix))


danielweck:
https://unpkg.com/browse/valtio@1.0.1/index.module.js

```js
import { snapshot, subscribe, getVersion } from './vanilla';
export * from './vanilla';
```

...should point to https://unpkg.com/browse/valtio@1.0.1/vanilla.module.js
...but instead point to https://unpkg.com/browse/valtio@1.0.1/vanilla.js

Same issue with https://unpkg.com/browse/valtio@1.0.1/utils.module.js
```js
import { subscribe, snapshot, proxy } from './vanilla';
```

danielweck:
FYI, to work around this problem, I use `@rollup/plugin-replace` in my build process with the following configuration:

```js
{
  include: /.+valtio.+/,
  delimiters: ['', ''],
  "'./vanilla'": "'./vanilla.module'"
}
```

danielweck:
Ah, actually, when I use Preact WMR's development server, `vanilla.module.js` imported by `index.module.js` (via `import ... from 'valtio'`) is not equal to the one imported by `utils.module.js` (via `import ... from 'valtio/utils'`), causing the proxy to be corrupted!
So my final fix is in fact to replace `import ... from ''./vanilla'` to `import ... from ''valtio/vanilla'` in both `index.module.js` and `utils.module.js`. Thanks to the `package.json` `exports`, this works fine.

dai-shi:
Thanks for reporting.
This is causing by #112. So, unpkg follows "module" in package.json but not "exports"...

I somehow thought `*.module.js` is better (maybe because of types, but we are now copying them), but probably `esm/*.js` would solve it.

danielweck:
Thank you for your prompt reply. I look forward to testing the fix, once merged and published.

I am hitting an interesting Preact WMR "bug" (maybe a "feature", I'm not sure), where the export map is correctly used by the Rollup-based production build system, but at development time some module caching seems to be involved which results in breaking the proxy functionality:

This happens when I import `devtools()` from `valtio/util`, which in turn invokes `subscribe()` from `./vanilla`:

https://github.com/pmndrs/valtio/blob/8e625551783f2e6364670faf59411fe2b926793d/src/utils.ts#L64
=>
https://github.com/pmndrs/valtio/blob/8e625551783f2e6364670faf59411fe2b926793d/src/vanilla.ts#L188-L192

[...]

PR #140, opened 2021-04-05
fix: no relative import in bundle

dai-shi:
This basically reverts #112, which was a wrong fix for #110.

This should resolve the issue mentioned in https://github.com/pmndrs/valtio/issues/138#issuecomment-813502220 by @danielweck.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -55,10 +55,7 @@
       },
       "alias": {
         "extensions": [".js", ".jsx", ".ts", ".tsx", ".json"],
-        "map": [
-          ["valtio/vanilla", "./src/vanilla.ts"],
-          ["valtio", "./src/index.ts"]
-        ]
+        "map": [["valtio", "./src/index.ts"]]
       }
     }
   },
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,14 +12,7 @@
     "plugin:import/errors",
     "plugin:import/warnings"
   ],
-  "plugins": [
-    "@typescript-eslint",
-    "react",
-    "prettier",
-    "react-hooks",
-    "import",
-    "jest"
-  ],
+  "plugins": ["@typescript-eslint", "react", "prettier", "react-hooks", "import", "jest"],
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
     "ecmaVersion": 2018,
@@ -38,22 +31,13 @@
       "import/namespace": "off",
       "import/default": "off",
       "@typescript-eslint/explicit-module-boundary-types": "off",
-      "no-unused-vars": [
-        "warn",
-        { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }
-      ],
-      "@typescript-eslint/no-unused-vars": [
-        "warn",
-        { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }
-      ],
+      "no-unused-vars": ["warn", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }],
+      "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }],
       "@typescript-eslint/no-use-before-define": "off",
       "@typescript-eslint/no-empty-function": "off",
       "@typescript-eslint/no-empty-interface": "off",
       "@typescript-eslint/no-explicit-any": "off",
-      "jest/consistent-test-it": [
-        "error",
-        { "fn": "it", "withinDescribe": "it" }
-      ]
+      "jest/consistent-test-it": ["error", { "fn": "it", "withinDescribe": "it" }]
     }
   },
   "settings": {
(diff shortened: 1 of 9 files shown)


danielweck:
For the sake of completion, I would reflect the entire export map (i.e. 3 items) in the path aliases, just in case they will be used in unit tests or demo code, in this codebase:

https://github.com/pmndrs/valtio/blob/cbf5ce31485cdb75638f4fb3c4fff6fbe8176310/package.json#L26-L40

=>

`.eslintrc.json `

https://github.com/pmndrs/valtio/blob/ba1ee4ecd18fc6a0ce516b0b0b81372e456e0b50/.eslintrc.json#L56-L62

(missing `["valtio/utils", "./src/utils.ts"]` and `["valtio/macro", "./src/macro.ts"]`

=>

`package.json` (Jest config)

https://github.com/pmndrs/valtio/blob/ba1ee4ecd18fc6a0ce516b0b0b81372e456e0b50/package.json#L103-L106

(missing `"^valtio/macro$": "<rootDir>/src/macro.ts"` and `"^valtio/utils$": "<rootDir>/src/utils.ts"`)

=>

`tsconfig.json`

https://github.com/pmndrs/valtio/blob/ba1ee4ecd18fc6a0ce516b0b0b81372e456e0b50/tsconfig.json#L10-L18

(missing `"valtio/macro": [ "./src/macro.ts" ]` and `"valtio/utils": [ "./src/utils.ts" ]`)

dai-shi:
> just in case they will be used in unit tests or demo code

I wouldn't like to have such cases without explicit notice, so I'm actually happy that it raises an error.
But, if you think it's more consistent and easy to view the project, I can do that as it's harmless.

danielweck:
yes, personally I have a preference for API / design consistency. but you're the project owner so of course do as you please :)

objectively: this change would not add unnecessary code per se (in terms of the generated JS bundle), as it would only add safe project configuration (no risk of regressions).

dai-shi:
I took a different approach. The bundled JS is exactly the same and d.ts files look better in relative path.

danielweck:
i will test this as soon as published :)
(don’t have time to self-build right now)
+1

dai-shi:
@danielweck 
If you are using npm/yarn, you can install the build by codesandbox ci.

image

Click the last "Details" link and look for "Local Install Instructions".

danielweck:
Tested, works great :)
PR Merge?

`https://ci.codesandbox.io/status/pmndrs/valtio/pr/140/builds/116797`
=>
`npm i https://pkg.csb.dev/pmndrs/valtio/commit/6adb7887/valtio`
=>
```diff
                "valtio": {
-                       "version": "1.0.2",
-                       "resolved": "https://registry.npmjs.org/valtio/-/valtio-1.0.2.tgz",
-                       "integrity": "sha512-AR6cR3Xeehpe1B7/LGiC05HYhWWi2YKMGjPVg3oEddTC6joB4PM803SFrPLTzOkjV0ViyHmyEvVK5US9tdx4aQ==",
+                       "version": "https://pkg.csb.dev/pmndrs/valtio/commit/6adb7887/valtio",
+                       "integrity": "sha512-fJlhsxIkkU3rRpzedUC+gJXHNdeg3Ej6/n7t8z9MGzCHJx0llIYur4nUriVGrPcv7sTtcP0y8c5Ej6QiKYS9Yg==",
                        "requires": {
                                "proxy-compare": "1.1.6"
                        }
```

dai-shi:
Great! Thanks for confirming.

danielweck:
By the way: one thing I really love about Valtio and Zustand is that I can simply alias `react` to `preact/hooks` (which my project already depends on), and everything works beautifully 👍  

By contrast, several other micro-libs depend on React features such that a pure Preact project is forced to alias with `preact/compat`, which adds nearly 9KB (not gzipped) of additional code. Sure, tree shaking and dead code removal for production builds help, but if I can avoid it :)

With Jotai, it's just a case of separating React's `createContext` and `createElement` imports from the hooks imports, but that's an extra build / preprocessing step I am happy not to need in Valtio ;)

Keep up the good work!

dai-shi:
Thanks for the feedback. So, with only jotai, it will require `preact/compat`, but it will work with it, correct?

danielweck:
I did not run serious tests with Jotai, but purely from a build perspective, the React/Preact compatibility story looks positive. No need for `preact/compat` as far as I can make out, just a case of importing `createContext` and `createElement` (aka `h`) from `preact` intead of `react`, and of importing all the React hooks from `preact/hooks` instead of `react`. This can be done in a single pass by writing a simple `preact` + `preact/hooks` wrapper, that simply re-exports all exports from both packages. Then, alias Valtio's`react` imports to this "special wrapper", so that a line like this can compile:

https://github.com/pmndrs/jotai/blob/48d71cf23a2cb71e5d93f3de7461dd28bf2d01a8/src/core/Provider.ts#L1
=>
```js
import React, { createElement, useCallback, useRef, useDebugValue } from 'react'
```

Note that the default `React` import in the above line does not matter, as it is only used for the TypeScript type `React.FC` which is erased in deployed JS production builds. See: https://unpkg.com/browse/jotai@0.16.0/index.module.js

PR #152, opened 2021-05-05
Update readme.md

zbeyens:
> The component will only re-render when the parts of the state you access have changed, it is render-optimized

This is clear after re-reading but it was initially counter-intuitive when seeing this pattern for the first time. Coming from zustand, `useSnapshot(state)` is very close to `useStore(state => state)` so it was confusing.

I would add a sentence confirming that it excludes the parts of the state we don't access from the component.


dai-shi (approved):
This looks good with a minimal addition for clarification. Thanks!

PR #157, opened 2021-05-15
Add watch to valtio/utils

### LINKED ISSUE #149: Feature request: Watch
`watch` (pending name) is a proposed function exportable from `valtio` that allows automatically tracking valtio instances instead of explicitly declaring the `valtio` instances you need to watch (subscribe) for. This is similar to `useEffect` except that you don't have to explicitly define the dependency list.

This proposed function allows you to easily compose subscriptions to `valtio` instances and not to individually watch them (`subscribe`).

#### Example

```js
const count = proxy({ value: 0 });

const stop = watch(() => {
  // We mark `count` as our dependency. This
  // callback will then only run whenever count updates.
  console.log(`Count: ${count.value}`);
  
  return () => {
    // We may perform cleanups here when `stop` is called.
  };
});

count.value++; // Count: 1
```

Multiple subscriptions:

```js
const recipient = proxy({ value: 'Dai Shi' });
const greeting = proxy({ value: 'Hello' });

watch(() => {
  console.log(`${greeting.value}, ${recipient.value}`);
});

recipient.value = 'John Doe'; // Hello, John Doe
greeting.value = 'Bonjour'; // Bonjour, John Doe
```

LXSMNSYC:
Resolves #149

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -267,8 +267,8 @@ export const proxyWithComputed = <T extends object, U extends object>(
   return proxyObject
 }
 
-export type WatchGet = <T extends object>(value: T) => T
-export type WatchCallback = (get: WatchGet) => (() => void) | void | undefined
+type WatchGet = <T extends object>(value: T) => T
+type WatchCallback = (get: WatchGet) => (() => void) | void | undefined
 
 let currentCleanups: Set<() => void>
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -270,7 +270,7 @@ export const proxyWithComputed = <T extends object, U extends object>(
 type WatchGet = <T extends object>(value: T) => T
 type WatchCallback = (get: WatchGet) => (() => void) | void | undefined
 
-let currentCleanups: Set<() => void>
+let currentCleanups: Set<() => void> | undefined
 
 /**
  * watch
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -266,3 +266,99 @@ export const proxyWithComputed = <T extends object, U extends object>(
   const proxyObject = proxy(initialObject) as T & U
   return proxyObject
 }
+
+export type WatchGet = <T extends object>(value: T) => T
+export type WatchCallback = (get: WatchGet) => (() => void) | void | undefined
+
+let currentCleanups: Set<() => void>
+
+/**
+ * watch
+ *
+ * Creates a reactive effect that automatically tracks proxy objects and
+ * reevaluates everytime one of the tracked proxy objects updates. It returns
+ * a cleanup function to stop the reactive effect from reevaluating.
+ *
+ * Callback is invoked immediately to detect the tracked objects.
+ *
+ * Callback passed to `watch` receives a `get` function that "tracks" the
+ * passed proxy object.
+ *
+ * Watch callbacks may return an optional cleanup function, which is evaluated
+ * whenever the callback reevaluates or when the cleanup function returned by
+ * `watch` is evaluated.
+ *
+ * `watch` calls inside `watch` are also automatically tracked and cleaned up
+ * whenever the parent `watch` reevaluates.
+ *
+ * @param callback
+ * @returns A cleanup function that stops the callback from reevaluating and
+ * also performs cleanups registered into `watch`.
+ */
+export const watch = (callback: WatchCallback): (() => void) => {
+  const cleanups = new Set<() => void>()
+  const subscriptions = new Set<object>()
+
+  let alive = true
+
+  const cleanup = () => {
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -266,3 +266,99 @@ export const proxyWithComputed = <T extends object, U extends object>(
   const proxyObject = proxy(initialObject) as T & U
   return proxyObject
 }
+
+export type WatchGet = <T extends object>(value: T) => T
+export type WatchCallback = (get: WatchGet) => (() => void) | void | undefined
```
dai-shi, line 271:
```suggestion
type WatchGet = <T extends object>(value: T) => T
type WatchCallback = (get: WatchGet) => (() => void) | void | undefined
```
I prefer hiding types if that's not too much trouble, which allows us to refactor things without hustle.

src/utils.ts
```diff
@@ -266,3 +266,99 @@ export const proxyWithComputed = <T extends object, U extends object>(
...
+
+let currentCleanups: Set<() => void>
```
dai-shi, line 273:
```suggestion
let currentCleanups: Set<() => void> | undefined
```
This is probably proper typing. Or, we could explicitly initialize it with `null`.

dai-shi (commented):
Great work. Overall it looks good.
I hadn't thought about nested watches. Would you like to add tests that depend currentCleanups?

src/utils.ts
LXSMNSYC: Sure thing, will fix.

src/utils.ts
LXSMNSYC: The motivation for the lack of "undefined" is to raise TS2454 during compilation. However, it seems that it isn't useful at this point. Will do the change.

LXSMNSYC:
> I hadn't thought about nested watches. Would you like to add tests that depend currentCleanups?

I have added a test for this, however it isn't that verbose 😅

tests/watch.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    expect(callback).toBeCalledTimes(1)
+  })
+  it('should cleanup recursively when stopped', () => {
+    const callback = jest.fn()
+
+    const stop = watch(() =>
+      watch(() => () => {
+        callback()
+      })
+    )
```
dai-shi, line 74:
```suggestion
    const stop = watch(() =>
      watch(() => () => {
        callback()
      })
      return () => {
        outerCallback()
      }
    )
```

Would it be nice to add this to make sure it doesn't override, but merge cleanups?

What would be the real use case? I think you have something in your mind. It would be nice to have that in tests.

LXSMNSYC: I could probably replace it with a `subscribe` example instead of just `watch` just to demonstrate a cleanup.

dai-shi:
Thanks for the improvement. I'm still not sure the recursive use case, but it's fine for now.

dai-shi (approved):

LXSMNSYC:
I could probably remove it as the nested watch and the cleanup use-cases are already covered, no?

dai-shi:
That looks okay.

My (non-blocking) comment is about the use case of nested watch.

```ts
const stop = watch((get) => {
  // do something
  watch((get) => {
    // do something else
    return cleanup1
  })
  return cleanup2
})
```
When do you need something like this?

LXSMNSYC:
There may be a use-case wherein inside a watch there's a localized proxy object, in which you can `watch` it from the inside. Also, there's nothing that going to stop users from doing nested `watch` calls, so adding an auto-cleanup maybe convenient.

dai-shi:
Thanks for the clarification.

> Also, there's nothing that's going to stop users from doing nested watch calls, so adding an auto-cleanup may be convenient.

Again, I'm fine with your proposal. But, if I were to design it, I would require users to do manual cleanup instead of the auto-cleanup.

```js
const stop = watch(() => {
  const signal = new AbortSignal();
  
  const innerStop = watch(() => () => {
    signal.abort();
  });
  
  // Do other stuffs

  return () => {
    // Do other cleanups

    innerStop();
  };
});
```

I think it's just a design choice.

I prefer manual one, because you never know if you are in watch or not.

```ts
const fn = () => {
  watch(() => {
    return cleanup
  }
}

// expecting this usage
const stop = watch(() => {
  fn()
})

// but someone else could use it directly
fn()
```

dai-shi:
I was trying to resolve conflicts, but github warns me about the branch. I guess it should be fine, but I'd leave it (= resolving conflicts) to you.

dai-shi:
Let me merge this and cut a release.

PR #159, opened 2021-05-18
Remove warnings 

Aslemammad:
I tried replacing the fake `Promise`s with `wait*`, but I read that there's no need for that.


dai-shi (approved):
Wow, thanks so much. I didn't realize it. So, this means it's tricky to test async behaviors.

PR #169, opened 2021-07-04
fix: allow exports type import

### LINKED ISSUE #168: Unsupported export map resolution
Currently, the package' export map looks like this:

```json
{
  "types": "./utils.d.ts",
  "module": "./esm/utils.js",
  "default": "./utils.js"
}
```

However, most bundlers that uses the export map (e.g. Vite, ESBuild, Skypack) uses the `import`/`require` resolution (see [here](https://www.npmjs.com/package/resolve.exports)) and [NodeJS supports this natively](https://nodejs.org/api/packages.html#packages_conditional_exports). Currently, the behavior is that the CJS bundle gets picked up even if the bundle target is ESM.

Suggested solution would be to use `import`/`require` over `module`. (We can still retain "default")

dai-shi:
close #168

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,29 +1,29 @@
 {
   "index.js": {
-    "bundled": 4603,
-    "minified": 2099,
-    "gzipped": 917,
+    "bundled": 4608,
+    "minified": 2095,
+    "gzipped": 919,
     "treeshaked": {
       "rollup": {
         "code": 169,
         "import_statements": 59
       },
       "webpack": {
-        "code": 1320
+        "code": 1302
       }
     }
   },
   "vanilla.js": {
-    "bundled": 5875,
-    "minified": 2857,
-    "gzipped": 1104,
+    "bundled": 5977,
+    "minified": 2959,
+    "gzipped": 1139,
     "treeshaked": {
       "rollup": {
         "code": 22,
         "import_statements": 22
       },
       "webpack": {
-        "code": 1087
+        "code": 1189
       }
     }
   },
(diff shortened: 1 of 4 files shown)


dai-shi:
@LXSMNSYC Does it look OK? Can you confirm?

LXSMNSYC:
I'll try and test it out locally

LXSMNSYC:
Sorry for the late reply.

It seems that it's not just the Symbols that are supposed to be the shared state between the ESM and CJS builds, here are others:

https://github.com/pmndrs/valtio/blob/master/src/vanilla.ts#L10

https://github.com/pmndrs/valtio/blob/master/src/vanilla.ts#L29-L36

https://github.com/pmndrs/valtio/blob/master/src/react.ts#L39-L48

At this point, I suggest to drop the ESM build and follow the 1st approach: https://nodejs.org/api/packages.html#packages_approach_1_use_an_es_module_wrapper

the only downside here is that we have to manually redefine all exports if the exports changes.

About the package.json, everything seems fine.

dai-shi:
Oh, good catch. Yeah, they are module states.

I'm skeptical about the `approach #1`.
```js
// ./node_modules/pkg/wrapper.mjs
import cjsModule from './index.cjs';
export const name = cjsModule.name;
```
This may work in node, but not sure all bundlers support this. importing `.cjs` seems non-traditional.

At this point, I guess we should accept dual package hazard, and try to educate community and library authors.

LXSMNSYC:
> This may work in node, but not sure all bundlers support this.
In ESBuild and Rollup, this would work for both CJS and ESM formats.

>  importing .cjs seems non-traditional.
The ESM -> CJS bridging is an anti-pattern. An example of this would be React. However, it lead React to not have tree-shaking

LXSMNSYC (approved):

Issue #171, opened 2021-07-08
Code comments / architecture explanation


akutruff:
Code comments and a high level design doc would be really appreciated.  I'm trying to grock what's going on and there's quite a bit of code in each function  / it's unclear what's happening in each layer.

For example:
https://github.com/pmndrs/valtio/blob/bfbf0c56ecb2707d82b92032e9a345f059820ea2/src/react.ts#L126 

My eyes are crossing trying to drill through this even before getting into the vanilla layer.

I'd start with useMutableSource and how it is used.  What's the snapshot supposed to be.  I know you're emulating the new API in React, but I can't make out what's actually in a snapshot.  It's a proxy?  When if ever is data copied and at what layer?

dai-shi:
valtio has two kinds of proxies, for write and read. We intentionally separate them for hooks and concurrent react.

`proxy()` creates a proxy object to detect mutation, aka `set` handler.
`snapshot()` creates an immutable object from the proxy object.
`useSnapshot()` wraps the snapshot object again with another proxy (with `proxy-compare`) to detect property access, aka `get` handler.

both kinds of proxies create nested proxies on demand (and that's a tough task, which makes code hard.)

I guess you want to start from `vanilla.ts`.

akutruff:
Thanks for that explanation.  It's a good start and I encourage you to put that in a document or in the code comments and expand a lot more.   It's your project of course, and if you feel the code is clear enough feel free to close this issue.

dai-shi:
Let me add some code comments and hope to get some feedback from you. I would say the code is not very readable without high level abstraction.

dai-shi:
Although I'm not sure how much is covered, #174 adds something.

akutruff:
I'd add interaction and data flows to a .md file or in the comments.  The change you put in are general details but for a person who doesn't know the code won't get much from.  Since you're not having high level abstraction in the actual code, you can provide that abstraction through comments.  Pretend you have no idea what anything is when documenting the code as a newbie like myself won't be able to connect the dots.

1. Starting with `useSnapshot()` what proxies are created and what are their conceptual names?  Briefly, how is `useMutableSource()` source used to setup the subscription?  It looks like there are listeners on a proxy, for deep property changes, how are those stored and tracked?  What will trigger the subscription to be fired?  Be detailed here.  
[...]

PR #174, opened 2021-07-14
chore: add overview comment

dai-shi:
suggested in #171


dai-shi:
Moving the effort to https://github.com/pmndrs/valtio/wiki/How-valtio-works.

PR #177, opened 2021-07-16
feat: subscribe ops

### LINKED ISSUE #155: Immer-like patches
Hi. Any chance this library could support something like immer patches? It would be an efficient mechanism to implement undo history and distribute state changes (e.g. via websocket), since only state deltas are stored.

Something similar was mentioned here https://github.com/pmndrs/valtio/issues/89.

Immer patches for references https://immerjs.github.io/immer/patches/.

dai-shi:
close #155 

I have been hesitating to add a feature to core. I kept saying it's not how valtio works.

@zcaudate suggested an api https://github.com/pmndrs/valtio/issues/155#issuecomment-881150875
and it's fairly reasonable.

Note again that this change is not necessary for useMutableSource, and for useSnapshot, this is a pure overhead (in terms of both cpu and bundle size).
Nevertheless, this is something that only core can do, so I'd accept it.

Hope someone can create a history(undo/redo) feature in utils.


dai-shi:
Oops, I forgot to add a value.

.size-snapshot.json
```diff
@@ -14,9 +14,9 @@
     }
   },
   "vanilla.js": {
-    "bundled": 5875,
-    "minified": 2857,
-    "gzipped": 1104,
+    "bundled": 6611,
+    "minified": 3116,
+    "gzipped": 1214,
```
dai-shi, line 19: overhead in bundle size.

dai-shi:
The benchmark result. Memory usage is 5% increase for some cases.

image

zcaudate:
oh wow, it's being added directly to `subscribe`.  

I have a few thoughts and questions:

- is it enabled in subscribeKey?
- `swap-rows` seem to be the most affected but the rest of the slowdown look reasonable. It will affect everyone that's using this though. What do you think? How about adding an optional flag.
- why is select row slower?
- I think it's fair to assume that memory will increase a bit as there is the the added overhead of the deltas but it should pretty much be garbage collected if not used but this might be another case for adding a flag to enable this.

dai-shi:
You know, at pmndrs, we follow the idea of minimal and intuitive apis, and having optional flags isn't very nice.
If I were to need such flags, I wouldn't add this at all. (and fallback to utils-only solution.)

I assume `+-` shows a variance or something, so the left most table of the benchmark result may not tell much difference, or we'd need to run multiple times and compare.

Yeah, we could re-implement `subscribeKey`. Let me try.

zcaudate:
looking really good. thanks for your quick responses and for all your work!

PR #180, opened 2021-07-22
fix(readme): explain how `this` works with snapshot

### LINKED ISSUE #178: Frozen page when proxying an instance of EventEmitter
From https://github.com/pmndrs/valtio/pull/62#issuecomment-882045006

The following code will result in:

```
Uncaught TypeError: 'set' on proxy: trap returned falsish for property 'Notification'
```

```js
import EventEmitter from 'events';

export const state = proxy({
  signals: new EventEmitter()
});

// anywhere
const { signals } = useSnapshot(state);
useEffect(() => {
 signals.on('Notification', fn);
 return () => {
   return signals.off('Notification', fn);
 };
}, [signals)
```

**Workaround**

Avoid proxying `EventEmitter` with `ref()`

```js
import { proxy, ref } from 'valtio';

export const state = proxy({
  isConnecting: false,
  signals: ref(new EventEmitter()),
  connection: null
});
```

I'd like to understand why it fails so badly. I'd assume that the proxy handles it transparently.

dai-shi:
close #178

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -128,9 +128,11 @@ function App() {
 
 #### Holding objects in state without tracking them
 
-This may be useful if you have large, nested objects with accessors that you don't want to proxy. `ref` allows you to keep these objects inside the state model.
+Plain objects and arrays are fully supported with proxy and snapshot. Objects with prototypes may not work for some cases with some unexpected errors. In such cases, you want to bypass tracking with `ref`.
 
-See https://github.com/pmndrs/valtio/pull/62 for more information.
+This may also be useful if you have large, nested objects with accessors that you don't want to proxy. `ref` allows you to keep these objects inside the state model.
+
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -128,9 +128,11 @@ function App() {
 
 #### Holding objects in state without tracking them
 
-This may be useful if you have large, nested objects with accessors that you don't want to proxy. `ref` allows you to keep these objects inside the state model.
+Plain objects and arrays are fully supported with proxy and snapshot. Objects with prototypes may not work for some cases with some unexpected errors. In such cases, you want to bypass tracking with `ref`.
```
StarpTech, line 131: For me, it still sounds scary that correct behavior can't be ensured. I'd add a more precise reason.

dai-shi: Yeah, on second thought, this doesn't add any new insight. I will revert it.

PR #181, opened 2021-07-22
feat: proxyWithHistory

### LINKED ISSUE #89: A new util method `proxyWithHistory` with history support
ref: https://twitter.com/dai_shi/status/1361147616000544769

It would be nice. I don't have any idea yet.

dai-shi:
close #89 

There would be various ways to implement history features.
This one is a simple one with snapshots.
It's mostly syntax sugar and one can do something similar on user end.
So, it's more or less a reference implementation.
Everything is in the proxy object, so extending it would be easier.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -293,6 +293,21 @@ const state = proxyWithComputed({
 })
 ```
 
+#### `proxyWithHistory` util
+
+This is a utility function to create a proxy with snapshot history.
+
+```js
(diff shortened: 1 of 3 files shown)


zcaudate:
awesome. is the new release going to be push out?

dai-shi:
Yeah, after merging two pending PRs.

Issue #179, opened 2021-07-19
Regression with Snowpack from 1.0.6 -> 1.0.7


water-a:
# 🐛 bug report

## 🎛 Configuration (.babelrc, package.json, cli command)

index.html
```html
<html>
  <body>
    <script type="module" src="/index.js"></script>
  </body>
</html>
```

index.js
```js
import { proxy } from "valtio";
console.log(proxy);
```

package.json
```json
{
  "name": "reproduce",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "dependencies": {
    "react": "^17.0.2",
    "snowpack": "^3.8.2",
    "valtio": "^1.0.7"
  },
  "devDependencies": {},
  "scripts": {
    "start": "snowpack dev",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}
```

Commands to run
```
npm i 
npm start
```

## 🤔 Expected Behavior
It should properly imports the `proxy` export from Valtio.

## 😯 Current Behavior
![image](https://user-images.githubusercontent.com/38114545/126111602-6e74412a-fd07-4567-9fdd-078673ab14f1.png)

## 🔦 Context
There seems to be a regression when using the latest valtio version and using it with Snowpack.

However if you downgrade to valtio@1.0.6, it properly imports the `proxy` export.
```
npm i valtio@1.0.6
npm start
```

dai-shi:
Thanks for reporting. I was able to reproduce the issue with your instruction.

Looking at http://localhost:8080/_snowpack/pkg/valtio.v1.0.7.js,
it's using the esm/index.js, but from there it uses the cjs version of vanilla.js instead of esm/vanilla.js.

I'm not sure what is wrong. v1.0.6 was working because everything was cjs. As snowpack supports esm, we'd not want to use cjs. Hmmm..

cc @LXSMNSYC

LXSMNSYC:
It seems that snowpack uses the Node's default behavior:

> Node.js treats JavaScript code as CommonJS modules by default. Authors can tell Node.js to treat JavaScript code as ECMAScript modules via the .mjs file extension, the package.json "type" field, or the --input-type flag. See Modules: Packages for more details.

so even if the package is sufficiently defined with exports, Node would still need the `"type"` field or the `mjs` file extension.

[...]

PR #184, opened 2021-07-26
fix(build): Hybrid package change for ESM and CJS support

barelyhuman:
- Replicates: https://github.com/pmndrs/zustand/pull/507
- Attempts to fix:  https://github.com/pmndrs/valtio/issues/179

Browserify Working Example : https://codesandbox.io/s/react-browserify-forked-2ikqw?file=/src/index.js (Since the generated template isn't updated to work properly)

Working Snowpack Example: https://codesandbox.io/s/react-snowpack-forked-nq77c?file=/src/index.jsx for the attempted fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "version": "1.1.0",
   "description": "💊 Valtio makes proxy-state simple for React and Vanilla",
-  "main": "index.js",
+  "main": "./index.js",
   "module": "./esm/index.mjs",
   "types": "index.d.ts",
   "typesVersions": {
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "1.1.0",
   "description": "💊 Valtio makes proxy-state simple for React and Vanilla",
   "main": "index.js",
-  "module": "esm/index.js",
+  "module": "./esm/index.mjs",
   "types": "index.d.ts",
   "typesVersions": {
     "<4.0": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.1.0",
   "description": "💊 Valtio makes proxy-state simple for React and Vanilla",
   "main": "index.js",
```
dai-shi, line 6:
```suggestion
  "main": "./index.js",
```

package.json
```diff
@@ -20,26 +20,26 @@
     "./package.json": "./package.json",
```
dai-shi, line 20:
```suggestion
    "./package.json": "./package.json",
    "./": "./",
```

src/index.ts
```diff
@@ -1,2 +1,8 @@
-export * from './vanilla'
-export * from './react'
+export { useSnapshot } from "./react"
```
dai-shi, line 1:
```suggestion
// export * from './vanilla'
// export * from './react'
// NOTE snowpack doesn't allow export *, so we name all instead
export { useSnapshot } from "./react"
```

barelyhuman: I already added this, I think

src/index.ts
```diff
@@ -1,2 +1,9 @@
...
-export * from './react'
+/* 
+  Dynamic re-exports blocked to support snowpack ts re-export issues
+*/
+// export * from "./vanilla"
+// export * from "./react"
+
+export { useSnapshot } from './react'
+
+export { ref, proxy, getVersion, subscribe, snapshot } from './vanilla'
```
dai-shi, line 9:
```suggestion
export { ref, proxy, getVersion, subscribe, snapshot } from './vanilla'
export { useSnapshot } from './react'
```

just a preference..

dai-shi: actually,
```ts
export * from './react'
```
is not a problem?

barelyhuman: Yeah, surprisingly that works but I did it for consistency in the export
As for how I got to know that this fix works is when I manually added the named exports in the compiled ESM file
talking about 
`dist/esm/index.mjs` and it imported the contents of compiled `react.ts` but not for vanilla and since the rollup was aliasing vanilla I didn't want to remove it so I added the named exports manually to test and it worked. 

There is one another solution, if you want me to try that, I can do that on my fork and send through the links for the same

dai-shi: I can understand the preference on the consistency.
I think the current workaround is good enough.

dai-shi:
I suppose this is ready for review.

dai-shi (approved):
Thanks for your contribution!

src/index.ts
dai-shi: @barelyhuman I change my mind. We are fine with named re-exports, which is more explicit. Let me remove the comments.

barelyhuman: Cool 👍🏻

Issue #191, opened 2021-07-28
Could not find a declaration file for module 'valtio/utils'. 


Nunatic02:
Using `valtio/utils` with typescript like 

`import { proxyWithComputed } from "valtio/utils"` 

will result in error 

```
Could not find a declaration file for module 'valtio/utils'. '/Users/me/valtioProject/node_modules/valtio/utils.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/valtio` if it exists or add a new declaration (.d.ts) file containing `declare module 'valtio/utils';`ts(7016)
```

And `@types/valtio` doesn't exist.

Could you please add `utils.d.ts`? Thank!

dai-shi:
Which version has this issue please?

nosferatu500:
the same with `proxyWithHistory`.

Version 1.1.0 @dai-shi

dai-shi:
Does this fix with v1.0.6?

#184 is probably something different, but you can try the "Local install instructions" in https://ci.codesandbox.io/status/pmndrs/valtio/pr/184

---
Can anyone help?
@barelyhuman Do you have any insight?

barelyhuman:
@dai-shi  the utils.d.ts is never added to the dist folder that's the reason for the issue, it's for both the esm exports and the cjs exports, want me to raise a PR for the same?

dai-shi:
Whoa! Didn't know that. How could that happen?
Hmm, v1.0.7 has `utils.d.ts`. So, it's the only issue in v1.1.0.

barelyhuman:
@dai-shi it's a small change, the `utils.ts` uses a type from `vanilla.ts` but vanilla.ts doesn't export it and so the file never gets generated due to that error. 

the type is `AsRef` 
you just need to export it from `vanilla.ts` and build, check if utils.d.ts is built, if not force import the type in `utils.ts` in the `import type` line.

the build should then work

dai-shi:
So,
https://github.com/pmndrs/valtio/commit/40de10439ace8dff1a383a9b9453aa856bd9171a#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R2
this caused the issue...

PR #193, opened 2021-07-28
fix(utils): declaration generation for utils.ts

barelyhuman:
```
inline typecasted the ref output to fix the type generation blocker
```

Attempts a fix for 
https://github.com/pmndrs/valtio/issues/191


dai-shi (approved):
Thanks so much for the fix!

Issue #196, opened 2021-07-30
Can't import the named export on valtio 1.1.1


fortezhuo:
I have found bugs while upgrade valtio 1.1.1. And the problem is solved while I downgrade to 1.1.0

```
ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 88:23-41
Can't import the named export 'affectedToPathList' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 133:9-20
Can't import the named export 'createProxy' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 95:73-83
Can't import the named export 'getVersion' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 110:57-66
Can't import the named export 'isChanged' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 120:35-44
Can't import the named export 'isChanged' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 106:50-58
Can't import the named export 'snapshot' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 117:25-33
Can't import the named export 'snapshot' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 128:71-79
Can't import the named export 'snapshot' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 116:48-57
Can't import the named export 'subscribe' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 116:14-25
Can't import the named export 'useCallback' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 90:2-15
Can't import the named export 'useDebugValue' from non EcmaScript module (only default export is available)

ERROR in /Volumes/External/Workspace/project/node_modules/valtio/esm/index.mjs 44:2-11
[...]

PR #197, opened 2021-07-30
fix(pkg-exports): .mjs to .js

barelyhuman:
```
figured that .mjs is not needed for most bundlers and
causing issues with babel-loader for webpack so changed it back
```

For: https://github.com/pmndrs/valtio/issues/196#issuecomment-889756239
Sandboxes tested on PR: https://github.com/barelyhuman/valtio/pull/3

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "💊 Valtio makes proxy-state simple for React and Vanilla",
   "main": "./index.js",
   "module": "./esm/index.js",
-  "types": "index.d.ts",
+  "types": "./index.d.ts",
   "typesVersions": {
     "<4.0": {
       "esm/*": [
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "1.1.1",
   "description": "💊 Valtio makes proxy-state simple for React and Vanilla",
   "main": "./index.js",
-  "module": "./esm/index.mjs",
+  "module": "./esm/index.js",
   "types": "index.d.ts",
   "typesVersions": {
     "<4.0": {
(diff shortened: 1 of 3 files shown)


dai-shi:
codesandbox ci is failing. if it was temporary, adding an empty commit will trigger a retry.

dai-shi:
Note: this is reverting some changes in #184.

package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.1.1",
   "description": "💊 Valtio makes proxy-state simple for React and Vanilla",
   "main": "./index.js",
-  "module": "./esm/index.mjs",
+  "module": "./esm/index.js",
   "types": "index.d.ts",
```
dai-shi, line 8:
```suggestion
  "types": "./index.d.ts",
```
maybe this looks more consistent?

package.json
```diff
@@ -4,8 +4,8 @@
   "version": "1.1.1",
   "description": "💊 Valtio makes proxy-state simple for React and Vanilla",
   "main": "./index.js",
-  "module": "./esm/index.mjs",
-  "types": "index.d.ts",
+  "module": "./esm/index.js",
```
dai-shi, line 7: One more: "type" is missing. it's in https://github.com/pmndrs/zustand/pull/507/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R8

```suggestion
  "module": "./esm/index.js",
  "type": "commonjs",
```
We forgot to add in #184?

barelyhuman: Not forgot, 
during the zustand fixes I was testing out every possible scenario for that issue and was constantly change the type of the outer package and inner so everything was forced, since I understood the layout by the time I worked with valtio I didn't add it in since it's defaulted to commonjs and the `esm/` folder has a forced `type:module` anyway.

dai-shi: cool. that's what I had expected. so, we can remove it in https://github.com/pmndrs/zustand/pull/507 ?

barelyhuman: done

dai-shi (approved):

PR #201, opened 2021-07-31
feat: derive util

### LINKED ISSUE #176: adding a `derive` function
I know there's addComputed but it'd be great to have a `derive` function based off of `subscribe` to create dependent proxies.

### LINKED ISSUE #198: passing a `notifyInSync` flag to addComputed
I noticed that `addComputed` also uses subscribe.

https://github.com/pmndrs/valtio/blob/master/src/utils.ts#L192

Can a `notifyInSync` flag be passed into the function as well?

dai-shi:
close #176 
close #198 

This will deprecate `addComputed` in favor of new `derive` and existing `proxyWithComputed`.

- [x] impl
- [x] tests
- [x] deprecate `addComputed` with fallback impl
- [x] docs


zcaudate:
awesome. I think after this it's gonna be about the fifth time I've rewritten my store lib.

zcaudate:
I like what you did with `addCompute` emulation. It's pretty cool.

dai-shi:
@zcaudate Please use this **seriously** before we merge it. I'd like to avoid seeing you report issues after releasing this.

zcaudate:
@zcaudate<https://github.com/zcaudate> Please use this seriously before we merge it. I'd like to avoid seeing you report issues after releasing this

Yes Sir. Where is the sandboxed version that I can test?

dai-shi:
https://ci.codesandbox.io/status/pmndrs/valtio/pr/201
Here you are. You can also install locally. See "Local Install Instructions"

src/utils/derive.ts
```diff
@@ -0,0 +1,135 @@
...
+ * })
+ */
+export const derive = <T extends object, U extends object>(
+  derivedFns: {
+    [K in keyof U]: (get: DeriveGet) => U[K]
+  },
+  options?: {
+    proxy?: T
+    sync?: boolean
+    registerCleanup?: (cleanup: () => void) => void
```
dai-shi, line 36: I would like to have a cuter option name for this, if any.

zcaudate:
How does cleanup/unsubscribe work for a derived proxy anyways?

I usually call my methods `setup` and `teardown` if that’s any help.

On 31 Jul 2021, at 11:40 pm, Daishi Kato ***@***.******@***.***>> wrote:



@dai-shi commented on this pull request.

________________________________

In src/utils/derive.ts<https://github.com/pmndrs/valtio/pull/201#discussion_r680373862>:

> + * })
+ *
+ * derive({
+ *   tripled: (get) => get(state).count * 3,
+ * }, {
+ *   proxy: state,
+ * })
+ */
+export const derive = <T extends object, U extends object>(
+  derivedFns: {
+    [K in keyof U]: (get: DeriveGet) => U[K]
+  },
+  options?: {
+    proxy?: T
+    sync?: boolean
+    registerCleanup?: (cleanup: () => void) => void


I would like to have a cuter option name for this, if any.

zcaudate:
Okay this is super embarrassing. I’m in China and for some reason my VPN decided to stop working 2 days ago so I’ve been relying on whatever the great firewall is letting through for the last couple of days.

It’s not letting me access the link.

[cid:7D49DA79-8EAB-49FF-90D0-2FF9DE4790DB]


On 31 Jul 2021, at 11:38 pm, Daishi Kato ***@***.******@***.***>> wrote:



https://ci.codesandbox.io/status/pmndrs/valtio/pr/201
Here you are. You can also install locally. See "Local Install Instructions"

zcaudate:
I’ll try tomorrow.

dai-shi:
> How does cleanup/unsubscribe work for a derived proxy anyways?

Even if you no longer use a derived proxy, it's not garbage collected. It should be an edge case, but I would like to provide a way to unsubscribe.
It can be `destroy`, but `registerFoo` is a bit long.

dai-shi:
> It’s not letting me access the link.

well then, you can clone/check out the repo and `yarn build`, if you have access to git.

zcaudate:
okay, so doing some quick and dirty tests with the repl, this gives me an error:

```
ut = require('valtio/utils')
vanilla = require('valtio')
proxy = vanilla.proxy
var s0 = proxy({
  text: '',
  count: 0,
})

const computeDouble = (x) => x * 2

var s1 = ut.derive({
  doubled: (get) => computeDouble(get(s0).count),
})
```

```
Uncaught Error: Please use proxy object
    at getVersion (/Users/chris/Development/buffer/valtio/dist/utils.js:208:11)
    at get (/Users/chris/Development/buffer/valtio/dist/utils.js:462:29)
    at doubled (REPL13:2:35)
    at evaluate (/Users/chris/Development/buffer/valtio/dist/utils.js:466:19)
    at /Users/chris/Development/buffer/valtio/dist/utils.js:498:5
    at Array.forEach (<anonymous>)
    at Object.derive (/Users/chris/Development/buffer/valtio/dist/utils.js:440:27)
>
```


On 31 Jul 2021, at 11:57 pm, Daishi Kato ***@***.******@***.***>> wrote:



It’s not letting me access the link.

well then, you can clone/check out the repo and yarn build, if you have access to git.

dai-shi:
That seems weird. It's covered by a test.
Not sure if it's build issue, but can't dig into it without a repro.

zcaudate:
Yeah I know. I was following the test examples.

To reproduce my steps:

 - I cloned the repo, 
 - switch to the feature branch, 
 - do yarn build, 
 - cd to the ./dist folder 
 - launched a node repl there.
 - copy and paste the above script into the repl.

———

I was getting some type warnings on the build though.

```
[build:macro] (!) Plugin typescript: @rollup/plugin-typescript TS2345: Argument of type 'VariableDeclaration' is not assignable to parameter of type 'Node | NodePath<Node>'.
[build:macro]   Type ***@***.******@***.******@***.***/types/lib/index").VariableDec
laration' is not assignable to type 'babel.types.VariableDeclaration'.                                                                                              
[build:macro]     Types of property 'declarations' are incompatible.                                                                  
[build:macro]       Type ***@***.******@***.******@***.***/types/lib/index").Variabl
eDeclarator[]' is not assignable to type 'babel.types.VariableDeclarator[]'.                                                                                        
[build:macro]         Type ***@***.******@***.******@***.***/types/lib/index").Varia
bleDeclarator' is not assignable to type 'babel.types.VariableDeclarator'.                                                                                          
[build:macro]           Types of property 'id' are incompatible.                                                                      
[build:macro]             Type ***@***.******@***.******@***.***/types/lib/index").L
Val' is not assignable to type 'babel.types.LVal'.                                                                                                                  
[build:macro]               Type 'Identifier' is not assignable to type 'LVal'.                                                       
[build:macro]                 Type ***@***.******@***.******@***.***/types/lib/index
").Identifier' is not assignable to type 'babel.types.Identifier'.                                                                                                  
[build:macro]                   Types of property 'decorators' are incompatible.                                                      
[build:macro]                     Type ***@***.******@***.******@***.***/types/lib/i
ndex").Decorator[] | null | undefined' is not assignable to type 'babel.types.Decorator[] | null | undefined'.                                                      
[build:macro]                       Type ***@***.******@***.******@***.***/types/lib
/index").Decorator[]' is not assignable to type 'babel.types.Decorator[]'.                                                                                          
[build:macro]                         Type ***@***.******@***.******@***.***/types/l
ib/index").Decorator' is not assignable to type 'babel.types.Decorator'.                                                                                            
[build:macro]                           Types of property 'expression' are incompatible.                                                
[build:macro]                             Type ***@***.******@***.******@***.***/typ
es/lib/index").Expression' is not assignable to type 'babel.types.Expression'.                                                                                      
[build:macro]                               Type 'Identifier' is not assignable to type 'Expression'.                                   
[build:macro] src/macro.ts: (14:7)                                                                                                    
[build:macro]                                                                                                                         
[build:macro] 14       t.variableDeclaration('const', [                                                                               
[build:macro]          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                               
[build:macro] 15         t.variableDeclarator(snap, t.callExpression(hook, [proxy])),                                                 
[build:macro]    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                 
[build:macro] 16       ])                                                                                                               
[build:macro]    ~~~~~~~~   
```

dai-shi:
Ah, I can repro now.
...and it's a build issue.

The issue is already in master, so let me fix there and merge it into here.
Thanks for this.

dai-shi:
@zcaudate It should  be fixed now.

zcaudate:
Yep. It’s working for me now. 

I had a play with it and it seems fine. Is there a way to undo the derive? This comes in handy when derive is used on itself.

```
ut = require('valtio/utils') 
vanilla = require('valtio') 
proxy = vanilla.proxy 
var s0 = proxy({ 
text: '', 
count: 0, 
}) 

const computeDouble = (x) => x * 2 

var s1 = ut.derive({ 
doubled: (get) => computeDouble(get(s0).count), 
})

var s2 = ut.derive({ 
ddoubled: (get) => computeDouble(get(s1).doubled), 
}, {proxy: s1})

s2.undoDerive("ddoubled”)
```

dai-shi:
I think it's out of the scope of the official derive, but you are free to extend/fork it on your end.
A quick question would be, if you undo derive, do you undo the base `count`? Otherwise, there's inconsistency.
If it's like overriding, it should be something similar to `atomWithDefault` in jotai.

zcaudate:
I'm not too worried about it because it's the same as addComputed. It might actually be a good thing - though subscribe returns an unsubscribe so I'd assume that derive would too.

What about letting derive return an array instead of a proxy - [val, unsubscribeFn]?

I think if derive does have undo it should just unsubscribe the watcher from its target without removing any values.

zcaudate:
basically, as of now if `proxy` key is passed in as the argument, the return value is pretty useless.

dai-shi:
I'm between returning `cleanup` and receiving `registerCleanup`.
Assuming the use case that requires cleanup is rare and returning just a proxy object should be more intuitive. minimalistic. (and technically more performant because it's granular.)
I also recommend the primary usage is **not** to pass `proxy` option. Creating a new one is safer and less error prone. You can attach the returned proxy to some existing proxies.

zcaudate:
What would the signature of `cleanup` or `registerCleanup` be?

How might there be errors when passing in the `proxy` option?

On 1 Aug 2021, at 2:18 pm, Daishi Kato ***@***.******@***.***>> wrote:



I'm between returning cleanup and receiving registerCleanup.
Assuming the use case that requires cleanup is rare and returning just a proxy object should be more intuitive. minimalistic. (and technically more performant because it's granular.)
I also recommend the primary usage is not to pass proxy option. Creating a new one is safer and less error prone. You can attach the returned proxy to some existing proxies.

dai-shi:
```ts
type Cleanup = () => void
type RegisterCleanup = (cleanup: Cleanup) => void
```

zcaudate:
okay...

how would cleanup do an unsubscribe? and when would register cleanup actually get called?

Going by our conversation, It seems to me that it's easier to implement something based off of subscribe rather be forced into some sort of paradigm with `derive`.

dai-shi:
I think we are not on the same page.

> how would cleanup do an unsubscribe? and when would register cleanup actually get called?

Sorry for the confusion. In my context, cleanup/destroy/unsubscribe are all the same. register is like returning a callback function. I think I will re-consider another way.

> it's easier to implement something based off of subscribe

How would that look like? Can you create a code snippet (even if it doesn't work)?

zcaudate:
Okay. This is not a jab or anything. It's to try and explain my point of view.

So firstly, you're doing a fantastic job with all the libraries.

Secondly, valtio rocks and it's a core library I've chosen to build around so I'm quite partial to it. 

I'm not too fond of maintaining open source code, having done a lot of that in the past - so apologies for not contributing code. However I believe that use cases and ideas are just as relevant. Use cases are how code gets shaped and hardened over time.

I got a little bit frustrated by your statements of 

- `out of the scope of the official derive ...`
- `recommend the primary usage...`

For me, once something becomes `official` it means that the thing is basically set in stone. This shouldn't be the case for an api that's not even released. The user should determine how an api is used. Without actual use cases, no one can recommend anything unless there are absolute pitfalls that need to be aware of. Just saying `this is safer and less error prone` is a religious argument.

 ----

So in terms of a code snippet, I don't really think it matters overall because it's just an added detail to one possible way valtio can be used.

As far as I can tell, there's 4 parts to the library:

1. the base store
2. eventing
3. react integration
4. store composition

1-3 are pretty solid already.
4 is very basic and rather arbitrary at the moment. 

If I were responsible for valtio (thankfully I'm not because open source libs are a lot of work), I would try to make the code base share as much code as possible whilst exposing as much functionality as I can. So in this case, I would double down on using `subscribe` wherever I can and would find a way to share code between `watch` and `derive` - probably writing one in terms with the other like you did with `addComputed`. Then I would expose all the intermediate functions that I used to construct them, including a `builder api` for both myself as well as for advanced users that may need those extra features. I wouldn't presume anything about consumers of my library. My main goal would be to make the library as easy to maintain as I can for myself - and builder functions help alot in this case.

---

Anyways. the subscribe api is really solid, especially with the diffing. There's a million different uses for this.

dai-shi:
My apologies to make you frustrated.

For me, you are the only person giving the feedback on `derive`. So, if you say you wouldn't use `derive` in utils but `subscribe` in core works for you, I was in panic.
I would like to learn how you would use `subscribe` and tune `derive` into your usage if possible.

As it sounded like you have an idea how you would do with `subscribe`, I asked a code snippet. At this point, I'm not requesting code contribution. (I have keep saying it though, including 3rd-party libs, as you have enough skills and interests.)

So, apart from `registerCleanup` which I'd re-consider, how does or does not `derive` solve your use case?

zcaudate:
I've outlined my usage building on top of subscribe here:

https://gist.github.com/zcaudate/e8c9fe85da0982d372bc856f3b03fe2b#file-store-js-L111-L188

Your feedback was that it was too level. However, I do need that finer grain control over how watches on objects are created, listed and removed. This I would considered the `builder api` - if it existed.

---

I probably wouldn't use `derive` in it's current form because of the limitations I had outlined. (I'll stick to my current approach for now).

dai-shi:
Thanks for the gist!
I think I get your idea. So, each property has a different subscription.
Right, it's not possible with `derive`. And, you do pretty well with `subscribe`.

Anyway, that was a great hint for me. Let me re-work on `registerCleanup` (= handling `unsubscribe`).
It still may not fit with your need, but that's okay. One thing cant fit all.

`derive` is something originally you named, but I can agree it's something different now, because I'm motivated to deprecate `addComputed` with it.

Thanks for your feedback anyways!

dai-shi:
So, my idea for unsubscribe without `options.registierUnsub` is `underive`.
Thanks to @zcaudate 's gist, I added `keys` option to selectively unsubscribe properties (a.k.a. partial underive).

Hope someone can try it.

zcaudate:
A couple more requests from me.

Can you export `getSubscriptions` as well as move `addSubscription` `removeSubscription ` out of the class and export those as well. `underive` can be refactored to use `removeSubscription`.

zcaudate:
also, for completeness, I think `watch` can also receive the same treatment given to `addComputed` by refactoring it to use `addSubscription`

dai-shi:
Well, I know you wanted building blocks, but I have a strong preference to hide them as implementation details. I don't want to increase public apis. Not only about `derive` but for all other functions/types in valtio. Our preferences simply conflict.

I thought about reusing removeSubscription, but thought it couldn't at some point. Maybe I can revisit.

I had thought about sharing some code with `watch`, but I thought it difficult because of `watch` supporting nesting `watch`. Maybe I'm wrong. cc @LXSMNSYC

zcaudate:
They are already there. It's my choice if I want to shoot myself in the foot. 

If you don't want to pollute the `valtio` module, just put everything into `valtio/builder`. But please just make it available because it makes a huge difference to how I will use it downstream.

I really need the `crud` functionally for subscriptions. If it's encapsulated, I'd be forced to reimplement the exact same functionality - which I have currently done. Like I said before, if it gets added in valtio, most of the code I wrote can just disappear.

zcaudate:
I could argue that because `addSubscription` etc... uses `subscribe` then technically `subscribe` is also an implementation detail that should be hidden away. 

Why is that a bad idea? Because it breaks composability. 

I think the governing principle for good software design is composability. Composability leads to cleanliness, not the other way around.

LXSMNSYC:
> Well, I know you wanted building blocks, but I have a strong preference to hide them as implementation details. I don't want to increase public apis. Not only about `derive` but for all other functions/types in valtio. Our preferences simply conflict.
> 
> I thought about reusing removeSubscription, but thought it couldn't at some point. Maybe I can revisit.
> 
> I had thought about sharing some code with `watch`, but I thought it difficult because of `watch` supporting nesting `watch`. Maybe I'm wrong. cc @LXSMNSYC

We could probably create a cleanup scope so that automatic cleanup tracking is done in a single pass and for the implementation to be scalable. This way, cleanup logic is composable (although more magic-y)

I'm thinking of

```js
const cleanup = batchCleanup(() => {
  // watch's cleanup is automatically tracked
  watch(() => {...});
  // As well as subscribe
  subscribe(state, () => {...});
  // or if you want to use your own cleanup.
  unstable_onCleanup(() => {
   // ...
  });
});
```

Although I'm not sure if this is valtio-y to do 😅

dai-shi:
> They are already there. It's my choice if I want to shoot myself in the foot.

No, because if we make it public, we need to maintain.

> If you don't want to pollute the `valtio` module, just put everything into `valtio/builder`.

This sounds possible, and we can mark "unstable" (= we don't guarantee anything.)

> it makes a huge difference to how I will use it downstream.

Can you show how it becomes different, with/without `addSubscription` in `valtio/builder`?

Simply put, I'm not very confident if `addSubscription` is a good abstraction, and I might want to change it at some time. So, for now, you want to compose functionalities with solid `subscribe` as you do now.

zcaudate:
> No, because if we make it public, we need to maintain.

You're going to have to maintain regardless. That's why it's better to write one function in terms of another. It's better that the functionality is broken up into smaller modules and be exposed to the public allowing earlier feedback if something does go wrong.

---

back to the gist here:

https://gist.github.com/zcaudate/e8c9fe85da0982d372bc856f3b03fe2b#file-store-js-L111-L188

The only difference between the functionality with this gist and the the functionality in the `derive` codebase is that I'm also `tagging` each proxy with an id. If the functionality gets exposed in derive. Then I can delete lines 111-188 and just use native valtio functionality.

dai-shi:
I see you want to know the dependencies of each derived property. That, I didn't cover. Hmm. It might be useful for debugging if dependencies are many.

zcaudate:
I have my dev environment embedded in the runtime itself (for live eval) so all these `debug` features are super handy to have.

out

dai-shi:
I think @zcaudate 's suggestion is probably right, but I'm still not confident about exposing building block functions at this point. To move forward anyway and collect more feedback from others, I will expose only getSubscriptions as an experiment with `unstable_` prefix.

dai-shi:
(Getting to understand more about the suggestion. Yeah, I will probably be confident if we make a better "subscriptions" abstraction that works for both `derive` and `watch`, and ideally with one more util based on it.)

Merging this for now.

zcaudate:
Sweet. 

For context regarding workflow, here's a video from the author of https://github.com/omcljs/om (the library that inspired redux) - https://www.youtube.com/watch?v=3HxVMGaiZbc - explaining a similar concept.

PR #202, opened 2021-07-31
fix: return value in snapshot promise getter after promise is resolved

israelidanny:
This fixes issue https://github.com/pmndrs/jotai/issues/624 with jotai/valtio where throwing a promise in a snapshot getter with the intention to suspend react render ends up causing an infinite loop by returning the value once the promise is resolved.


dai-shi (commented):
Wow, very nice catch.
The fix seems perfectly correct.

Would you be able to add a test for this?
If it's too hard, I can merge without a test.

src/vanilla.ts
```diff
@@ -104,6 +104,10 @@ export const proxy = <T extends object>(initialObject: T = {} as T): T => {
           const errorOrPromise = (value as any)[PROMISE_ERROR] || value
           Object.defineProperty(snapshot, key, {
             get() {
+              if (PROMISE_RESULT in (value as any)) {
+                return (value as any)[PROMISE_RESULT]
+              }
+
```
dai-shi, line 110:
```suggestion
```

This is just a stylistic preference.
I know it's too packed currrently and having some empty lines would help, but not here (or not only here).

dai-shi:
I fixed an issue around build tool. 93e71586a2689816e7e529cb4d0d2ce5237168bb
Please merge it and re-create .size-snapshot.json.

israelidanny:
I merged master and added a test for this exact issue

dai-shi:
Hi, I should have said generating size snapshot as follows:
```bash
rm .size-snapshot.json
yarn build -m 1
```

israelidanny:
@dai-shi done

dai-shi (approved):
Looks great! Thanks for your contribution!

PR #203, opened 2021-08-01
feat: types for states that include promises

israelidanny:
This PR addresses the typing issue of states with proxy, allowing jotai's atomWithProxy to have proper typing. It's connected to this PR over at jotai:
https://github.com/pmndrs/jotai/pull/625

It achieves this by introducing a new `SubstitutePromises` type that recursively replaces all promises with a union of `Promise<V>  | V`.


israelidanny:
@dai-shi I think that this typing solution not only neatly helps https://github.com/pmndrs/jotai/pull/625 and accurately reflects the actual behavior of the library (values indeed do change from Promise to the resolved value).

IMHO This typing is more helpful because it reflects the actual lifecycle of the state (Promise<V> turns to V).

What do you think?

dai-shi:
This is something I want to avoid in valtio library perspective. This is not valtio responsibility. It's library consumer responsibility.
```js
const state = proxy<{ foo?: Promise<number> }>({})
state.foo = Promise.resolve(0) // this is ok
state.foo = 0 // this should fail (TS error)
```

dai-shi:
> values indeed do change from Promise to the resolved value

No, it doesn't change.

israelidanny:
@dai-shi If I understand you correctly, you are saying that the change only occurs on snapshots? In this case it would be useful to update the snapshot to have accurate typing like this - but keep this typing out of the actual proxy. I still think it's a benefit to have accurate typing on snapshots (which do change into resolved value) - do you agree? 

I think it can be rather easily adjusted.

dai-shi:
So, my perspective is https://github.com/pmndrs/jotai/pull/625 is jotai/valtio's problem. valtio itself is just fine as is.

dai-shi:
We already have DeepResolveType.

israelidanny:
@dai-shi I see. Makes sense. We can close this, I guess? I'll resume work over at jotai/valtio

dai-shi:
Yes, thanks for your understanding.

PR #213, opened 2021-08-12
Change master branch to main in everywhere

soilSpoon:


soilSpoon:
Sorry, I didn't consider the other parts in #212, so there were 2 PRs.

examples/counter/src/App.js
```diff
@@ -85,6 +85,26 @@ const Controls = () => {
...
+
+const ButtonDown = ({ onClick }) => (
+  <svg viewBox="0 0 430 452" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path
+      onClick={onClick}
+      d="M214.83 451.523C347.216 451.523 429.602 360.898 429.602 226.523C429.602 187.816 422.852 152.786 410.112 122.5L106 426.214C136.689 442.604 173.299 451.523 214.83 451.523Z"
+      fill="#FFBEC2"
+    />
+  </svg>
+)
```
dai-shi, line 107: Oh, you want to put them in. Okay, why not.

dai-shi (approved):
Thanks again for the fix. We would need to to the same in jotai repo.

examples/counter/src/App.js
soilSpoon: > Oh, you want to put them in. Okay, why not.

I'm sorry, I'm not good at English, can you explain in more detail?

dai-shi: It was intentional to avoid having `ButtonUp` and `ButtonDown` there. But, we can have them.

soilSpoon: I added it using ButtonUp and ButtonDown in lines 83~84.

Is this the problem now?

dai-shi: No, it's not a problem. I will merge this.
(As there was an intention not to have it, we may remove it in the future.)

PR #215, opened 2021-08-17
fixing subscribeKey typing

Andarius:
Fixes the type of the callback value of `subscribeKey`


dai-shi (approved):
Nice catch. I overlooked this. Thanks!

PR #227, opened 2021-08-28
fix(utils): Adds skipSubscribe option to proxyWithHistory

promontis:


dai-shi (changes requested):
Hi, thanks for suggestion. I wanted proxyWithHistory to be fully customizable, but this part `subscribe` wasn't customizable.
Here're suggestions to cover your use case and more.
Please let me know if you have a better name for the second argument.

promontis:
@dai-shi really liked your suggestions! I think `skipSubscribe` is a better name, can't think of a better one.

dai-shi:
oops, would you please run prettier?

promontis:
@dai-shi done

dai-shi (approved):

Issue #229, opened 2021-08-29
Use action names in redux devtools if action methods are defined in state


geocine:
If I am right, it is close to impossible to use action names on redux devtools if actions are defined in a module, since valtio wouldn't have an idea what these method names are.

However, would it be possible to show action names on redux devtools if they are defined in state, instead of using the generic `Update ...` message? It would atleast help for debugging purposes.

![Screenshot 2021-08-29 at 6 33 13 PM](https://user-images.githubusercontent.com/507464/131247406-99e30eb3-7888-4fb9-8f99-6107b190aa9e.png)

dai-shi:
As we have #177 now, it might be possible to show more info (not action names but updated paths). Let me try something naively.

geocine:
That would be cool and much more explicit , I did not thought about that . My only concern about that is we can modify exactly the same paths on any location. It would still be hard for debugging if we would want to find which triggered the changed among all those methods/actions/statements that would seem to change the exact same paths.

I don't have an answer, but that is the problem that I would probably run into in the long term.

dai-shi:
From proxy point of view, we can't know how it's changed, only the result of the change.

Please try #230.
https://ci.codesandbox.io/status/pmndrs/valtio/pr/230
See "Local Install Instructions" in ☝️ .
I didn't test it, and I'm not sure if it works.

geocine:
It works, thank you!
![Screenshot 2021-08-29 at 7 05 31 PM](https://user-images.githubusercontent.com/507464/131248249-a1b64e0c-44a5-4007-84d9-2d4ad912f760.png)

dai-shi:
Great to hear that.
Isn't it too long? Do we need the datetime?
I wonder if we make the action an object, it would make it better.

geocine:
Yes it is long but not sure how we could improve that. I also noticed some additional information showing app when skip is clicked. I don't really know if these information needs to go to the devtools

![Screenshot 2021-08-29 at 7 47 47 PM](https://user-images.githubusercontent.com/507464/131249327-27e1454d-0802-46a3-bc80-2f5e69abe2f0.png)

dai-shi:
Oh, that doesn't look nice. Did it happen before this PR?

---
I can take a deeper look some time later, but not soon.
If there's anyone who would like to contribute on it, I'm happy to discuss.

geocine:
Not exactly but yes. 

[...]

PR #230, opened 2021-08-29
fix(utils): devtools to show ops as action

dai-shi:
for #229


dai-shi:
closing as #232 is continued work.

PR #232, opened 2021-09-03
Fix: remove useless internal info in devtools

Aslemammad:
I just added a simple symbol to see what's @dai-shi's opinion on that! #229

### DIFF
--- a/src/utils/devtools.ts
+++ b/src/utils/devtools.ts
@@ -36,7 +36,7 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
   const unsub1 = subscribe(proxyObject, (ops) => {
     const action = ops
       .filter(([_, path]) => path[0] !== DEVTOOLS)
-      .flatMap(([op, path]) => `${op}:${path.map(String).join('.')}`)
+      .map(([op, path]) => `${op}:${path.map(String).join('.')}`)
       .join(', ')
 
     if (!action) {
@@ -48,11 +48,10 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
     } else {
       const proxyWithoutDevtools = Object.assign({}, snapshot(proxyObject))
       delete (proxyWithoutDevtools as any)[DEVTOOLS]
-
-      devtools.send(
-        `${action} - ${new Date().toLocaleString()}`,
-        proxyWithoutDevtools
-      )
+      devtools.send({
+        type: action,
+        updatedAt: new Date().toLocaleString(),
+      })
     }
   })
   const unsub2 = devtools.subscribe((message: Message) => {
--- a/src/utils/devtools.ts
+++ b/src/utils/devtools.ts
@@ -1,4 +1,6 @@
-import { snapshot, subscribe } from '../vanilla'
+import { snapshot, subscribe, DEVTOOLS } from '../vanilla'
+
+type Message = { type: string; payload?: any; state?: any }
 
 /**
  * devtools
@@ -41,18 +43,15 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
     }
   })
   const unsub2 = devtools.subscribe(
-    (message: { type: string; payload?: any; state?: any }) => {
+    (message: Message) => {
       if (message.type === 'DISPATCH' && message.state) {
         if (
           message.payload?.type === 'JUMP_TO_ACTION' ||
           message.payload?.type === 'JUMP_TO_STATE'
         ) {
           isTimeTraveling = true
-        }
-        const nextValue = JSON.parse(message.state)
-        Object.keys(nextValue).forEach((key) => {
-          ;(proxyObject as any)[key] = nextValue[key]
-        })
+        } 
+        ;(proxyObject as any)[DEVTOOLS] = message
       } else if (
         message.type === 'DISPATCH' &&
         message.payload?.type === 'COMMIT'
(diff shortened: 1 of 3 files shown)


dai-shi (changes requested):
we want to avoid changing core for this. i guess you can do it in devtools.ts.

src/utils/devtools.ts
```diff
@@ -32,58 +34,69 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
...
   const devtools = extension.connect({ name })
   const unsub1 = subscribe(proxyObject, (ops) => {
-    const action = ops.map(([op, path]) => `${op}:${path.join('.')}`).join(', ')
+    const action = ops
+      .flatMap(([op, path]) =>
+        path.some((x) => typeof x === 'symbol')
+          ? []
+          : [`${op}:${path.join('.')}`]
+      )
+      .join(', ')
```
dai-shi, line 43: How about this?
```js
const action = ops.map(([op, path]) => (
  `${op}:${path.map(String).join('.')}`
).join(', ')
```

src/utils/devtools.ts
```diff
@@ -32,58 +34,69 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
...
+
+    if (!action) {
+      return
+    }
```
dai-shi, line 47: Why do you need to return?

src/utils/devtools.ts
```diff
@@ -32,58 +34,69 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
...
     if (isTimeTraveling) {
       isTimeTraveling = false
     } else {
+      const proxyWithoutDevtools = Object.assign({}, proxyObject)
+      delete (proxyWithoutDevtools as any)[DEVTOOLS]
+
       devtools.send(
         `${action} - ${new Date().toLocaleString()}`,
-        snapshot(proxyObject)
+        proxyWithoutDevtools
```
dai-shi, line 56: You can't skip `snapshot()`. Please try removing DEVTOOLS after taking a snapshot.

src/utils/devtools.ts
Aslemammad: Because we don't want the log in the redux devtools, it's just an empty log.

src/utils/devtools.ts
Aslemammad: Let me try it!

src/utils/devtools.ts
dai-shi: Oh, I missed that `action` can be an empty string.

src/utils/devtools.ts
```diff
@@ -29,61 +33,68 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
 
   let isTimeTraveling = false
   const devtools = extension.connect({ name })
-  const unsub1 = subscribe(proxyObject, () => {
+  const unsub1 = subscribe(proxyObject, (ops) => {
+    const action = ops
+      .filter(([_, path]) => path[0] !== DEVTOOLS)
+      .flatMap(([op, path]) => `${op}:${path.map(String).join('.')}`)
```
dai-shi, line 39: From the code, it seems we can use `map` instead. Can you confirm?
```suggestion
      .map(([op, path]) => `${op}:${path.map(String).join('.')}`)
```

src/utils/devtools.ts
```diff
@@ -29,61 +33,68 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
...
     } else {
+      const proxyWithoutDevtools = Object.assign({}, snapshot(proxyObject))
+      delete (proxyWithoutDevtools as any)[DEVTOOLS]
+
       devtools.send(
-        `Update - ${new Date().toLocaleString()}`,
-        snapshot(proxyObject)
+        `${action} - ${new Date().toLocaleString()}`,
+        proxyWithoutDevtools
       )
```
dai-shi, line 55: I feel like we have a better way to show an action. At least, this would be good.
```js
devtools.send({
  type: action,
  updatedAt: new Date().toLocaleString(),
})
```

We could make the `action` string more compact and put full infor in the object. what do you think?

src/utils/devtools.ts
```diff
@@ -29,61 +33,68 @@ export const devtools = <T extends object>(proxyObject: T, name?: string) => {
...
-          const action =
-            actions[index] || `Update - ${new Date().toLocaleString()}`
+      isTimeTraveling = true
 
-          Object.keys(state).forEach((key) => {
-            ;(proxyObject as any)[key] = state[key]
-          })
+      computedStates.forEach(({ state }: { state: any }, index: number) => {
+        const action =
+          actions[index] || `Update - ${new Date().toLocaleString()}`
```
dai-shi, line 84: Do you know when action is not found?
It feels like this is fine or it would reveal it.
```suggestion
          actions[index] || 'No action found'
```

dai-shi (commented):
This looks a great improvement.
Can you address some smaller things commented?

src/utils/devtools.ts
Aslemammad: Using map just messed the output there!

src/utils/devtools.ts
Aslemammad: Yes, i think this is a better solution!

src/utils/devtools.ts
Aslemammad: That one is better! Done.

dai-shi (approved):

PR #257, opened 2021-10-17
feat(utils): jump state support in devtools

Aslemammad:
Inspired by https://github.com/pmndrs/zustand/pull/590 - mutating the object when the message.state is safe (no actionsById, which happens in state jumping)
It worked for me, it's up to @dai-shi if he wants to see if this works well!


dai-shi:
Would anyone like to check the behavior?

PR #267, opened 2021-11-01
Add related libraries section

Noitidart:
Just a thought, completely understand if we don't want to add this.


dai-shi:
As you implied, we don't want this for some reasons, like not having review process.
For now, let's do what did in zustand.
Would you create a wiki page like https://github.com/pmndrs/zustand/wiki/3rd-Party-Libraries with the disclaimer, and create a PR to add a link to the wiki page?
Hope you find some other libraries than yours, and list them. (I personally do have one.)

Noitidart:
Oh no problem, I'll make a wiki like that :)

Noitidart:
Done. :) https://github.com/pmndrs/valtio/wiki/3rd-Party-Libraries

PR #288, opened 2021-12-01
chore(test): Use SWC as Jest transformer

Thisen:
Using @swc/jest as the Jest transformer.

Went from 10.351 secs to 3.621 secs on my machine. That's about 65% faster.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,14 +102,10 @@
   "jest": {
     "rootDir": ".",
     "testEnvironment": "jsdom",
-    "preset": "ts-jest",
-    "globals": {
-      "ts-jest": {
-        "tsconfig": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -102,14 +102,10 @@
...
-      "ts-jest": {
-        "tsconfig": {
-          "target": "es2015",
-          "jsx": "react-jsx"
-        }
-      }
+    "transform": {
+      "^.+\\.(t|j)sx?$": [
+        "@swc/jest"
+      ]
```
dai-shi, line 108: This is nice. It's now consistent with other two.

dai-shi (approved):
Thanks!!

PR #307, opened 2022-01-04
fix: duplication of stores in HMR and multiple devtools call

### LINKED ISSUE #302: DevTools: duplication of stores
Hi,
Creating a store using `proxy` and then adding it to `devtools` is causing a weird bug where stores are duplicated when viewing in Redux chrome extension.

This is the code I'm using to create the store:
```typescript
import { devtools } from 'valtio/utils';
import { isProduction } from './env-utils';

export const createStore: <T extends ObjectOf<T>>(
  initialObject: T,
  storeName: string,
) => T = (initialObject, storeName: string) => {
  const store = proxy(initialObject);

  if (!isProduction()) {
    devtools(store, storeName);
  }
  return store;
};

// https://github.com/typescript-eslint/typescript-eslint/issues/2720#issuecomment-729762246
type ObjectOf<T> = { [P in keyof T]: T[P] };
```
**Valtio version: 1.2.7
Redux extension version: 2.17.2**

Is this a known issue? Is there a workaround for it maybe?

Thanks

Aslemammad:
I guess it is ready now!

close #302

### DIFF
--- a/src/utils/devtools.ts
+++ b/src/utils/devtools.ts
@@ -4,12 +4,6 @@ type Message = { type: string; payload?: any; state?: any }
 
 const DEVTOOLS = Symbol()
 
-const instances =
-  (window as any).__REDUX_DEVTOOLS_EXTENSION__.instances ||
-  new Map<string | undefined, any>()
-
-;(window as any).__REDUX_DEVTOOLS_EXTENSION__.instances = instances
-
 /**
  * devtools
  *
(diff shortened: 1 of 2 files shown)


dai-shi (commented):
Sorry to be late...
Hmmm, it doesn't look a good idea to use `name` as a key.
I don't think I understand the original problem. Going back to #302.

dai-shi (changes requested):
I don't feel comfortable to use name as key. It can break normal use cases.
Let's focus on HMR issue #302. I'll think about it too.

PR #312, opened 2022-01-08
fix(build): Generate umd and system files

romulo94:
Follow the thread: https://github.com/pmndrs/zustand/pull/740

Hi @dai-shi , I did the same config here.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -87,6 +87,55 @@ function createCommonJSConfig(input, output) {
   }
 }
 
+function createUMDConfig(input, output) {
+  return {
+    input,
+    output: {
+      file: output,
+      format: 'umd',
+      exports: 'named',
+      name: 'zustand',
+      esModule: true,
+    },
+    external,
+    plugins: [
+      alias({
+        entries: {
+          './vanilla': 'valtio/vanilla',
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -87,18 +87,71 @@ function createCommonJSConfig(input, output) {
...
 }
 
+function createUMDConfig(input, output) {
+  return {
+    input,
+    output: {
+      file: output,
+      format: 'umd',
+      exports: 'named',
+      name: 'zustand',
```
dai-shi, line 97:
```suggestion
      name: 'valtio',
```

rollup.config.js
```diff
@@ -87,18 +87,71 @@ function createCommonJSConfig(input, output) {
...
+      esModule: true,
+    },
+    external,
+    plugins: [
+      alias({
+        entries: {
+          './vanilla': 'valtio/vanilla',
+          '../vanilla': 'valtio/vanilla',
```
dai-shi, line 105: So, this is the trick.

romulo94: Yeah! I'm following the same config for alias

dai-shi (approved):
👍

PR #318, opened 2022-01-10
feat: provide kind of Set and Map data types in utils package

fkhadra:
Hey, as discussed #317. 

I've started to work on a `Set` implementation. I took the liberty to add more methods to the `proxySet` object, notably `map` and `filter` given how useful they are when working with React. Let me know what you think about it. I've added them as well for `Map` (not committed yet)

The current `Set` implementation is not a 1 to 1 mapping. Below are the missing methods. TBH, I don't know if it's worth having them given that I've added `map` and `filter`, I don't recall using them often. This is pretty subjective 😆
- Set.prototype[@@iterator] (iterator symbol)
- Set.prototype.values
- Set.prototype.entries

The implementation of Map will follow later.


dai-shi:
Thanks for working on this!
valtio strongly prefers providing "normal" JS API.
So, it's strongly suggested to implement all methods and no extra methods.
If it's technically impossible, let me know. (I think we'd `throw` in such cases.)
And, I'd personally use iterators, quite often. 😁

fkhadra:
Thanks for the feedback. I'll stick to the "normal" API then 👌

dai-shi (commented):
Looking good as WIP. Some first-look comments.

src/vanilla.ts
```diff
@@ -16,7 +17,9 @@ export const ref = <T extends object>(o: T): T & AsRef => {
 const isSupportedObject = (x: unknown): x is object =>
   typeof x === 'object' &&
   x !== null &&
-  (Array.isArray(x) || !(x as any)[Symbol.iterator]) &&
+  (Array.isArray(x) ||
+    (x as any)[CUSTOM_TYPE] ||
+    !(x as any)[Symbol.iterator]) &&
```
fkhadra, line 22: Given that the `proxySet` implements the iterator symbol I needed a way to bypass this check. I didn't want to expose something from `utils` inside the `core` package. So I thought that having a dedicated symbol to identity custom data types would be useful. 

It would be better to move the symbol in a dedicated file, maybe.

If you see a better way to solve this issue, I'm open to it 😊

fkhadra:
So, I've managed to be really close to the native `Set` implementation. I've implemented a test case to ensure that the output of the `proxySet` always matches with a native `Set`.

I had to do one modification in the core.

src/vanilla.ts
dai-shi: As you might guess, I don't prefer this solution. tbh, I don't like how `isSupportedObject` is hacked by me. Give me some time for this. I'd like to come up with an idea I'm happy with.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+  Object.defineProperties(set, {
+    data: {
+      writable: true,
+    },
+  })
+
+  Object.seal(set)
+
+  return set as Set<T>
```
dai-shi, line 79: This should return a proxy in my opinion:
```suggestion
  return proxy(set) as Set<T>
```

tests/proxySet.test.tsx
```diff
@@ -0,0 +1,272 @@
...
+    { id: 2, val: 'world' },
+  ]
+
+  const state = proxy({
+    set: proxySet(initialValue),
+  })
+  expect(JSON.stringify(snapshot(state).set)).toMatch(
+    JSON.stringify(initialValue)
+  )
+})
```
dai-shi, line 272: Please add tests more for objects in `Set`, and nested `Set`s.

for example:
```js
  const state = proxy({
    set: proxySet([{ a: 1 }]),
  })
  const item = Array.from(state.set.values())[0]
  item.a = 2
```

src/utils/proxySet.ts
```diff
@@ -0,0 +1,80 @@
...
+  // used for JSON.stringify
+  toJSON(): T[]
+  [CUSTOM_TYPE]: boolean
+}
+
+export function proxySet<T>(initialValues: Iterable<T> = []): Set<T> {
+  const set: InternalProxySet<T> = {
+    data: [...new Set(initialValues)],
+    has(value) {
+      return this.data.indexOf(value) !== -1
```
dai-shi, line 15: This doesn't work if the value is a non-proxy object. `proxy(value)` solves it, but the problem is `isSupportedObject` is not exposed.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,80 @@
...
+interface InternalProxySet<T> extends Set<T> {
+  data: T[]
+  // used for JSON.stringify
+  toJSON(): T[]
+  [CUSTOM_TYPE]: boolean
+}
+
+export function proxySet<T>(initialValues: Iterable<T> = []): Set<T> {
+  const set: InternalProxySet<T> = {
+    data: [...new Set(initialValues)],
```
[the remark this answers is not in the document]
dai-shi: Hm, I may think the elegance of your solution is having `data` as an array. It seemed limited, but works.

So, I think we can implement `proxyMap` with an array, too.

Yeah, that would simplify the implementation. My previous suggestion is void. Let's explore with array backed solution.

`...` on iterator doesn't work in some browsers, let's use `Array.from`.

```suggestion
    data: Array.from(new Set(initialValues)),
```

dai-shi (commented):
sorry, I changed my mind during the review. hope it's readable. I will think about the vanilla part.

src/vanilla.ts
fkhadra: Fully agree with you on this one. Can't we replace `!(x as any)[Symbol.iterator])` by explicit check for `Map` and `Set`?  If I'm not mistaken those are the built-in objects that implement the iterator: `String`, `Array`, `TypedArray`, `Map`, and `Set`

```js
!(x instanceof Map) &&
!(x instanceof Set) &&
!(x instanceof TypedArray)
```

src/utils/proxySet.ts
fkhadra: Just for my understanding. If it returns a proxy this could be used as follow
```js
const state = proxySet([1,2,3])
```
Which is equivalent to
```js
const state = proxy(proxySet([1,2,3]))
```
So returning a proxy as you suggested would allow the end-user to use the `proxySet` directly. Which is more elegant, make sense.

src/utils/proxySet.ts
fkhadra: Not sure to understand the issue here.

fkhadra:
> sorry, I changed my mind during the review. hope it's readable. I will think about the vanilla part.

No problem 😄thanks for the detailed review

src/utils/proxySet.ts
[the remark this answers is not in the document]
fkhadra: nice catch thanks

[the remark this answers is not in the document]
dai-shi: Cool. Seems like you have better understanding with your experiments than I do. Look forward to it.

src/utils/proxySet.ts
dai-shi: Actually, both are considerable.
```js
const state = proxySet(new Set([1, 2, 3])) // returns a proxy
const state = proxy(emulateSet(new Set([1, 2, 3]))
```
I just feel the name `proxySet` takes a `Set` (or any Iterable) and returns a set. If `emulateSet` (or some other name) has some use cases, it can be a good option too. We need to decide which to go.

src/utils/proxySet.ts
dai-shi:
```js
const obj = {}
const state = proxySet([obj])
state.has(obj) // ---> expecting true
```

src/vanilla.ts
dai-shi: I will work on the vanilla part soon (not strictly tied to this PR). Please review it later to see if it works for this PR.

dai-shi:
I will work on the core (vanilla.ts) in a separate PR. So, please wait for it for the integration part.
Meanwhile, I hope you can add more tests, especially for `Set` items being objects (arrays) and `Set`s.

fkhadra:
No problem. I'll work on adding more tests.

src/utils/proxySet.ts
fkhadra: This does work (tested locally) even if `proxySet` is not wrapper by `proxy`.
Screenshot 2022-01-11 at 13 18 23

src/utils/proxySet.ts
fkhadra: I like your first suggestion, the more I think about it the more It feels natural to me.

```js
const state = proxySet(new Set([1, 2, 3])) // returns a proxy
```
Also, I would say this it's less error-prone given it returns a `proxy`. 

With the latter, I think a user can forget to wrap it with `proxy`.

src/utils/proxySet.ts
dai-shi: What about wrapping with proxy?

src/utils/proxySet.ts
dai-shi: Cool. And we can still do the original usage:
```js
const state = proxy({
  set1: proxySet(new Set([1, 2, 3])),
  set2: proxySet([4, 5, 6])),
})
```

src/utils/proxySet.ts
fkhadra: I'm a bit confused 🤔. Without proxy it returns `true` but with proxy returns `false`. Is it because I'm reading from the state instead of a snapshot. I'll have to dig a bit more
 
Screenshot 2022-01-11 at 15 01 26

dai-shi:
Well, if you ask me, I'd suggest to emulate native `Set` behavior.
Basically, Set is an iterable, which can't be stringified.
You want to convert an iterable to an array before stringify.
```js
> JSON.stringify(Array.from(new Set([{ id: 1, val: 'hello' }, { id: 2, val: 'world' }])))
'[{"id":1,"val":"hello"},{"id":2,"val":"world"}]'
```

fkhadra:
Now that you mention it. It's wiser, so the end-user knows what to expect. Also, this simplifies the implementation, `toJSON` will only need to return `{}`

dai-shi:
Don't you even need to implement `toJSON`?

fkhadra:
> Don't you even need to implement `toJSON`?

Indeed it's not needed if `data` and `size` are not enumerable
```js
 Object.defineProperties(set, {
    data: {
      writable: true,
      enumerable: false,
    },
    size: {
      enumerable: false,
    },
  })
```

fkhadra:
So I've been playing with #321. 

Inside proxySet function, I override `canProxy` so it always return `true`.
```js
  const target = proxy(set)
  const handler = unstable_getHandler(target)
  // for testing purpose
  handler.canProxy = () => true

  return target as Set<T>
``` 

When I use the `proxySet` directly as below it works
```js
const state = proxySet([1,2,3])

const App = () => {
  const snapshot = useSnapshot(state);

  return (
      <>
        {/* size updated as expected */}
        <div>size: {snapshot.size}</div>
        <button onClick={() => state.add('aValue')}>button</button>
      </>
    )
}
```

If I use the `proxySet` inside another "state" this won't because the `canProxy` only applies to his target.
```js
const state = proxy({
  set: proxySet([1,2,3])
  otherValue: "foo"
})

const App = () => {
  const snapshot = useSnapshot(state);

  return (
      <>
        {/* size not updated*/}
        <div>size: {snapshot.set.size}</div>
        <button onClick={() => state.set.add('aValue')}>button</button>
      </>
    )
}
```

One workaround would be to use `useSnapshot` on the set itself
```js
const snapshot = useSnapshot(state.set)
```

I'll keep playing with the API maybe I'm missing something

dai-shi:
I don't think you misunderstand. I think I overlooked something...

> I have the feeling that the `!(x as any)[Symbol.iterator])` is a bit too restrictive as it will prevent objects that implement it to be proxied?

Yeah, now it's technically configurable, I can think about changing the default.

dai-shi:
What I overlooked was that, if we already have a proxy, we should skip `canProxy`. Added a new commit in #321.

fkhadra:
Just tried with the new commit this seems to solve the issue! Thanks. I'll focus on the testing part now.

dai-shi:
Is it correct we don't even need to use `unstable_getHandler` in this PR?

fkhadra:
> Is it correct we don't even need to use `unstable_getHandler` in this PR?

Indeed. Which is even better 👌

src/utils/proxySet.ts
dai-shi: https://github.com/pmndrs/valtio/pull/318#discussion_r782055337

> ```js
> const obj = {}
> const state = proxySet([obj])
> state.has(obj) // ---> expecting true
> ```

My comment was incorrect. `state.has(obj)` should be `false`. Set.has should behave like Array.includes.

PR #324, opened 2022-01-13
fix(vanilla): improve DeepResolveType type

johnnynomad:
I think it would be a nice addition to declare snapshots as immutable on Typescript level. This makes it easier for users to notice that they are only supposed to modify the store, not the snapshot.

I also took the liberty to rename the type to `Snapshot`, it's probably better than `DeepResolveTypeReadonly`.
Awesome library by the way!

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -241,21 +241,21 @@ export const subscribe = (
   }
 }
 
-export type Snapshot<T> = T extends (...args: any[]) => any
+export type DeepResolveType<T> = T extends (...args: any[]) => any
   ? T
   : T extends AsRef
-  ? Readonly<T>
+  ? T
   : T extends Promise<infer V>
-  ? Readonly<V>
+  ? DeepResolveType<V>
   : T extends object
   ? {
-      readonly [K in keyof T]: Snapshot<T[K]>
+      readonly [K in keyof T]: DeepResolveType<T[K]>
     }
-  : T extends any
-  ? any
   : Readonly<T>
 
-export const snapshot = <T extends object>(proxyObject: T): Snapshot<T> => {
+export const snapshot = <T extends object>(
+  proxyObject: T
+): DeepResolveType<T> => {
   if (
     typeof process === 'object' &&
     process.env.NODE_ENV !== 'production' &&
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -241,21 +241,21 @@ export const subscribe = (
   }
 }
 
-export type DeepResolveType<T> = T extends (...args: any[]) => any
+export type Snapshot<T> = T extends (...args: any[]) => any
   ? T
   : T extends AsRef
-  ? T
+  ? Readonly<T>
   : T extends Promise<infer V>
-  ? V
+  ? Readonly<V>
   : T extends object
   ? {
-      [K in keyof T]: DeepResolveType<T[K]>
+      readonly [K in keyof T]: Snapshot<T[K]>
     }
-  : T
+  : T extends any
+  ? any
+  : Readonly<T>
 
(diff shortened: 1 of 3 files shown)


src/vanilla.ts
```diff
@@ -241,21 +241,21 @@ export const subscribe = (
   }
 }
 
-export type DeepResolveType<T> = T extends (...args: any[]) => any
+export type Snapshot<T> = T extends (...args: any[]) => any
```
dai-shi, line 244: Please don't change the type name. We may consider renaming in the future, but not this time. Even thought it's not a documented api, changing it can be type breaking.

```suggestion
export type DeepResolveType<T> = T extends (...args: any[]) => any
```

src/vanilla.ts
```diff
@@ -241,21 +241,21 @@ export const subscribe = (
...
   ? T
   : T extends AsRef
-  ? T
+  ? Readonly<T>
   : T extends Promise<infer V>
-  ? V
+  ? Readonly<V>
```
dai-shi, line 249: Actually, there was a bug previously. This has to be recursive.
```suggestion
  ? DeepResolveType<V>
```

src/vanilla.ts
```diff
@@ -241,21 +241,21 @@ export const subscribe = (
...
   : T extends object
   ? {
-      [K in keyof T]: DeepResolveType<T[K]>
+      readonly [K in keyof T]: Snapshot<T[K]>
     }
-  : T
+  : T extends any
```
dai-shi, line 254: How does this trick work? What was the issue without it?

dai-shi (changes requested):
Nice suggestion. Just curious, was the previous typing somehow causing a trouble?

johnnynomad:
Wow, that was a fast code review!

> Just curious, was the previous typing somehow causing a trouble?

No, not at all. I am a freelance developer and am looking for state management libraries I could potentially recommend in future projects. 
This is easier if other developers that are not familiar with this library and its concepts are automatically prevented from making common mistakes.
I personally played around with this library quite a bit now and did some reading, and know I understand the reason for the creating snapshots and why you can't modify them, but it took some time.

src/vanilla.ts
johnnynomad: Originally a test broke that was using `proxy<any>` which lead to Typescript complaining about `Readonly<any>` not being valid or something like that. But I actually can't reproduce this anymore. Maybe my IDE or the TS server had gotten stuck. Anyway I'm removing the code again.

src/vanilla.ts
```diff
@@ -246,12 +246,12 @@ export type DeepResolveType<T> = T extends (...args: any[]) => any
...
   : T extends Promise<infer V>
-  ? V
+  ? DeepResolveType<V>
   : T extends object
   ? {
-      [K in keyof T]: DeepResolveType<T[K]>
+      readonly [K in keyof T]: DeepResolveType<T[K]>
     }
-  : T
+  : Readonly<T>
```
dai-shi, line 254:
```suggestion
  : T
```

Looks like we don't need this.
Check this out: https://tsplay.dev/NlxQXN
Or, please let us know the case it might break.

johnnynomad: Oh yeah that's true, I actually hadn't studied the `Readonly<>` type in detail.
While checking this I noticed that the `Map`, `Set`and `Date` types aren't actually immutable with the current implementation. Do you think we should cover those cases as well, or would that be overkill?

dai-shi: They are indeed not immutable. So, it's correct. Actually, it's not deterministic. With #321, we could technically change the behavior at runtime. So, we can't provide 100% static type anyway.

PR #325, opened 2022-01-13
Feat/add proxy set

fkhadra:
I created a new PR as I rebased from fix(core): expose canProxy in handler #321. Below is the output of the tests that are generated. I can of course add more if needed.
Screenshot 2022-01-13 at 15 20 48
 
I was thinking to add an example as well inside the example folder `todo-with-proxySet` or whatever the name. Let me know if you think it's worth it


dai-shi:
> I was thinking to add an example as well inside the example folder `todo-with-proxySet` or whatever the name. Let me know if you think it's worth it

That's a very nice idea!

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
+import { proxy } from 'valtio'
+
+// properies that we don't want to expose to the end-user
+interface InternalProxySet<T> extends Set<T> {
```
dai-shi, line 4:
```suggestion
type InternalProxySet<T> = Set<T> & {
```
Just preferring type alias when we don't need type overloading.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+  data: T[]
+  toJSON: object
+  hasProxy(value: T): boolean
+}
+
+export function proxySet<T>(
```
dai-shi, line 10: In this case, it's controversial, but currently we don't use function declaration in valtio (in jotai, we prefer function declaration, so we might change it eventually.)
```suggestion
export const proxySet = <T>(
```

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+  initialValues: Iterable<T> | null | undefined = []
```
dai-shi, line 11: can this work?
```suggestion
  initialValues: Iterable<T> = []
```

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+    data: Array.from(new Set(initialValues)),
+    has(value) {
+      return this.data.indexOf(value) !== -1
+    },
+    hasProxy(value) {
+      let hasProxy = false
+      if (typeof value === 'object') {
+        hasProxy = this.data.indexOf(proxy(value as any)) !== -1
+      }
+
```
dai-shi, line 23:
```suggestion
```
just a preference.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+      return hasProxy
+    },
+    add(value) {
+      if (!this.has(value) && !this.hasProxy(value)) {
+        this.data.push(value)
+      }
+      return this
+    },
+    delete(value) {
+      if (!this.has(value) && !this.hasProxy(value)) {
```
dai-shi, line 33: should we check `hasProxy`? It doesn't seem to work correctly with it.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+        return false
+      }
+      this.data = this.data.filter((v: T) => v !== value)
```
dai-shi, line 36: Okay, we should not reassign data here. This is a subtle difference, but maybe important because how proxies work.
```js
  delete(value) {
    const index = this.data.indexOf(value)
    if (index === -1) {
      return false
    }
    this.data.splice(index, 1)
    return true
  }
```

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+      return true
+    },
+    clear() {
+      this.data = []
```
dai-shi, line 40: same here.
```suggestion
      this.data.splice(0)
```

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+    },
+    get size() {
+      return this.data.length
+    },
+    forEach(cb) {
+      for (let i = 0; i < this.data.length; i++) {
```
dai-shi, line 46: Why don't we use `this.data.forEach()`?

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+    },
+    get [Symbol.toStringTag]() {
+      return 'Set'
+    },
+    toJSON() {
+      return {}
+    },
+    [Symbol.iterator]() {
+      let index = 0
+
```
dai-shi, line 59:
```suggestion
```

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+    values() {
+      return this.data.values()
+    },
+    keys() {
+      // for Set.keys is an alias for Set.values()
+      return this.data.values()
+    },
+    entries() {
+      // array.entries returns [index, value] while Set [value, value]
+      return new Set(this.data).entries()
```
dai-shi, line 76: `values` and `keys` use `this.data`, but `entries` copies data with `new Set(this.data)`. While this is fine, feels a little error-prone. If we use `new Set(this.data)` in `values` and `keys`, the behavior is more consistent between the three?

Not sure which is good, because coping comes with a cost.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+    },
+  })
```
dai-shi, line 78: wow, this is so clean. I like it.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+    },
+    size: {
+      enumerable: false,
+    },
+    hasProxy: {
+      enumerable: false,
+    },
+  })
+
+  Object.seal(set)
```
dai-shi, line 93: Do Object.defineProperties and Object.seal work as expected? Could we have tests for them?

dai-shi (commented):
Once again, this is cool. Look forward to merging.

#321 is merged. Please merge (or rebase from) the main branch.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+  initialValues: Iterable<T> | null | undefined = []
+): Set<T> {
+  const set: InternalProxySet<T> = proxy({
+    data: Array.from(new Set(initialValues)),
+    has(value) {
+      return this.data.indexOf(value) !== -1
+    },
+    hasProxy(value) {
+      let hasProxy = false
+      if (typeof value === 'object') {
```
dai-shi, line 20:
```suggestion
      if (typeof value === 'object' && value !== null) {
```
Unfortunate JavaScript.

src/utils/proxySet.ts
fkhadra: I believe this would prevent `proxySet(null)`, `new Set(null)` is valid. I used the same signature as `Set`.
Screenshot 2022-01-14 at 07 37 33

src/utils/proxySet.ts
fkhadra: I think you're right, I'll double check

src/utils/proxySet.ts
fkhadra: No particular reason, I can switch to `forEach`

src/utils/proxySet.ts
fkhadra: I wanted to avoid the cost of creating a new `Set`. That's why I used `this.data` with `values` and `keys` but I can use the set approach if you prefer. I'm ok with both

src/utils/proxySet.ts
fkhadra: Yes it does when your object is wrapped in a proxy first. I'll add some tests for this

src/utils/proxySet.ts
fkhadra: nice catch

src/utils/proxySet.ts
fkhadra: For my personal understanding, why we should not reassign data? This would create a new proxy under the hood, right? Is it because of that?

dai-shi: Yes, reassignment will create a new proxy under the hood. With splice, we keep the same proxy without creating a new one.
image
https://codesandbox.io/s/valtio-puzzle-forked-5hner

src/utils/proxySet.ts
dai-shi: Oh, you are right. `null` has to be clearly defined. `undefined` can be omitted, because of the default parameter.

https://tsplay.dev/Ndd9YN
image

src/utils/proxySet.ts
dai-shi: That's fair. Let's leave them as they are.

src/utils/proxySet.ts
```diff
@@ -0,0 +1,96 @@
...
+    [Symbol.iterator]() {
+      let index = 0
+
+      return {
+        next: () =>
+          index < this.data.length
+            ? { value: this.data[index++], done: false }
+            : { done: true },
+      } as IterableIterator<T>
+    },
```
dai-shi, line 66: I wonder if this can be simplified:
```js
    [Symbol.iterator]() {
      return this.data[Symbol.iterator]
    },
```
Or:
```js
    [Symbol.iterator]() {
      return this.data.values()
    },
```

fkhadra: Wow how I missed that 😆really like it

src/utils/proxySet.ts
```diff
@@ -0,0 +1,89 @@
...
+  hasProxy(value: T): boolean
+}
+
+export const proxySet = <T>(initialValues: Iterable<T> | null = []): Set<T> => {
+  const set: InternalProxySet<T> = proxy({
+    data: Array.from(new Set(initialValues)),
+    has(value) {
+      return this.data.indexOf(value) !== -1
+    },
+    hasProxy(value) {
```
dai-shi, line 16: I wouldn't define this method, but just do it in `add`, which might make the bundle size a little smaller.

And if it's an object we don't need to check `has(value)` which is always false.
Hmmm, let me try:
```js
  add(value) {
    if (typeof value === 'object' && value !== null) {
      value = proxy(value as object)
    }
    if (this.data.indexOf(value) !== -1) {
      this.data.push(value)
    }
    return this
  }
```
Does this work??

dai-shi (commented):
Getting better and better.

src/utils/proxySet.ts
fkhadra: Oh yes, missed this one. Let me try

fkhadra: So i've tried, TS is not super happy. I have to do the following assertion
```ts
if (typeof value === 'object' && value !== null) {
        value = proxy(value as T & object)
      }
      if (this.data.indexOf(value) !== -1) {
        this.data.push(value)
      }
      return this
```

dai-shi: that looks fine!

fkhadra: My bad xD 

```js
  if (this.data.indexOf(value) === -1) { // instead of !== -1
        this.data.push(value)
      }
```

fkhadra: > And if it's an object we don't need to check has(value) which is always false.

This is not 100% true because `Map` and `Set` are objects but they cannot be proxied. So the code for add should be

```tsx
add(value) {
      let hasProxy = false
      if (typeof value === 'object' && value !== null) {
        hasProxy = this.data.indexOf(proxy(value as any)) !== -1
      }
      // this.data.indexOf(value) === -1 or !this.has(value)
      if (this.data.indexOf(value) === -1 && !hasProxy) {
        this.data.push(value)
      }
      return this
    }
```

dai-shi: > This is not 100% true because Map and Set are objects

Oh, I missed it. You are absolutely right.

> // add check like instanceof Map && instanceof Set etc...

This doesn't work as `canProxy` **is** customizable. I like the first one in your previous comment.

dai-shi (approved):
LGTM
This is very good. It doesn't look so magical. (Not easy for everyone though, like how `this` works for snapshots.)
Thanks for your work!

fkhadra:
Amazing, do you want me to add an example before we merge? I'm trying to figure out a good use case(a simple one) for using Set. I thought first about the todo but most of the operations relies on the todo id which makes sense

dai-shi:
You can work on examples in a separate PR. Can you add a short section in readme?

fkhadra:
> You can work on examples in a separate PR. Can you add a short section in readme?

Sure, no problem I'll do this later today when I'm back. I've also pushed another commit to remove the `writable` from data as we don't reassign it.

fkhadra:
Added a tiny readme section let me know if it's enough

dai-shi (approved):
sure, let's merge this and ship it.

PR #330, opened 2022-01-19
feat(utils): add proxyMap 

fkhadra:
WIP on adding the `proxyMap` util. For the implementation I thought about 2 solutions.
Using arrays vs using a Map and an Object.

I went for the latter. This allows us to lookup value with an O(1) (O(2) in that case as we have 2 lookups to do) complexity no matter the size of our state.
With arrays, we would have to iterate to find values and the complexity would increase with the size of our state.

Edit: went with array as Map was breaking snapshot behavior


src/utils/proxyMap.ts
```diff
@@ -0,0 +1,114 @@
...
+  data: {
+    [key: string]: V
+  }
+  toJSON: object
+  makeMap(): Map<K, V>
+}
+
+export const proxyMap = <K, V>(entries?: [K, V][] | null): Map<K, V> => {
+  // store user key and dataRef used to access the value provided by the user
+  const keyStore = new Map<K, number>()
```
dai-shi, line 13: I think having something outside `map` will break the behavior with `snapshot`.
The magic how proxySet/Map work is we translate Set/Map to plain objects(arrays).

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,114 @@
...
+      return 'Map'
+    },
+    [Symbol.iterator]() {
+      return this.makeMap()[Symbol.iterator]()
+    },
+    makeMap() {
+      const map = new Map<K, V>()
+      keyStore.forEach((v, k) => {
+        map.set(k, this.data[v] as V)
+      })
```
dai-shi, line 88: Let's simply have this in `entries()` and `[Symbol.iterator]()` to call `entries()`. Would there be any drawback? I'd like to avoid exposing new properties.

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,114 @@
...
+
+type InternalProxyMap<K, V> = Map<K, V> & {
+  data: {
+    [key: string]: V
+  }
+  toJSON: object
+  makeMap(): Map<K, V>
+}
+
+export const proxyMap = <K, V>(entries?: [K, V][] | null): Map<K, V> => {
```
dai-shi, line 11:
```suggestion
export const proxyMap = <K, V>(entries?: Iterable<readonly [K, V]> | null): Map<K, V> => {
```

We want to support `proxyMap(new Map(...))` too.

dai-shi (commented):
I agree O(1) lookup is desirable, but at least this doesn't seem to work.

```js
const p = proxyMap([[1,'a']])
const s = snapshot(p)
p.delete(1)
s.has(1) // should be true
```

fkhadra:
> I agree O(1) lookup would be desirable, but at least this doesn't seem to work.
> 
> ```js
> const p = proxyMap([[1,'a']])
> const s = snapshot(p)
> p.delete(1)
> s.has(1) // should be true
> ```

With the current implementation, only values are proxies. `keys` remain the way they are. Do we want to proxy keys as well? If yes what would be the benefit?

dai-shi:
> what would be the benefit?

It's not about the benefit. It should work correctly.
Basically, `proxyMap(new Map())` should behave like `proxy({})`, and it also supports any keys.

fkhadra:
> > what would be the benefit?
> 
> It's not about the benefit. It should work correctly. Basically, `proxyMap(new Map())` should behave like `proxy({})`, and it also supports any keys.

Idd, misunderstanding from my side ^^

fkhadra:
I'll add more tests, notably around the snapshotting behavior to ensure that I'm not breaking it.

Regarding our discussion around `ref`, I was wondering if it's a good idea to add something about it in the example block, doing so it would be visible in the IDE. WDYT?

Screenshot 2022-01-19 at 21 49 15

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,113 @@
+import { proxy } from 'valtio'
+
+type KeyValRecord<K, V> = {
+  key: K
+  value: V
+}
```
dai-shi, line 6: I was thinking of this:
```js
type KeyValueRecord<K, V> = [key: K, value: V]
```
but either is okay.

actually, this might generate smallest bundle:
```js
type KeyValRecord<K, V> = {
  k: K
  v: V
}
```

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,113 @@
...
+ * const state = proxyMap([["key", "value"]])
+ *
+ * //can be used inside a proxy as well
+ * const state = proxy({
+ *   count: 1,
+ *   map: proxyMap()
+ * })
+ */
+export const proxyMap = <K, V>(
+  entries: Iterable<readonly [K, V]> | null = []
```
dai-shi, line 30: can be this?
```suggestion
  entries?: Iterable<readonly [K, V]> | null = []
```

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,113 @@
...
+): Map<K, V> => {
+  const map = proxy<InternalProxyMap<K, V>>({
+    data:
+      entries === null
```
dai-shi, line 34: never becomes true.

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,113 @@
...
+        ? []
+        : Array.from(entries).map((v) => ({ key: v[0], value: v[1] })),
+    has(key) {
+      return this.data.findIndex((p) => p.key === key) !== -1
```
dai-shi, line 38: does this work?
```suggestion
      return this.data.some((p) => p.key === key)
```

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,113 @@
...
+    },
+    set(key, value) {
+      const idx = this.data.findIndex((p) => p.key === key)
+      const payload = {
+        key,
+        value,
+      }
+      if (idx !== -1) {
+        this.data[idx] = payload
```
dai-shi, line 47: For several reasons (one thing I notice is the case if the value is the same), we should not replace the existing object.
```suggestion
        this.data[idx].value = value
```
In this case, it's probably simpler to use `.find` instead of `.findIndex` in L41.

dai-shi (commented):
overall looks good.

dai-shi:
> Regarding our discussion around `ref`, I was wondering if it's a good idea to add something about it in the example block, doing so it would be visible in the IDE. WDYT?

I think it's a good idea. I'm not totally sure about various use cases though. (that's why I avoid changing the library code, so the example block idea works for me.)

src/utils/proxyMap.ts
fkhadra: Given it has a default value `[]` we don't have to put the question mark
Screenshot 2022-01-20 at 08 29 40

src/utils/proxyMap.ts
fkhadra: It does when we do `proxyMap(null)`. The default parameter `[]` is used only if the provided value is `undefined` or  if no value is passed to `proxyMap`

src/utils/proxyMap.ts
fkhadra: Nice one, I like it 👍🏽

src/utils/proxyMap.ts
dai-shi: My bad!

src/utils/proxyMap.ts
dai-shi: My bad again!


---
This is just a nit, but since I got confused, here's my preference (not a strong suggestion):

```js
export const proxyMap = <K, V>(
  entries?: Iterable<readonly [K, V]> | null
): Map<K, V> => {
  const map = proxy<InternalProxyMap<K, V>>({
    data: Array.from(entries || []).map((v) => ({ key: v[0], value: v[1] })),
```

And for `proxySet`:
```js
export const proxySet = <T>(initialValues?: Iterable<T> | null): Set<T> => {
  const set: InternalProxySet<T> = proxy({
    data: Array.from(new Set(initialValues)),
```

fkhadra: actually, I prefer your suggestion. It's way cleaner 👌. For `proxySet` do you want me to open a separate PR?

dai-shi: Cool. I often forget about the 2nd arg too.

If the change to `proxySet` is just that one, I think we can do it in this PR.

src/utils/proxyMap.ts
dai-shi: You decided to keep it as is, right? Just confirming as this is marked as resolved. Let me unmark it.

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,121 @@
...
+    },
+    values() {
+      return this.data.map(({ value }) => value).values()
+    },
+    entries() {
+      const map = new Map<K, V>()
+      this.data.forEach(({ key, value }) => {
+        map.set(key, value)
+      })
+      return map.entries()
```
dai-shi, line 97: If we were to have `type KeyValueRecord<K, V> = [key: K, value: V]`, this could become:
```ts
    return new Map<K, V>(this.data).entries()
```

Just an idea, not a suggestion.

dai-shi (commented):
The code looks good to me.

src/utils/proxyMap.ts
fkhadra: I like your idea. Especially the fact that we can create the map in one step. I'll switch the implementation if you are ok with it

fkhadra: Using an array also simplify the initialization of data
```ts
 data: Array.from(entries || [], (v) => ({ key: v[0], value: v[1] })),


 data: Array.from(entries || []) as KeyValRecord<K, V>[] // the assertion is needed as entries is readonly
```

dai-shi: I guess you need to copy the entry items? hm, but `proxy()` will copy it anyway. so, yeah, type assertion is actually fine.

dai-shi (commented):
Looks clean!

dai-shi (commented):
Just [one comment](https://github.com/pmndrs/valtio/pull/330#pullrequestreview-860149724). Other than that, I think it's ready to merge.
What do you think? You can work on examples in a separate PR.

src/utils/proxyMap.ts
```diff
@@ -0,0 +1,111 @@
...
+ * state.get(key) //value
+ *
+ * const key = {}
+ * state.set(key, "value")
+ * state.get(key) //undefined
+ */
+export const proxyMap = <K, V>(
+  entries?: Iterable<readonly [K, V]> | null
+): Map<K, V> => {
+  const map = proxy<InternalProxyMap<K, V>>({
```
dai-shi, line 41: This typing is different from `proxySet`.
```ts
  const set: InternalProxySet<T> = proxy({
```
Should they be consistent?

fkhadra: Indeed. Good remark. I've done the change

fkhadra:
As soon as it's merged I'll open a PR with the example then

dai-shi (approved):
This is great. Thanks for your work! Let's merge it.

PR #335, opened 2022-01-24
fix(utils): make derive glitch free

### LINKED ISSUE #296: This library doesn't seem to be glitch free?
First of all, thanks for sharing this very nice library!

I was doing some experiments, and I noticed that unlike most "push/pull functional reactive programming" libraries, this library is not glitch free, e.g. the callbacks of derived computations can be called with out-of-sync values, producing intermediate invalid results, and can be invoked multiple times per "leaf node" change, which can be inefficient in large "dataflow graphs".

Is this by design? To workaround this, one can apply ancient synchronous dataflow techniques, like giving each node in the dataflow a rank and using a priority queue to invoke the callbacks in the correct order. A bit like a parallel CPU stack.

Code example at https://stackblitz.com/edit/react-sqyzte

```
import React from "react";
import "./style.css";

import { derive } from 'valtio/utils'

const state = proxy({ value: 0 });

// derived1 = state
const derived1 = derive({
  value: (get) => get(state).value,
})

// derived1 = derived2
const derived2 = derive({
  value: (get) => get(derived1).value,
})

// derived3 = state + derived1 - derived2 = state + derived1 - derived1 = state + 0 = state
const derived3 = derive({
  value: (get) => {
    // notice in the console that this callback is invoked TWICE per state update.
    const v0 = get(state).value;
    const v1 = get(derived1).value;
    const v2 = get(derived2).value; // notice in the console that v2 is out-of-sync in the first callback invocation
    const r = v0 + (v1 - v2);
    // notice the glitches in the console
    // when state is changed, this callback is invoked TWICE with out-of-sync values.
    // glitches make debugging really hard, and can cause invalid intermediate values like NaN, div by zero, etc...
    console.log(`v0=${v0} v1=${v1} v2=${v2} => r=${r}`);
    return r;
  }
})


export default function App() {
  const snap = useSnapshot(derived3)
  return (
    <div>
      <code>{snap.value}</code>
      <br/>
      <button onClick={() => ++state.value}>+1</button>
    </div>
  );
}
```

dai-shi:
@mweststrate hinted that a solution to make glitch free is simply to use a counter. https://twitter.com/mweststrate/status/1484928361084379136

So I tried. closes #296 

I had to make the subscription sync and then async in derive. There might be some edge cases this won't work well, like mixing sync and async, but the basic patterns should be covered. Let's see how it goes with releasing this.

### DIFF
--- a/tests/derive.test.tsx
+++ b/tests/derive.test.tsx
@@ -300,7 +300,7 @@ it('basic underive', async () => {
   expect(callback).toBeCalledTimes(1)
 })
 
-describe('glich free', () => {
+describe('glitch free', () => {
   it('basic (#296)', async () => {
     const state = proxy({ value: 0 })
     const derived1 = derive({
--- a/tests/derive.test.tsx
+++ b/tests/derive.test.tsx
@@ -384,4 +384,51 @@ describe('glich free', () => {
     await findByText('value: 1')
     expect(computeValue).toBeCalledTimes(2)
   })
+
+  it('double chain', async () => {
+    const state = proxy({ value: 0 })
+    const derived1 = derive({
+      value: (get) => get(state).value,
+    })
+    const derived2 = derive({
+      value: (get) => get(derived1).value,
+    })
+    const derived3 = derive({
+      value: (get) => get(derived2).value,
+    })
+    const computeValue = jest.fn((get) => {
+      const v0 = get(state).value
+      const v1 = get(derived1).value
+      const v2 = get(derived2).value
+      const v3 = get(derived3).value
+      return v0 + (v1 - v2) + v3 * 0
+    })
+    const derived4 = derive({
+      value: (get) => computeValue(get),
+    })
+
+    const App = () => {
+      const snap = useSnapshot(derived4)
+      return (
+        <div>
+          value: {snap.value}
+          <button onClick={() => ++state.value}>button</button>
+        </div>
+      )
+    }
+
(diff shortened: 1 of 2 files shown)


tests/derive.test.tsx
```diff
@@ -211,7 +211,7 @@ it('nested emulation with derive', async () => {
     {
       doubled: (get) => computeDouble(get(state.math).count),
     },
-    { proxy: state.math }
+    { proxy: state.math, sync: true }
```
dai-shi, line 214: This is technically a behavioral change, but hope it's not too critical.

Noitidart: So if it's async it can't be glitch free?

dai-shi: No, but I didn't confirm it in the nested scenario. Without `sync` in this case, it simply double call the subscriber. It's a side effect of supporting glitch free.

Noitidart:
Just tested it, wow it's so awesome! Even when `false` its glitch free! I upgraded the package to 1.2.11 here and tried - https://codesandbox.io/s/dervie-glitch-0ji5r

dai-shi:
Thanks for confirming. That's great.
The reason I did #335 in the first place (and I didn't notice the bug because the tests are bad. appreciate for your fix!) is, as you might see, #341 implementation is very complicated. I liked the original very simple derive impl (even glitched), but let's try to improve the new impl.

Noitidart:
Thank you brother!

PR #339, opened 2022-01-24
Add Valtio Website Structure

roguesherlock:
- Adds a basic valtio website structure
- Adds a basic structure for docs
- docs are written in mdx and bundled via `mdx-bundler`
- website is built on next js/typescript stack

Find the live preview here,
https://valtio-xi.vercel.app/


website/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "next/core-web-vitals"
```
dai-shi, line 2: Try to fix CI error:
```suggestion
  "root": true,
  "extends": "next/core-web-vitals"
```

website/.eslintrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "root": true,
+  "extends": "next/core-web-vitals"
```
dai-shi, line 3: Hmm, let me try removing it for now.
```suggestion
  "root": true
```

dai-shi: @roguesherlock
This is temporary workaround, let's look for a better solution. We should run eslint on `website`.

dai-shi:
This is a good start! Let's keep going. Merging this.

website/.eslintrc.json
dai-shi: Ah, it wasn't okay. I fixed in #343.

PR #341, opened 2022-01-25
fix(utils): re-implement derive to make glitch free (#335 was buggy)

Noitidart:
Related: https://github.com/pmndrs/valtio/issues/296#issuecomment-1020736885

- [x] add failing test @Noitidart 
- [x] re-implement @dai-shi 
- [x] refactor @dai-shi

### DIFF
--- a/tests/derive.test.tsx
+++ b/tests/derive.test.tsx
@@ -1,5 +1,4 @@
-import { StrictMode, Suspense } from 'react'
-
+import { StrictMode, Suspense, useEffect, useRef } from 'react'
 import { fireEvent, render } from '@testing-library/react'
 import { proxy, snapshot, subscribe, useSnapshot } from 'valtio'
 import { derive, underive } from 'valtio/utils'
@@ -304,36 +303,25 @@ it('basic underive', async () => {
 describe('glitch free', () => {
   it('basic (#296)', async () => {
     const state = proxy({ value: 0 })
-    const derived1 = derive(
-      { value: (get) => get(state).value },
-      { sync: false }
-    )
-    const derived2 = derive(
-      { value: (get) => get(derived1).value },
-      { sync: false }
-    )
+    const derived1 = derive({ value: (get) => get(state).value })
+    const derived2 = derive({ value: (get) => get(derived1).value })
     const computeValue = jest.fn((get) => {
       const v0 = get(state).value
       const v1 = get(derived1).value
       const v2 = get(derived2).value
       return `v0: ${v0}, v1: ${v1}, v2: ${v2}`
     })
-    const derived3 = derive(
-      {
-        value: computeValue,
-      },
-      { sync: false }
-    )
+    const derived3 = derive({ value: computeValue })
 
-    let renderCount = 0
     const App = () => {
       const snap = useSnapshot(derived3)
-
-      renderCount++
-
+      const commitsRef = useRef(1)
+      useEffect(() => {
+        commitsRef.current += 1
+      })
       return (
         <div>
-          value: {snap.value}
+          value: {snap.value} (commits: {commitsRef.current})
           <button onClick={() => ++state.value}>button</button>
         </div>
       )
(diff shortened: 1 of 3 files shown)


tests/derive.test.tsx
```diff
@@ -303,24 +304,33 @@ it('basic underive', async () => {
...
+        value: computeValue,
+      },
+      { sync: false }
+    )
 
+    let renderCount = 0
     const App = () => {
       const snap = useSnapshot(derived3)
+
+      renderCount++
```
dai-shi, line 332: We should count commits not renders. Let me tweak the test.

tests/derive.test.tsx
```diff
@@ -329,18 +339,17 @@ describe('glitch free', () => {
...
+    await findByText('value: v0: 0, v1: 0, v2: 0')
     expect(computeValue).toBeCalledTimes(1)
+    expect(renderCount).toBe(1)
 
     fireEvent.click(getByText('button'))
-    await findByText('value: 1')
-    expect(computeValue).toBeCalledTimes(2)
+
+    await findByText('value: v0: 1, v1: 1, v2: 1')
+    expect(computeValue).toBeCalledTimes(3)
```
dai-shi, line 351: This should be `2`, right?

dai-shi:
Okay, I kind of new my impl in #335 was truly correct. I had another impl, and based on it, a commit is added. It's complicated, so let me try some refactoring.

Meanwhile, @Noitidart could you try if it solves the issue?

dai-shi:
Let's release this soon, because #342 is a bigger issue. v1.2.10 is full of bugs. 😓

dai-shi (approved):
Thanks for your investigation. I hope this is better than #335, but not sure if there are some bugs remaining. Please report issues if anything goes wrong.

tests/derive.test.tsx
Noitidart: Oh very interesting I didn't know useEffect was the way to count commits! I learned a lot here thanks!

Noitidart:
> Okay, I kind of new my impl in #335 was truly correct. I had another impl, and based on it, a commit is added. It's complicated, so let me try some refactoring.
> 
> Meanwhile, @Noitidart could you try if it solves the issue?

Trying now! Sorry for delay I'm on west coast us so just woke up haha.

Noitidart:
With v1.2.9 this used to happen:

```
​compute1 
compute2 
compute3: 
000 
render value: 
v0: 0, v1: 0, v2: 0 
INCREMENT
compute1 
compute3: 
110 
compute2 
compute3: 
111 
render value: 
v0: 1, v1: 1, v2: 1 
```

Notice how compute1 then compute3 happens, then again compute2 happens then compute3 happens.

Now in v1.2.11 its perfect:

```
compute1 
compute2 
compute3: 
000 
render value: 
v0: 0, v1: 0, v2: 0 
INCREMENT
compute1 
compute2 
compute3: 
111 
render value: 
v0: 1, v1: 1, v2: 1 
```

compute1 then compuet2 then compute3 and all variables are lined up! no extra callback is triggered!

dai-shi:
> Trying now! Sorry for delay I'm on west coast us so just woke up haha.

No worries! I knew the timezone difference. I wanted to replace troublesome v1.2.10 as soon as possible.

Noitidart:
Totally gotcha, thanks for giving me chance to take part of the emergency release. If possible it happens again no matter time if I see it I would definitely get on it, thanks for trying to include us man!

dai-shi:
By the way, this implementation might still have a limitation. A circular graph may result in an infinite loop. Would you like to test it??
(It may happen with v1.2.9 too.)

Noitidart:
> By the way, this implementation might still have a limitation. A circular graph may result in an infinite loop. Would you like to test it?? (It may happen with v1.2.9 too.)

Haha thanks! I don't understand the circular graph situation, if you could help me understand I would learn about this circular graph thing.

dai-shi:
The circular graph is something like this: A derived from B derived from C derived from A.

dai-shi:
Found this: https://en.wikipedia.org/wiki/Reactive_programming#Cyclic_dependencies

It's called cyclic dependencies.

Noitidart:
Oh thanks! Sorry I got busy with urgently implementing a Stripe payment system update. I'll have to revisit this please. Thanks very much for trying to get me involved!

PR #344, opened 2022-01-25
chore(docs): remove/rename some apis

dai-shi:
- put `underive` in `derive` doc (not often used, should be easier to read)
- no docs for unstable_ derive api (we are not sure about the usage)
- rename `unstable_getHandler` (it's still very experimental)
- delete `addComputed` (it's deprecated)
- add `proxyMap` and `proxySet` (new ones!)


docs/utils/proxyMap.mdx
```diff
@@ -1,7 +1,9 @@
...
-title: 'addComputed'
+title: 'proxyMap'
 section: 'Utils'
 description: ''
 ---
 
-# addComputed
+# proxyMap
+
+TODO
```
dai-shi, line 9: @fkhadra Feel free to work on `proxyMap` and `proxySet`. Nobody else will be working on them.

PR #354, opened 2022-02-08
fix(utils): multiple redos at once with proxyWithHistory

### LINKED ISSUE #323: Bug with history (again?)
Hello again, I think I found something.

[sandbox link](https://codesandbox.io/s/react-typescript-forked-6leim?file=/src/App.tsx)

I'm trying to implement transition (jump) between different states in my project and found one issue.

If I try to undo or redo a state in a loop it doesn't work as expected.

PS: Loop is just an example for this issue, I know I can get the value directly from the snapshots array.

Expected:
Case 1:
initial value 0
UpdateValue (x3) -> undo by step (x3) -> redo by step (x3) -> 3

Case 2:
initial value 0
UpdateValue (x3) -> undo all -> redo all -> 3

-----------------------------------------------------------------------------

Actual:
Case 1:
initial value 0
UpdateValue (x3) -> undo by step (x3) -> redo by step (x3) -> 3        <- fine

Case 2:
initial value 0
UpdateValue (x3) -> undo all -> redo all -> 0                                          <- bad

dai-shi:
close #323

### DIFF
--- a/src/utils/proxyWithHistory.ts
+++ b/src/utils/proxyWithHistory.ts
@@ -62,7 +62,7 @@ export const proxyWithHistory = <V>(initialValue: V, skipSubscribe = false) => {
     subscribe: () =>
       subscribe(proxyObject, (ops) => {
         if (
-          ops.some(
+          ops.every(
             (op) =>
               op[1][0] === 'value' &&
               (op[0] !== 'set' || op[2] !== proxyObject.history.wip)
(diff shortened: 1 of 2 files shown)


hee10k (changes requested):
It seems this code makes just 1 snapshot.
so you can change code like below
```ts

// fireEvent.click(getByText('inc'))
// fireEvent.click(getByText('inc'))
await waitFor(() => fireEvent.click(getByText("inc")));
await waitFor(() => fireEvent.click(getByText("inc")));
```

dai-shi:
> It seems this code makes just 1 snapshot.

wow! thanks so much!!

dai-shi:
eaa96e4 passes the tests. I'm not 100% sure if there are any other cases that this can't handle, but let's go with this for now.

PR #357, opened 2022-02-11
fix(docs): add note in readme about loose typing of useSnapshot

### LINKED ISSUE #327: `readonly` considered as a breaking change?
A `readonly` has been introduced in the recent minor 1.2.9 release

https://github.com/pmndrs/valtio/commit/a983b8b8bfa4bcedeb90774c74bc30270f27931e#diff-b68a2cbc8606fe37f9c2c72867a87ff50a11c2b38a6d9de45a94281f02bf0a4eR252

Is this considered to be a breaking change? The change makes sense to me and it looks useful, but it in my case, I haven't explicitly marked props as `readonly`. My build breaks in ts strict mode and in order to fix it, I have to mark all props as `readonly` (I haven thought much about it before, but to me props are always immutable) where I pass data from `useSnapshot`.

dai-shi:
close #327 

---
 With #359, we can do [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).

https://codesandbox.io/s/react-typescript-forked-qphpi?file=/src/App.tsx:46-125
```ts
declare module "valtio" {
  function useSnapshot<T extends object>(p: T): T;
}
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -48,6 +48,23 @@ function Counter() {
 ```
 
 <details>
+<summary>Note for TypeScript users: Return type of useSnapshot can be too strict.</summary>
+
+The `snap` variable returned by `useSnapshot` is a (deeply) read-only object.
+Its type has `readonly` attribute, which may be too strict for some use cases.
+
+To mitigate typing difficulties, you might want to consider creating a utility function like this:
+```ts
+const useSnapshotLooselyTyped = <T extends object>(
+  proxyObject: T,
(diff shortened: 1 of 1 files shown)


dai-shi:
Hmm, I wonder if [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) works in this case.

PR #358, opened 2022-02-11
fix(utils): do not depend on react in build

dai-shi:
fix #356

### DIFF
--- a/src/utils/proxyMap.ts
+++ b/src/utils/proxyMap.ts
@@ -1,4 +1,4 @@
-import { proxy } from 'valtio'
+import { proxy } from '../vanilla'
 
 type KeyValRecord<K, V> = [key: K, value: V]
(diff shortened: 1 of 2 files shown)


src/utils/proxyMap.ts
```diff
@@ -1,4 +1,4 @@
-import { proxy } from 'valtio'
+import { proxy } from '../vanilla'
```
Aslemammad, line 1: I faced this issue with jotai/devtools yesterday, you may consider updating it also!

Aslemammad: *in vite

dai-shi: what's the issue in jotai? jotai is tightly coupled with react for now. (I want to improve it eventually.)

PR #368, opened 2022-02-14
chore(examples): add readme for proxyWithHistory example

fkhadra:
This PR add the readme, so it's consistent with the other examples


dai-shi (approved):
Thanks for the chore!!
cc @RikuVan

PR #370, opened 2022-02-14
chore(doc): add proxySet documentation

fkhadra:


dai-shi (approved):
Thanks!
We want a codesandbox demo eventually.

PR #376, opened 2022-02-20
chore(docs): proxy

RikuVan:
First go at this. Improvement suggestions 🙏

### DIFF
--- a/docs/basic/proxy.mdx
+++ b/docs/basic/proxy.mdx
@@ -6,6 +6,8 @@ description: 'Create a proxy object.'
 
 # proxy
 
+The proxy tracks changes to the original object and all nested objects, notifying listeners when a new version of the object is created.
+
 ```jsx
 import { proxy, useSnapshot } from 'valtio'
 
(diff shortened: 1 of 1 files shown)


docs/basic/proxy.mdx
```diff
@@ -6,6 +6,8 @@ description: 'Create a proxy object.'
 
 # proxy
 
+The proxy tracks changes to the original object and all nested objects, notifying listeners when a new version of the object is created.
```
dai-shi, line 9: > created

Precisely speaking, object doesn't have to be created, but just modified. Like `++state.count` doesn't create a new object.

docs/basic/proxy.mdx
```diff
@@ -21,3 +23,82 @@ setInterval(() => {
...
+```
+
+#### Gotchas
+
+You cannot descructure the object from proxy and subscribe to a property.
+
+```jsx
+// don't
+const { user } = proxy({ user: { name: 'Timo' } })
+subscibe(user, () => {...})
```
dai-shi, line 70: This still works. Maybe you want to point out something more specific.

docs/basic/proxy.mdx
```diff
@@ -21,3 +23,82 @@ setInterval(() => {
...
+
+// do
+const state = proxy({ user: { name: 'Nina' } })
+subscibe(state.user, () => {...})
```
dai-shi, line 74: Do you mean `subscribe(state, () => {...})` so that we can detect `state.user = null`-kind-of mutations?

dai-shi (commented):
Thanks for working on this. Added some comments.

docs/basic/proxy.mdx
RikuVan: ok, I thought this didn't work 😮

dai-shi (approved):
Looks good. Thanks for your work!

PR #378, opened 2022-02-21
Fix "Please install/enable Redux devtools" warning in "test" environments like jest

AshConnolly:
Hi there! 👋 

I found the following issue:
Screenshot 2022-02-21 at 13 22 57

This is when I run jest. I'm testing a function that does not interact with Valtio state in any way, but it is in a file that has other functions that do interact with valtio state (so state is imported into this file). for example if I tested the `add()` function from the following file I would get the warning:
```
import { AppState } from '../state/AppState'

export const updateState = () => AppState.name = 'hello'

export const add = (x, y) => x + y
```

I've not looked at the repo before and I'm not a TypeScript pro, but I believe the rollup config sets the `__DEV__` global type. I have updated the rollup config and this is the compiled code after fix has been added, which is solves this issue:
```
  if (!extension) {
    if (process.env.NODE_ENV !== "production" && process.env.NODE_ENV !== "test" && typeof window !== 'undefined') {
      console.warn('[Warning] Please install/enable Redux devtools extension');
    }

    return;
  }
  ```

All tests currently pass fine, but wonder if `devtools.test.tsx` need updating to reflect this change.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -66,7 +66,8 @@ function createESMConfig(input, output) {
       }),
       resolve({ extensions }),
       replace({
-        __DEV__: '(import.meta.env&&import.meta.env.MODE)!=="production"',
+        __DEV__:
+          '(import.meta.env&&import.meta.env.MODE)!=="production"&&(import.meta.env&&import.meta.env.MODE)!=="test"',
         preventAssignment: true,
       }),
       getEsbuild('node12'),
@@ -88,7 +89,8 @@ function createCommonJSConfig(input, output) {
       }),
       resolve({ extensions }),
       replace({
-        __DEV__: 'process.env.NODE_ENV!=="production"',
+        __DEV__:
+          'process.env.NODE_ENV!=="production"&&process.env.NODE_ENV!=="test"',
         preventAssignment: true,
       }),
       babelPlugin(getBabelOptions({ ie: 11 })),
@@ -123,7 +125,8 @@ function createUMDConfig(input, output, env) {
       }),
       resolve({ extensions }),
       replace({
-        __DEV__: env !== 'production' ? 'true' : 'false',
+        __DEV__:
+          env !== 'production' ? (env !== 'test' ? 'true' : 'false') : 'false',
         preventAssignment: true,
       }),
       babelPlugin(getBabelOptions({ ie: 11 })),
@@ -150,7 +153,8 @@ function createSystemConfig(input, output, env) {
       }),
       resolve({ extensions }),
       replace({
-        __DEV__: env !== 'production' ? 'true' : 'false',
+        __DEV__:
+          env !== 'production' ? (env !== 'test' ? 'true' : 'false') : 'false',
         preventAssignment: true,
       }),
       getEsbuild('node12', env),

### CODE REVIEW

**Review by dai-shi — CHANGES_REQUESTED** [review]
Hi, thanks for digging into this.
However, test shouldn't run in production mode, which hide some important debug info.
We need a better fix only to hide the devtools warning in test env.

**Comment by dai-shi** [comment]
I wonder how Redux solves it...

**Comment by dai-shi** [comment]
https://github.com/reduxjs/redux-devtools/tree/main/packages/redux-devtools-extension
They don't seem to have such warning???

**Comment by AshConnolly** [comment]
Very welcome! Excited to contribute to this awesome tool! 😄 

> They don't seem to have such warning???

lol do they not?! 😅 

> However, test shouldn't run in production mode, which hide some important debug info.

I thought my change wasn't doing anything that would affect production? I thought it was just making the __DEV__ env type more specific / accurate - so that it does not include the test env. Is that incorrect?

If changing the rollup config is less than ideal would simply changing the devtools.ts file be a sufficient fix? (line 25):
```
  if (!extension) {
    if (
      __DEV__ &&
      process.env.NODE_ENV !== 'test' &&
      typeof window !== 'undefined'
    ) {
      console.warn('[Warning] Please install/enable Redux devtools extension')
    }
    return
  }
```
Thanks @dai-shi! 😄 👍

dai-shi:
I mean we should run test with `__DEV__=true`.

> If changing the rollup config is less than ideal would simply changing the devtools.ts file be a sufficient fix? (line 25):

This isn't nice either. NODE_ENV is only used in CJS build.

> lol do they not?! 😅

Can you confirm?? If it's true, we can follow and remove it in all zustand/jotai/valtio.

AshConnolly:
> Can you confirm?? If it's true, we can follow and remove it in all zustand/jotai/valtio.

Oh sorry, I thought you were saying that Redux has no such warning. I'm not sure how they do it!

> I mean we should run test with __DEV__=true.

Ah because running tests is still part of a dev task / environment... I understand. 
Not sure what the best solution to this is then! 🤔

dai-shi:
> > Can you confirm?? If it's true, we can follow and remove it in all zustand/jotai/valtio.
> 
> Oh sorry, I thought you were saying that Redux has no such warning. I'm not sure how they do it!

You got it right and what I was asking for you is to confirm if Redux has no such warning. Sorry about the confusion!

dai-shi:
@Aslemammad Would you be able to investigate the Redux code?
@AshConnolly Are you still around?

Aslemammad:
I personally don't have a problem with the log!
But we've got two solutions:
* trying to mock the devtools API as we do for testing: https://github.com/pmndrs/valtio/blob/6d2c3855f087427d072e444077115f5c1a8db970/tests/devtools.test.tsx#L37
* wrap your `devtools` calls with `if (process.env.NODE_ENV === 'development')` so they can only run in dev mode (if you don't need them in testing!


cc @AshConnolly @dai-shi

dai-shi:
@Aslemammad My question was just about the implementation/behavior of the officieal redux. And, it looks like they don't have such a warning. So, let's follow it.

dai-shi:
Closing in favor of #400.

PR #380, opened 2022-02-22
fix(devtools): trace

### LINKED ISSUE #331: [devtools] Trace support
One useful feature of redux devtools is [tracing](https://github.com/reduxjs/redux-devtools/blob/main/extension/docs/Features/Trace.md): it essentially gives you the stack trace of an action / mutation.

https://github.com/reduxjs/redux-devtools/blob/main/extension/docs/Features/Trace.md

Note that the following does _not_ work. I naively tried modifying [`utils/devtools`](https://github.com/pmndrs/valtio/blob/main/src/utils/devtools.ts) and adding

```js
  const devtools = extension.connect({ name, trace: true })
```

but this does not work as it results in a trace in the same file of 

```
      devtools.send(
        {
          type: action,
          updatedAt: new Date().toLocaleString(),
        },
        snapWithoutDevtools
      )
```

Aslemammad:
Resolves #331, completely WIP. not ready for review.

### DIFF
--- a/src/utils/devtools.ts
+++ b/src/utils/devtools.ts
@@ -1,4 +1,4 @@
-import { snapshot, subscribe } from '../vanilla'
+import { snapshot, STACK, subscribe } from '../vanilla'
 
 type Message = { type: string; payload?: any; state?: any }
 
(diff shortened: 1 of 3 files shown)


Aslemammad:
Relies on https://github.com/reduxjs/redux-devtools/issues/1100

dai-shi:
Closing for now.

PR #383, opened 2022-02-25
chore(docs): initial description and examples of watch

RikuVan:


docs/utils/watch.mdx
```diff
@@ -5,3 +5,52 @@ description: 'watch for changes.'
 ---
 
 # watch
+
+This utility behaves similarly to `useEffect` in React. Proxy objects are tracking with a `get` function passed to the callback. Notice the callback will run eagerly as soon as a proxy is tracked, even if not yet mutated.
```
dai-shi, line 9: Never thought about the similarity to useEffect. Not sure how others think.

docs/utils/watch.mdx
```diff
@@ -5,3 +5,52 @@ description: 'watch for changes.'
...
+setTimeout(() => {
+  userState.user.name = "Anonymous"
+}, 200)
+// 200ms -> 'Anonymous's session is expired"
+// 'Cleaning up' logged
+```
+
+#### Gotchas
+
+If you remove the setTimeout in the example above, 'Juuso's session is expired' will become 'Anonymous's session is expired', logged twice now. Valtio will batch updates by default. You may pass `{sync: true}` as a second argument to `watch` to disable batching. However, this is relatively untested and should be used with caution.
```
dai-shi, line 55: "logged twice now instead of three times." could be clearer?

dai-shi (commented):
added comments, but i'm fine to merge it as is.

docs/utils/watch.mdx
RikuVan: Not sure I undersand your comment - it was not logged three times before.

The change is with a timeout you get:
> Juuso's session is expired
> Anonymous's session is expired

But without the timeout you get
> Anonymous's session is expired
> Anonymous's session is expired

dai-shi (approved):
Merging this for now. Feel free to open new PRs if anyone has ideas to improve it.

Issue #349, opened 2022-02-02
infinite loop when two derived properties return non primitive values


lukebarlow:
To start, I want to echo others by saying thanks for a great library. I came across this infinite loop. It may be a dependency tracking issue which is considered outside the scope of valtio, but here it is in any case.

```
describe('two derived properties', () => {
  test('two derived properties both returning non primitive values', async () => {
    const state = proxy({
      a: 1
    })

    derive({
      d1: (get) => {
        console.log('in the d1 getter')
        const a = get(state).a
        return {} // 'return a' here works fine. Bug is only if you return a reference value
      }
    }, { proxy: state })

    derive({
      d2: (get) => {
        console.log('in the d2 getter')
        const a = get(state).a
        return {} // again, 'return a' works here
      }
    }, { proxy: state })
  })
})
```

dai-shi:
Thanks for reporting.
Without looking into the detail, the first question came to mind is, does this happen both in v1.2.11 and v1.2.9?

lukebarlow:
Yes, it happens with both those versions

dai-shi:
Thanks.
Hm, this is probably similar to what I concerned in https://github.com/pmndrs/valtio/pull/341#issuecomment-1021700413.
But, this is in a simpler form.
I think we would like to to fix this eventually, but this specific case can be worked around.

```js
describe.only('two derived properties', () => {
  test('two derived properties both returning non primitive values', async () => {
    const state = proxy({
      a: 1
    })

    derive({
      d1: (get) => {
        console.log('in the d1 getter')
        const a = get(state).a
        return {} // 'return a' here works fine. Bug is only if you return a reference value
      },
      d2: (get) => {
        console.log('in the d2 getter')
        const a = get(state).a
        return {} // again, 'return a' works here
      }
    }, { proxy: state })
  })
})
```

It may not work for your real case, though.

dai-shi:
Hoping @Noitidart will be back soon.

Noitidart:
Heyyy! Thanks so much for waiting for me on this, I'm still working that Stripe implementation right now super close. Thanks so much for hanging in there for me! If anyone else (not dai-shi he does so much) can help brother dai-shi out that would be way cool. @lukebarlow would you be able to help write a test for it?

dai-shi:
@lukebarlow Would you like to try #399?
https://ci.codesandbox.io/status/pmndrs/valtio/pr/399
☝️ Find "Local Install Instructions"

[...]

PR #387, opened 2022-03-07
test for issue #349

lukebarlow:
3 tests, two passing and one failing at the time of writing, which illustrate the issue described in #349


dai-shi:
This was really tough. Finally done in #399. Thanks for the test!

PR #389, opened 2022-03-10
fix(utils): devtools better types

Aslemammad:
> This PR tries to bring the native `window.__REDUX_DEVTOOLS_EXTENSION__` types from redux devtools and then avoids some custom types which may be incorrect in some cases!

Related: https://github.com/pmndrs/zustand/pull/819, https://github.com/pmndrs/jotai/pull/1049

### DIFF
--- a/src/utils/devtools.ts
+++ b/src/utils/devtools.ts
@@ -36,8 +36,7 @@ export function devtools<T extends object>(proxyObject: T, name = '') {
   }
 
   let isTimeTraveling = false
-  const devtools: ReturnType<NonNullable<typeof extension>['connect']> =
-    extension.connect({ name })
+  const devtools = extension.connect({ name })
   const unsub1 = subscribe(proxyObject, (ops) => {
     const action = ops
       .filter(([_, path]) => path[0] !== DEVTOOLS)
--- a/src/utils/devtools.ts
+++ b/src/utils/devtools.ts
@@ -1,6 +1,12 @@
 import { snapshot, subscribe } from '../vanilla'
+import type {} from '@redux-devtools/extension'
 
-type Message = { type: string; payload?: any; state?: any }
+// FIXME https://github.com/reduxjs/redux-devtools/issues/1097
+export type Message = {
+  type: string
+  payload?: any
+  state?: any
+}
 
 const DEVTOOLS = Symbol()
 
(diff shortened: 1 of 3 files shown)


src/utils/devtools.ts
```diff
@@ -30,7 +36,8 @@ export function devtools<T extends object>(proxyObject: T, name?: string) {
   }
 
   let isTimeTraveling = false
-  const devtools = extension.connect({ name })
+  const devtools: ReturnType<NonNullable<typeof extension>['connect']> =
+    extension.connect({ name })
```
dai-shi, line 40: I guess we can remove the annotation.
```suggestion
  const devtools = extension.connect({ name })
```

src/utils/devtools.ts
```diff
@@ -1,6 +1,12 @@
 import { snapshot, subscribe } from '../vanilla'
+import type {} from '@redux-devtools/extension'
 
-type Message = { type: string; payload?: any; state?: any }
+// FIXME https://github.com/reduxjs/redux-devtools/issues/1097
+export type Message = {
```
dai-shi, line 5: does this work?
```suggestion
type Message = {
```

src/utils/devtools.ts
```diff
@@ -15,10 +21,10 @@ const DEVTOOLS = Symbol()
  * const state = proxy({ count: 0, text: 'hello' })
  * const unsub = devtools(state, 'state name')
  */
-export function devtools<T extends object>(proxyObject: T, name?: string) {
-  let extension: any
+export function devtools<T extends object>(proxyObject: T, name = '') {
```
dai-shi, line 24: Does the original one work? No?
```suggestion
export function devtools<T extends object>(proxyObject: T, name?: string) {
```

dai-shi (commented):
Please check some minor comments.

src/utils/devtools.ts
Aslemammad: No it does not! It's not aligned with extension.connect!

dai-shi: Alright!

src/utils/devtools.ts
```diff
@@ -50,12 +56,19 @@ export function devtools<T extends object>(proxyObject: T, name?: string) {
...
   })
-  const unsub2 = devtools.subscribe((message: Message) => {
+  const unsub2 = (
+    devtools as unknown as {
+      // FIXME https://github.com/reduxjs/redux-devtools/issues/1097
+      subscribe: (
+        listener: (message: Message) => void
+      ) => (() => void) | undefined
+    }
+  ).subscribe((message: Message) => {
```
dai-shi, line 71: Can be this:
```suggestion
  ).subscribe((message) => {
```

dai-shi (approved):
Other than that, looks good.

PR #396, opened 2022-03-23
[docs] Improve codesandbox style

LuckyPigeon:
The current `codesandbox` iframe can be improved by font sizing, top & bottom padding, and more.
Here is the before & after

before:
image

after:
image

### DIFF
--- a/docs/advanced/subscribe.mdx
+++ b/docs/advanced/subscribe.mdx
@@ -34,4 +34,4 @@ state.arr.push('world')
 
 ## Codesandbox demo in VanillaJS
 
-https://codesandbox.io/s/valtio-photo-booth-demo-forked-xp8hs?file=/src/main.js
+https://codesandbox.io/s/valtio-photo-booth-demo-forked-xp8hs?file=/src/main.js&codemirror=1&fontsize=14&hidenavigation=1&theme=light&hidedevtools=1
--- a/docs/basic/useSnapshot.mdx
+++ b/docs/basic/useSnapshot.mdx
@@ -79,4 +79,4 @@ return (
 
 ## Codesandbox demo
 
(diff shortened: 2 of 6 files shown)


roguesherlock (commented):
Looks good. I was wondering if we can write a remark plugin to append those query params to codesandbox url in the post? That way users wouldn't have to append styling params to codesandbox urls every time. Does it makes sense ?

roguesherlock:
@LuckyPigeon I think we can get around this buy writing our own remark plugin. That plugin should go before oembed plugin. That way, it'll grab any sandbox url and add our query params. It should looks something like this.

```
const remarkPlugins: U.PluggableList = [
  [
    remarkCodeSandboxURLUpdater, // <----- our own custom remark plugin
    // @ts-expect-error
    remarkEmbedder,
    {
      handleError: handleEmbedderError,
      handleHTML: handleEmbedderHtml,
      transformers: [oembedTransformer],
    },
  ],
];
```

```
const remarkCodeSandboxURLUpdater = (options = {}) => {
   // find links which pass codesandbox domain
   // replace links with our query
}
```

Does it makes sense?
Let me know if you need help with writing the plugin (I'd start by looking at how other remark plugins are implemented, https://github.com/remarkjs/remark/blob/main/doc/plugins.md)

dai-shi (changes requested):
@LuckyPigeon Will you be able to work on the suggested change?

LuckyPigeon:
@dai-shi 
Sure! I'm looking at the implementation of plugins. I think it will take some time. This can merge first.

dai-shi:
Great to hear that. We don't plan to merge this as is. Please take your time.
If you have questions about the implementation, @roguesherlock would be able to help.

roguesherlock:
yes. let me know if you have any questions or need any help :)

dai-shi:
Can close in favor of #437, right?

PR #400, opened 2022-03-27
fix(utils/devtools): enabled option and supress warning

dai-shi:
This adds `enabled` option.
It shows the missing devtools ext warning only if `enabled=true`.
It deprecates bare string `name` option (use `{ name }` instead).

Following the change in https://github.com/pmndrs/zustand/pull/880.


dai-shi:
@AshConnolly (cc @Aslemammad )
Here's my approach to #378. Let me know if it solves your issue.

PR #401, opened 2022-03-28
Update subscribing example in readme

lfades:
I think the dependency of the `useEffect` was supposed to be `state.arr`, however, does it make a difference to have it listed in the dependencies if subscribing to it means watching for its changes anyway? 🤔 

Also added a space with `br` in the `details` section about using devtools, to avoid the image getting in the same block of the text.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -198,7 +198,7 @@ function Foo() {
   const total = useRef(0)
   useEffect(() => subscribe(state.arr, () => {
     total.current = state.arr.reduce((p, c) => p + c)
-  }), [state.obj])
+  }), [state.arr])
   // ...
 ```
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -198,7 +198,7 @@ function Foo() {
   const total = useRef(0)
   useEffect(() => subscribe(state.arr, () => {
     total.current = state.arr.reduce((p, c) => p + c)
-  }), [state.obj])
+  }), [state.arr])
```
dai-shi, line 201: This was probably somewhat intentional, but maybe misleading and incorrect. Let's remove it completely.
```suggestion
  }), [])
```

dai-shi (commented):
Thanks for your suggestion. Added a comment. For `<br>`, I'm not sure if it's ideal but it's okay for now.

dai-shi (approved):
Let's merge.

PR #405, opened 2022-03-30
refactor(vanilla): Use the default parameters

ahaoboy:
This PR refactors vanilla.ts to use default parameter instead of if statement.

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -57,13 +57,10 @@ export function proxy<T extends object>(initialObject: T = {} as T): T {
   }
   let version = globalVersion
   const listeners = new Set<Listener>()
-  const notifyUpdate = (op: Op, nextVersion?: number) => {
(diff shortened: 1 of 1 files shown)


src/vanilla.ts
```diff
@@ -57,13 +57,10 @@ export function proxy<T extends object>(initialObject: T = {} as T): T {
   }
   let version = globalVersion
   const listeners = new Set<Listener>()
-  const notifyUpdate = (op: Op, nextVersion?: number) => {
-    if (!nextVersion) {
-      nextVersion = ++globalVersion
-    }
+  const notifyUpdate = (op: Op, nextVersion = ++globalVersion) => {
```
dai-shi, line 60: Cool. So, the expression will not be evaluated if the argument is present.

dai-shi (approved):
Looks good. Thanks for your contribution.

PR #406, opened 2022-03-30
chore: refine type and remove copy-paste

ahaoboy:


ahaoboy:
I don't know why repeat  Snapshot, I try export and import, and works good. 
1F09C8D9

dai-shi:
It's intentionally not exported to hide it as an implementation detail.
We may reconsider this in the future, but let's keep it for now.

PR #407, opened 2022-03-30
feat: refine computed type [breaking change]

ahaoboy:
I try a new way to solve computed type, use this to access snap
The current implementation is very rough, I will continue to improve it if necessary
The demo is in example/counter-ts

image

image


dai-shi:
Hmmm, but I prefer `snap` to `this`. In general, I'd avoid `this` if possible.
What do think of the idea to create a 3rd-party package for your implementation?

ahaoboy:
> Hmmm, but I prefer `snap` to `this`. In general, I'd avoid `this` if possible. What do think of the idea to create a 3rd-party package for your implementation?

Ok

PR #421, opened 2022-04-10
chore: website landing page with animation duration state example

RikuVan:
experimenting with a landing page similar to Zustand with example

Screenshot 2022-04-10 at 21 03 04
Screenshot 2022-04-10 at 21 02 37


RikuVan:
This was done is a bit of a rush and undoubtably needs polishing

website/pages/index.tsx
```diff
@@ -1,29 +1,312 @@
...
+    state.baseDuration + "s"
+  );
+});
+
+const exampleCode = `
+  const state = proxy({
+    baseDuration: 4
+  });
+  const incDuration = () => {
+    state.baseDuration = ++state.baseDuration;
```
dai-shi, line 28:
```suggestion
    ++state.baseDuration;
```
should be fine?

website/pages/index.tsx
```diff
@@ -1,29 +1,312 @@
...
+    );
+  });
+  
+  <div>
+    <h3>
+      {snapshot.baseDuration}
+    </h3>
+    <button
+      disabled={snapshot.baseDuration <= 3}
+      onClick={() => decDuration()}>
```
dai-shi, line 47:
```suggestion
      onClick={decDuration}>
```
should be fine?

dai-shi (commented):
Thanks for working on this!
It's a bit unfortunate that the animation is reset when you click the inc/dec buttons.

website/pages/index.tsx
RikuVan: yeah of course

website/pages/index.tsx
RikuVan: yep, done

RikuVan:
> Thanks for working on this! It's a bit unfortunate that the animation is reset when you click the inc/dec buttons.

Yeah I agree it is jarring. I suppose this could be done with a library that to smooth this out...

dai-shi (approved):
LGTM

PR #422, opened 2022-04-11
chore: resources pages to website

RikuVan:


docs/resources/community.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+## Recipes wiki
+
+Valtio is unopinionated about best practices.
+The community is working on recipes on wiki pages.
+
+- [How to organize actions](https://github.com/pmndrs/valtio/wiki/How-to-organize-actions)
+- [How to persist states](https://github.com/pmndrs/valtio/wiki/How-to-persist-states)
+- [How to use with context](https://github.com/pmndrs/valtio/wiki/How-to-use-with-context)
+- [How to split and compose states](https://github.com/pmndrs/valtio/wiki/How-to-split-and-compose-states)
```
dai-shi, line 16: Actually, I'm fine moving wiki pages into `docs/recipes/` and so on. (And leave links in those wiki pages.) Wiki pages were temporary solution because we didn't have docs. (and, it's editable without approval.)
What do you think?

dai-shi: Let me merge this PR anyway, and we can do changes later if we agree.

dai-shi (approved):

Issue #424, opened 2022-04-12
Snapshot not updated when using deep object with useEffect


devnomic:
I have a code similar like below structure,
I set the initial state on useEffect, in this case `state.item.amount = 500`

then on button click i want to read from snapshot `item.amount`,
but it shows empty string while i expect 500.
Why is that?

The strange thing, if i put the `item.amount` snapshot as react element (not on click function) It will show correct `item.amount`,
and click function work as expected.

So i think this is a bug? Since this is too big as gotcha.
Please share some light. Thanks

Code:

````
import { useEffect } from "react";
import { proxy, useSnapshot } from "valtio";

const state = proxy({
  item: {
    amount: "",
    attrs: {
      name: "Phone"
    }
  }
});
export default function App() {
  const { item } = useSnapshot(state);

  useEffect(() => {
    state.item.amount = "500";
  }, []);

  return (
    <div className="App">
      <h1>Name: {item.attrs.name}</h1>
      <button
        onClick={() => {
          console.log(item.amount);
        }}
      >
        Click
      </button>
    </div>
  );
}

````

Repro Link:
https://codesandbox.io/s/bold-sutherland-9kpudc?file=/src/App.js

If i put `item.amount` as element, everything will works as expected.
````
<h1>Name: {item.attrs.name} - {item.amount}</h1>
````

devnomic:
Thank you for the clarification  @dai-shi san

so how should i read the state on onClick function?
Is it safe to read state.item.amount directly when onClick function?

devnomic:
I just re-read your reply, it seems you mean the rule of thumb is read the state directly if you are inside function. Thanks again

PR #425, opened 2022-04-12
fix(readme): tweak rule of thumb description

dai-shi:
https://github.com/pmndrs/valtio/issues/424#issuecomment-1096374399

> I think we should fix "Rule of thumb" in readme.md.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -32,7 +32,7 @@ setInterval(() => {
 
 #### React via useSnapshot
 
-Create a local snapshot that catches changes. Rule of thumb: read from snapshots in React render function, otherwise use the source. The component will only re-render when the parts of the state you access have changed, it is render-optimized.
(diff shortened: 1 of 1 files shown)


RikuVan (approved):

readme.md
```diff
@@ -32,7 +32,7 @@ setInterval(() => {
 
 #### React via useSnapshot
 
-Create a local snapshot that catches changes. Rule of thumb: read from snapshots, mutate the source. The component will only re-render when the parts of the state you access have changed, it is render-optimized.
+Create a local snapshot that catches changes. Rule of thumb: read from snapshots in React render function, otherwise use the source. The component will only re-render when the parts of the state you access have changed, it is render-optimized.
```
dai-shi, line 35: Let's make it simpler.
```suggestion
Create a local snapshot that catches changes. Rule of thumb: read from snapshots in render function, otherwise use the source. The component will only re-render when the parts of the state you access have changed, it is render-optimized.
```

PR #436, opened 2022-04-17
breaking(utils): make proxyWithComputed to work with optional proxy-memoize

### LINKED ISSUE #435: Change proxy's value but proxyWithComputed's computed value  doesn't change
https://stackblitz.com/edit/vitejs-vite-1zhh5q?file=src/App.tsx&terminal=dev

For updates correctly, I need to read data.length
```
const state: { data: string[]; filter: string } = {
  data: [],
  filter: '',
};
const store = proxyWithComputed(state, {
  todos(s) {
    // If access data.length, it works
    // if (!s.filter && s.data.length) return s.data;
    if (!s.filter) return s.data;
    return s.data.filter((i) => !i.includes(s.filter));
  },
});
```

dai-shi:
close #435

This is breaking change in behavior. The API is the same, but a user needs to memoize function on their end. This allows to configure `proxy-memoize`'s options or use some other library, and even they can use without memoization.

### DIFF
--- a/docs/utils/proxyWithComputed.mdx
+++ b/docs/utils/proxyWithComputed.mdx
@@ -8,49 +8,43 @@ description: ''
 
 #### add a computation to a proxy
 
-You can have computed values with dependency tracking with property access.
-Dependency tracking in `proxyWithComputed` conflicts with the work in `useSnapshot`.
-React users should prefer using `derive`.
-`proxyWithComputed` works well for some edge cases and for vanilla-js users.
+You can define own computed properties within a proxy.
+By combining with a memoization library such as
+[proxy-memoize](https://github.com/dai-shi/proxy-memoize),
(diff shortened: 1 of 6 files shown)


ahaoboy:
How about add an option ?
```
 proxyWithComputed(obj, fn, {memoizeSize: number}) 

    const memoizedGet = memoize(get, {size: memoizeSize} )

```

dai-shi:
Do you have a use case to customize the behavior? I would actually like to make it completely optional. Yeah, it opens up more usage. I'll make some changes.

ahaoboy:
The use of rgb as a cache in image processing generally requires a long cache list, although it can be implemented in other ways, but it is best if it is supported directly

dai-shi:
Now, you can do anything you want.

PR #437, opened 2022-04-17
chore(website): remark plugin adds codesandbox params to urls

RikuVan:
Had a go at adding the custom remark plugin to add codepen params to the urls for embeds in the docs as suggested in comments of this PR https://github.com/pmndrs/valtio/pull/396. If this does the job I guess this one is all that is needed.


RikuVan:
FYI @LuckyPigeon  ^

LuckyPigeon:
Seems it is. I'll try later, thanks @RikuVan

PR #442, opened 2022-04-20
chore: typo

ahaoboy:


dai-shi (approved):
Thanks!!! (Didn't know there were so many. There must be similar errors in jotai and zustand.)

PR #446, opened 2022-04-26
fix(examples): Unexpected token in todo example

wangliang101:


examples/todo/src/store.ts
```diff
@@ -1,17 +1,17 @@
 import { proxy, useSnapshot } from 'valtio'
 
 export interface Todo {
-  id: number
-  name: string
-  completed?: boolean
+  id: number;
```
dai-shi, line 4: I think this was what prettier did. How did no-semi give you an error??

wangliang101: When I started the todo demo according to README.md, I found an error 'Line 5:3: Parsing error: Unexpected token', so I suggested a fix. But after I migrated this demo with the latest create-react-app, I did not have this problem.

dai-shi: Interesting. Let's revisit if we find any env that will cause issues. I guess there will be some workarounds.

PR #447, opened 2022-04-27
feat: add connect util for React.Component

ahaoboy:
demo: https://stackblitz.com/edit/vitejs-vite-bsp6pv?file=src%2FApp.tsx&terminal=dev

```ts
const store = proxy({ count: 1 });
type Store = typeof store;
const mapState = (store: Store) => {
  return {
    c: store.count,
  };
};
type MapState = ReturnType<typeof mapState>;
type MapActions = ReturnType<typeof mapActions>;
const mapActions = (store: Store) => {
  return {
    inc: () => {
      store.count++;
    },
  };
};
type Props = {};
class A extends React.Component<MapActions & MapState & Props> {
  render() {
    return (
      <div>
        <h2>c:{this.props.c}</h2>
        <button onClick={this.props.inc}>inc</button>
      </div>
    );
  }
}

export default connect(store, mapState, mapActions)(A);

```


dai-shi:
Thanks for the suggestion. However, I'd like to declare that this is out of the scope of the official package. That said, it can be a good option for 3rd-party libraries. Please feel free to develop one.

We have a list of 3rd-party libraries here: https://github.com/pmndrs/valtio/wiki/3rd-Party-Libraries

PR #455, opened 2022-05-09
chore: devtools docs

ahaoboy:
devtools(state, name) is deprecated


dai-shi (approved):
Thanks for the chore!!

Issue #452, opened 2022-05-02
using valtio with skypack.dev


s-cork:
Currently it fails to import:
```js
>>> var mod = await import("https://cdn.skypack.dev/valtio")

Uncaught SyntaxError: The requested module '/-/use-sync-external-store@v1.0.0-ZzO0qn27iKOvDc2m57IV/dist=es2019,mode=imports/unoptimized/shim/index.js' does not provide an export named 'useSyncExternalStore' (at valtio.js:4:9)
```

Similar problem with zustand:
```js
>>> var mod = await import("https://cdn.skypack.dev/zustand")

Uncaught SyntaxError: The requested module '/-/use-sync-external-store@v1.1.0-2zD0nsSlmd6FK3zepjJs/dist=es2019,mode=imports/unoptimized/shim/with-selector.js' does not provide an export named 'useSyncExternalStoreWithSelector' (at zustand.js:4:9)
```

Both errors are from the `use-sync-external-store` library
versions prior to the inclusion of this library built using skypack.dev

dai-shi:
Hi, thanks for reporting.
I'm not sure if this is fixable on our end.
Hope someone has some suggestions.

PR #456, opened 2022-05-10
fix(build): dyn import in skypack (native ESM)

ahaoboy:
fix: dyn import in skypack (close: #452)
I compared the versions that works, this is the difference 

error
````
import {useSyncExternalStore} from "/-/use-sync-external-store@v1.1.0-2zD0nsSlmd6FK3zepjJs/dist=es2019,mode=imports/unoptimized/shim/index.js";
````
ok
```
import de from "/-/use-sync-external-store@v1.1.0-2zD0nsSlmd6FK3zepjJs/dist=es2019,mode=imports/unoptimized/shim/index.js";
```

### CODE REVIEW

**Comment by dai-shi** [comment]
Thanks for the investigation.
This looks like a valid workaround until `use-sync-external-store` provides ESM.
I'll add a commit for some comments.

**Comment by dai-shi** [comment]
https://codesandbox.io/s/react-forked-2d7r4r?file=/src/App.js This one seems working.

**Comment by dai-shi** [comment]
I'm a little confused. Why didn't/doesn't `'react'` causing the same issue?

https://github.com/pmndrs/valtio/blob/af26b60319ff71dcdc10e7173170272f936c6e1c/src/react.ts#L1

**Comment by dai-shi** [comment]
Maybe because it's a **peer** dependency.

**Comment by dai-shi** [comment]
I'm not 100% confident, but let's merge this and ship it. We will see then.

**Comment by dai-shi** [comment]
@ahaoboy I'm thinking about reverting this because @s-cork reported that it didn't fix. https://github.com/facebook/react/issues/24590

**Comment by s-cork** [comment]
When I do that I get:

```txt
use-sync-external-store-shim.development.js:17 Uncaught ReferenceError: process is not defined
    at use-sync-external-store-shim.development.js:16:3
    at createCommonjsModule (use-sync-external-store-shim.development.js:4:10)
    at use-sync-external-store-shim.development.js:15:22
 ```
 
 When I look at what's trying to be imported from `use-sync-external-store-shim` there are some references to `(process.env.NODE_ENV !== "production")`

**Comment by ahaoboy** [comment]
I've made an example,   if it's what you need? or if you can give some specific details of the problem
@s-cork 
[Web Platform (forked) - StackBlitz](https://stackblitz.com/edit/web-platform-vyhihd?file=index.html)

**Comment by ahaoboy** [comment]
> When I do that I get:
> 
> ```
> use-sync-external-store-shim.development.js:17 Uncaught ReferenceError: process is not defined
>     at use-sync-external-store-shim.development.js:16:3
>     at createCommonjsModule (use-sync-external-store-shim.development.js:4:10)
>     at use-sync-external-store-shim.development.js:15:22
> ```
> 
> When I look at what's trying to be imported from `use-sync-external-store-shim` there are some references to `(process.env.NODE_ENV !== "production")`

You should add some code before your script tag, like this:

```
    <script>
      globalThis.process = { env: { NODE_ENV: "dev" } };
    </script>
```

PR #464, opened 2022-05-23
chore: adjust button style

btea:
Before:
image

After: 
image


RikuVan (approved):
I have not been able to reproduce this issue but the picture looks credible and I can't see any harm in the wrapping div.

PR #528, opened 2022-08-18
feat(vanilla): unstable_buildProxyFunction instead of unstable_getHandler

### LINKED ISSUE #522: custom snapshot method
Similar to #511, developers should be able to override the default [`createSnapshot` method](https://github.com/pmndrs/valtio/blob/7c85c1413992a02e7a65e698b803fdef500d008e/src/vanilla.ts#L135-L137) used internally by valtio that's accessed by `SNAPSHOT`. 

This in effect would also allow developers to decide when to re-render since `useSnapshot` internally [depends on this method](https://github.com/pmndrs/valtio/blob/7c85c1413992a02e7a65e698b803fdef500d008e/src/react.ts#L122).

---

For context, I'm integrating immer into valtio since immer provides a nice API for producing patches via `createDraft` and `finishDraft`. 

However, because the object given by `createDraft` is deeply _mutable_, a snapshot of it via `createSnapshot` will always reflect recent mutations. The result is [`lastSnapshot` will equal `nextSnapshot`](https://github.com/pmndrs/valtio/blob/7c85c1413992a02e7a65e698b803fdef500d008e/src/react.ts#L128-L133) and never re-render.

To get around this, immer provides a method called [`current` which provides a snapshot that doesn't reflect future mutations](https://immerjs.github.io/immer/current). In valtio, this would be used when a draft is detected.

```ts
      if (prop === SNAPSHOT) {
        return isDraft(target) ? current(target) : createSnapshot(target, receiver)
      }
```

https://codesandbox.io/s/cool-fire-fol6zz?file=/src/App.js

Notice how "edit" mutates the proxy but doesn't re-render. This is because no difference in snapshots is detected.

dai-shi:
close #522

- [x] first impl
- [x] second impl
- [x] update docs
- [x] third impl
- [x] fourh impl
- [x] fifth impl
- [x] sixth impl

This is a breaking change on unstable features.

### Use deepEqual instead of Object.is

```js
import { unstable_buildProxyFunction } from 'valtio'
export { useSnapshot } from 'valtio'
export const [proxy] = unstable_buildProxyFunction(deepEqual)
```

### Attach `get` trap to proxy handler

```js
import { unstable_buildProxyFunction } from 'valtio'
export { useSnapshot } from 'valtio'
export const [proxy] = unstable_buildProxyFunction(
  undefined,
  (target, handler) => {
    handler.get = ...
    return new Proxy(target, handler)
  }
)
```

### Avoid some objects to proxy instead of using `ref` for all

```js
import { unstable_buildProxyFunction } from 'valtio'
export { useSnapshot } from 'valtio'
const canProxyOrig = unstable_buildProxyFunction()[7]
export const [proxy] = unstable_buildProxyFunction(
  undefined,
  undefined,
  (x) => {
    if (...) {
      return false
    }
    return canProxyOrig(x)
  }
)
```


src/vanilla.ts
```diff
@@ -49,167 +48,195 @@ const snapshotCache = new WeakMap<
...
+        Object.defineProperty(baseObject, key, desc)
       } else {
-        nextValue = value
+        proxyObject[key as keyof T] = initialObject[key as keyof T]
       }
-      Reflect.set(target, prop, nextValue, receiver)
-      notifyUpdate(['set', [prop], value, prevValue])
-      return true
-    },
+    })
```
BrianHung, line 232: Could be simplified by copying of keys in `Reflect.ownKeys(initialObject)` into the default `createProxy` method, which also allows developer to decide which keys from initialObject to keep.

```ts
const proxyObject = customCreateProxy(initialObject, handler);
proxyCache.set(initialObject, proxyObject);
return proxyObject;


const createProxy = <T extends object>(
  initialObject: T,
  handler: ProxyHandler<T>
): readonly [T, T] => {
  const baseObject = Array.isArray(initialObject)
    ? []
    : Object.create(Object.getPrototypeOf(initialObject))
  const proxyObject = new Proxy(baseObject, handler)

  Reflect.ownKeys(initialObject).forEach((key) => {
    const desc = Object.getOwnPropertyDescriptor(
      initialObject,
      key
    ) as PropertyDescriptor
    if (desc.get || desc.set) {
      Object.defineProperty(baseObject, key, desc)
    } else {
      proxyObject[key as keyof T] = initialObject[key as keyof T]
    }
  })
  return proxyObject
}
```

src/vanilla.ts
```diff
@@ -49,167 +48,195 @@ const snapshotCache = new WeakMap<
...
+const buildProxyFunction = (
+  customizeObjectIs: (fn: typeof Object.is) => typeof Object.is = identity,
+  customizeCanProxy: (fn: typeof canProxy) => typeof canProxy = identity,
+  customizeCreateSnapshot: (
+    fn: typeof createSnapshot
+  ) => typeof createSnapshot = identity,
+  customizeCreateProxy: (
+    fn: typeof createProxy
+  ) => typeof createProxy = identity
+) => {
```
BrianHung, line 115: Maybe change this to argument object to be less unwieldy:

```ts
const buildProxyFunction = ({
  objectIs = Object.is,
  canProxy = canProxy,
  createSnapshot  = createSnapshot,
  createProxy = createProxy, // default methods
} : {
   objectIs: typeof Object.is,
  canProxy: typeof canProxy,
  createSnapshot: typeof createSnapshot,
  createProxy: typeof createProxy, // types 
}) => {
```

BrianHung (commented):
For more context, here's my fork of valtio that I made to work with immer: 
https://gist.github.com/BrianHung/ec6c0ce729acfa0361334c128eeef7ef

Noticeable change is that by forking valtio, I can write back to the initial object and access it as well.

src/vanilla.ts
```diff
@@ -3,7 +3,6 @@ import { getUntracked, markToTrack } from 'proxy-compare'
 const VERSION = __DEV__ ? Symbol('VERSION') : Symbol()
 const LISTENERS = __DEV__ ? Symbol('LISTENERS') : Symbol()
 const SNAPSHOT = __DEV__ ? Symbol('SNAPSHOT') : Symbol()
-const HANDLER = __DEV__ ? Symbol('HANDLER') : Symbol()
 const PROMISE_RESULT = __DEV__ ? Symbol('PROMISE_RESULT') : Symbol()
 const PROMISE_ERROR = __DEV__ ? Symbol('PROMISE_ERROR') : Symbol()
```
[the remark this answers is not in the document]
dai-shi: For now, I still want to hide them. I might consider combining into one symbol later (then it can ease customizing.) I first tried in this PR, but it made things complicated.

src/vanilla.ts
dai-shi: Then, you can't reuse the existing functions. One point of this style is to avoid exporting other functions (and symbols).

src/vanilla.ts
```diff
@@ -49,167 +48,195 @@ const snapshotCache = new WeakMap<
...
+    fn: typeof createSnapshot
+  ) => typeof createSnapshot = identity,
+  customizeCreateProxy: (
+    fn: typeof createProxy
+  ) => typeof createProxy = identity
+) => {
+  const customObjectIs = customizeObjectIs(Object.is)
+  const customCanProxy = customizeCanProxy(canProxy)
+  const customCreateSnapshot = customizeCreateSnapshot(createSnapshot)
+  const customCreateProxy = customizeCreateProxy(createProxy)
```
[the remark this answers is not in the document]
dai-shi: My imaginary use case is something like:
```js
const customizeCanProxy = (origCanProxy) => {
  const canProxy = (x) => {
    if (...) return true/false
    return origCanProxy(x)
  }
  return canProxy
}
```

src/vanilla.ts
dai-shi: I'm not sure if I understand your comment.

dai-shi:
> I can write back to the initial object and access it as well.

This PR should allow such use cases too.

```js
customizeCreateProxy = (origCreateProxy) => {
  const createProxy = (initialObject, handler) => {
    return [initialObject, new Proxy(initialObject, handler)]
  }
  return createProxy
}
```

I didn't try it though.

src/vanilla.ts
BrianHung: I think exporting the internals would be better, since to reuse existing functions you already need to know how they work.  

Also since custom proxy is marked as unstable, developers could also understand that applies to exported internals too.

src/vanilla.ts
BrianHung: I’m suggesting to move the for loop which copies keys from the initial object into the proxy object into the default createProxy function. 

So developers can also fully customize which keys to copy if they decide not to use initialObject.

src/vanilla.ts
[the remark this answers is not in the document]
BrianHung: Similar to the comment I left below, but I think most developers will either use all-or-none of the default behavior, instead of a mix and match. So exporting and exposing defaults would be better than passing defaults as arguments into a higher order function that the developer returns.

src/vanilla.ts
dai-shi: Do you mean to export many unstable functions? Anyway, if this is just a preference in styles rather than capability, I'd go with this for this time.

src/vanilla.ts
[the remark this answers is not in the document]
BrianHung: Here’s an example of using one symbol only: https://github.com/immerjs/immer/blob/e28b53f2d12696954b1f0125494498e34d77a3df/src/core/proxy.ts#L55

It uses one symbol to access an object which in valtio’s case can contain the listeners and version.

src/vanilla.ts
dai-shi: Ah, I probably see your point.
I have considered that, and for this time, I decided not to add it after the consideration.
If we got use cases require such capability, we consider it in a follow up PR.

(after all, it's the balance between customization and forking.)

src/vanilla.ts
BrianHung: Yes. Yeah it’s a preference in style only.

src/vanilla.ts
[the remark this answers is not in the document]
dai-shi: Yep. I wish to combine `SNAPSHOT` too. Removing symbols completely and using WeakMaps might be possible, but maybe performance-wise a little worth. Anyway, let's tackle it later. I would like to know what use cases this don't cover (and then the solution is either expose more or fork.)

src/vanilla.ts
dai-shi: Hm, maybe I can do that in this PR. Let me try.

dai-shi:
9bc48a3 is to expose more functions. I'm not sure if it's worth the bundle size increase.
1% of users would use custom objectIs, 0.1% of users would use custom canProxy, 0.01% of users would use custom createSnapshot, and the last one is only for you (and I'm not even sure if you would use it.)
In this approach, I think three symbols are actually good. (otherwise, we would build all four functions.)

dai-shi:
> otherwise, we would build all four functions.

Hmm, let me try this.

src/vanilla.ts
```diff
@@ -43,240 +15,296 @@ type Op =
...
+                return (value as any)[PROMISE_RESULT]
+              }
+              throw errorOrPromise
+            },
+          })
+        }
+      } else if (value?.[SNAPSHOT]) {
+        snapshot[key] = value[SNAPSHOT]
+      } else {
+        snapshot[key] = value
```
BrianHung, line 103: Is this change from `value?.[LISTENERS]` to `value?.[SNAPSHOT]` intentional? 

- Previously, it was checking if the value was a valtio proxy -- since valtio proxies return a `Set` for `Listeners` and if true, then call `SNAPSHOT`. Maybe there should be a way, that given a valtio proxy, you're able to tell if it's a valtio proxy from the outside i.e. `value?.[VALTIO_STATE]`.
- Not sure when `value[SNAPSHOT]` will return `undefined` or `null` in `createSnapshot`. It is possible custom `snapshot` method will intentionally return `undefined` or `null` too.
- Checking if `value?.[SNAPSHOT]` is undefined, then assigning it to `snapshot[key]` seems like redundant computation as well.

dai-shi: Oh! good point. It was intentional, but wrong way in #533.
I'll revert it in this PR.
btw, snapshot will always be an object.

src/vanilla.ts
```diff
@@ -43,240 +15,296 @@ type Op =
...
+  ,
+  ,
+  ,
+  ,
+  ,
+  proxy,
+  getVersion,
+  subscribe,
+  snapshot,
+] = buildFunctions()
```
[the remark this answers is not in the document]
dai-shi: I know what you mean, but I'm fearful about the bundle size increase.

This will be second iteration. combining symbols or so will improve a little.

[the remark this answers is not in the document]
dai-shi: Hmm, maybe we don't need to follow the order.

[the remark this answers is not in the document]
BrianHung: Ah, didn't know returning as array vs object resulted in difference in tree-shaking. 👀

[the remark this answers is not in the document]
dai-shi: It's minification, not tree-shaking.
In terms of tree-shaing, this PR makes it really worse, because, for example, `ref` is rarely used and can be tree-shaked previously, but this PR prevents it.
That said, it's fairly micro optimization, and capability would be more important.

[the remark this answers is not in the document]
dai-shi: Wait... I think we still should create public functions outside. Otherwise, useSnapshot must be re-created.

BrianHung:
> https://gist.github.com/BrianHung/ec6c0ce729acfa0361334c128eeef7ef

My overall use case and how it differs from valtio currently is:
- custom snapshot method
- custom proxy handler `get`, `deleteProperty`, `set`
- access to `LISTENERS` and `SNAPSHOT` to build those custom methods

src/vanilla.ts
```diff
@@ -43,177 +15,241 @@ type Op =
...
+    !(x instanceof WeakSet) &&
+    !(x instanceof Error) &&
+    !(x instanceof Number) &&
+    !(x instanceof Date) &&
+    !(x instanceof String) &&
+    !(x instanceof RegExp) &&
+    !(x instanceof ArrayBuffer),
+
+  PROMISE_RESULT = __DEV__ ? Symbol('PROMISE_RESULT') : Symbol(),
+  PROMISE_ERROR = __DEV__ ? Symbol('PROMISE_ERROR') : Symbol(),
```
BrianHung, line 60: Should these be moved to `shared state` and be next to `VERSION`, `LISTENERS`, `SNAPSHOT`.

BrianHung: Is it also possible to pass `VERSION`, `LISTENERS`, `SNAPSHOT` into `buildProxyFunction` so that a custom `createSnapshot` and custom `proxyFunction` could access these methods, but they're not directly exported?

dai-shi:
> * custom snapshot method

Pass your own `createSnapshot`.

> * custom proxy handler `get`, `deleteProperty`, `set`

Pass your own `newProxy`.

> * access to `LISTENERS` and `SNAPSHOT` to build those custom methods

Run `buildProxyFunction` to get them.

> so that a custom createSnapshot and custom proxyFunction could access these methods

I might be misunderstanding something in my approach. If you have some ideas, please feel free to share code snippets.

BrianHung:
> Run `buildProxyFunction` to get them.
> The trick is you would need to call buildProxyFunction function twice.

Oh my confusion was over this since I didn't realize you could access `LISTENERS` and `SNAPSHOTS`, just with a second initialization. 😅

dai-shi:
Please check out some examples in the PR description, if you haven't.

dai-shi:
I agree it's confusing, but no other ideas at this moment.
It should at least cover the three use cases listed in the description.
Let's ship it. We may expect some refactors in the future anyway.

Issue #516, opened 2022-08-06
Unexpected behavior using proxyWithHistory


LukasSchwarz4202:
Hello,

working with valtio's-proxyWithHistory I stumbled over a case with unexpected behavior.
It is quite hard to explain so I recreated the problem in codesandbox.io: https://codesandbox.io/s/valtio-proxywithhistory-undo-redo-l0360p

dai-shi:
Hmm, it looks like a bug. I'm not sure how it can cause it. Can you investigate more in detail?

LukasSchwarz4202:
Via https://github.com/pmndrs/valtio/blob/7c85c1413992a02e7a65e698b803fdef500d008e/src/vanilla.ts#L188 we ultimately reach line https://github.com/pmndrs/valtio/blob/7c85c1413992a02e7a65e698b803fdef500d008e/src/vanilla.ts#L208 where the assignment of the wrong value happens.
I think it has to do with assigning reference vs assigning deep copy of an object (I hope I could help; am still quite new to Javascript)

dai-shi:
Hmm, but "deep copy" is by design in atomWithHistory. Assigning reference won't work.
Why wrong value can happen?

LukasSchwarz4202:
Another try: could it be that there is a problem using proxyCache for an object being an array:
https://github.com/pmndrs/valtio/blob/7c85c1413992a02e7a65e698b803fdef500d008e/src/vanilla.ts#L56

dai-shi:
I think it will work with arrays, but you can try disabling it as it's just a cache:
```diff
- const found = proxyCache.get(initialObject) as T | undefined
+ const found = undefined
```

LukasSchwarz4202:
> I think it will work with arrays, but you can try disabling it as it's just a cache:
> 
> ```diff
> - const found = proxyCache.get(initialObject) as T | undefined
> + const found = undefined
> ```

This fixes the bug. What do you think: how serious is the impact of disabling the proxyCache on performance?

dai-shi:
We shouldn't disable it.
(Now, I'm confused what's the real problem. Do we modify the original array? I thought it's frozen. Hmm.)

LukasSchwarz4202:
Could the problem be that the lowest level array is just an array of numbers and not an array of proxies of numbers?
[...]

PR #531, opened 2022-08-27
fix(utils): fix proxyWithHistory behavibor with nested objects

LukasSchwarz4202:
## Related Issues

Test to reproduce issue #516.

And, a fix.


dai-shi:
Let's wait for #528, which might provide a better option.

dai-shi:
> Let's wait for #528, which might provide a better option.

On second thought, deepClone can be more efficient, because it only happens for undo/redo. (The other method I had in mind is to do copy on `saveHistory`.)

dai-shi (approved):

PR #545, opened 2022-09-10
fix(vanilla, react): use experimental_use and some refactors

dai-shi:
Ref: https://github.com/facebook/react/pull/25084

In addition:
- Refactor with unified Symbol `PROXY_STATE`.
- Precise `SNAPSHOT` type with Awaited.


src/react.ts
```diff
@@ -1,7 +1,17 @@
-import { useCallback, useDebugValue, useEffect, useMemo, useRef } from 'react'
 import {
-  affectedToPathList,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  experimental_use,
```
SukkaW, line 4: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62189

It seems that the typing of `experimental_use` is added to the `@types/react`, and `ts-ignore` could be removed.

dai-shi: Oh, yeah, I updated [there](https://github.com/jotai-labs/jotai-signal) and [there](https://github.com/jotai-labs/jotai-uncontrolled), but forgot here. Thanks.

dai-shi:
According to https://github.com/reactjs/rfcs/pull/229, the use of `use` here is violating the rule/convention of React.
From our perspective, this PR is not breaking.
When React finalizes the spec and release a stable version, we will consider following their pattern, and in that case, our library comes with **breaking** changes. i.e. snapshot will not resolve promises.
Meanwhile, let's merge this and release a patch version.

src/react.ts
```diff
@@ -14,6 +24,38 @@ import type { INTERNAL_Snapshot as Snapshot } from './vanilla'
 
 const { useSyncExternalStore } = useSyncExternalStoreExports
 
+// customized version of affectedToPathList
+// we need to avoid invoking getters
```
stephenh, line 28: @dai-shi ahhh, okay, this change (skipping getters in the output of `affectedToPathList`) is why I've been confused.

I've been watching the output of `useAffectedDebugValue` (maybe a little too religiously in retrospect) and was very confused by the lack of getters.

This made me think "valtio is not working" / "doesn't know my getters were invoked", when obviously I could tell it did, b/c the re-renders still worked correctly.

I finally reproduced the behavior in `proxy-compare` and "get it"--getters (and method invocations) are always in the affected map, and are always handled correctly by `isChanged`, it is merely `affectedToPathList` that is leaving them off.

What was the rationale here? I'm sure it makes sense, just curious b/c it tripped me up. Wondering if I can work the rationale/behavior in a doc/"gotchas" update.

Thanks!

dai-shi: Without it, it loops infinitely and hangs, AFAIR.

PR #548, opened 2022-09-11
docs: fix inline code format

kidonng:
## Related Issues

N/A

## Summary

Markdown does not work when interleaved with HTML, inline code should use `<code>` instead of backticks.

## Check List


dai-shi (approved):
Thanks!

PR #555, opened 2022-09-29
fix: subscribers should not be called after unsubscribe

jluxenberg:
## Related Issues

N/A

## Summary

There is an edge case in the subscribe notification logic which can cause a subscriber to be called after it has been unsubscribed.

To reproduce the issue:
  1. create a subscription "A"
  2. create a subscription "B"
  3. in the listener for "A", unsubscribe "B"

Expected result:
- "B" is never called (assuming listeners are processed in the order they were added)

Actual result:
- "B" is called once, and then never again

The fix is to check whether the subscription has been unsubscribed before firing the subscription's notification callback.

We ran into issues with this causing a "state update after unmount" error in our React app:
```
Can't perform a React state update on an unmounted component. This is a no-op, but 
it indicates a memory leak in your application. To fix, cancel all subscriptions and 
asynchronous tasks in a useEffect cleanup function.
```

and this fixes the issue for us.

## Check List

### DIFF
--- a/tests/subscribe.test.tsx
+++ b/tests/subscribe.test.tsx
@@ -44,7 +44,7 @@ describe('subscribe', () => {
     const obj = proxy({ count: 0 })
     const handler = jest.fn()
 
-    const unsubscribeA = subscribe(obj, () => {
+    const _unsubscribeA = subscribe(obj, () => {
       unsubscribeB()
     })
 
--- a/tests/subscribe.test.tsx
+++ b/tests/subscribe.test.tsx
@@ -40,6 +40,22 @@ describe('subscribe', () => {
     expect(handler).toBeCalledTimes(0)
   })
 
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Oh! Very nice catch! LGTM
Thanks!

tests/subscribe.test.tsx
```diff
@@ -40,6 +40,22 @@ describe('subscribe', () => {
     expect(handler).toBeCalledTimes(0)
   })
 
+  it('should be able to unsubscribe from a subscriber', async () => {
+    const obj = proxy({ count: 0 })
+    const handler = jest.fn()
+
+    const unsubscribeA = subscribe(obj, () => {
```
dai-shi, line 47:
```suggestion
    const _unsubscribeA = subscribe(obj, () => {
```

PR #565, opened 2022-10-08
fix(type): proxyWithHistory

dai-shi:
#561 change was bad for `proxyWithHistory`. The type alias is somehow inlined. This seems to fix it.


dai-shi:
We need to merge this for #564

PR #567, opened 2022-10-10
fix(utils): Return accurate data at toJSON of proxySet/proxyMap

chiefGui:
## Summary

**Please read https://github.com/pmndrs/valtio/pull/567#issuecomment-1273657366.**

~~Although `proxySet` and `proxyMap` have feature parity with native `Set` and `Map`, these proxies are not caught by `instanceof Set` and/or `instanceof Map`. This can be a problem when we want to validate whether a certain proxied value is either a `Set` or a `Map`, because prior to this patch, these proxies are not understood as being actual instances of `Set`/`Map`. Check out these examples:~~

```ts
if (proxySet([1, 2, 3]) instanceof Set) {
  // this code will never run because this condition will always be `false`
}

if (proxyMap() instanceof Map) {
  // this code will never run because this condition will always be `false`
}
```

~~After this patch, the conditions above will be satisfied and return `true`.~~

~~**NOTE:** I don't know if `proxySet` and `proxyMap` are actually expected to _not_ be understood as an instance of `Set`/`Map`. If that's the case, please feel free to dismiss this Pull Request although I would appreciate an explanation as to why and if there's any way I can have my `Set`/`Map` proxies to get caught by `instanceof <Set|Map>` at implementation-level.~~

---

~~To give you a use-case: I am using [`teleport-javascript`](https://github.com/postmanlabs/teleport-javascript) to serialize data. And it relies on `instanceof <Set|Map>` to be able to serialize them: https://github.com/postmanlabs/teleport-javascript/blob/master/index.js#L239. Because valtio's `proxySet`/`proxyMap` don't pass through these conditions, `teleport-javascript` is not able to serialize them.~~

## Check List

### DIFF
--- a/src/utils/proxyMap.ts
+++ b/src/utils/proxyMap.ts
@@ -70,7 +70,7 @@ export function proxyMap<K, V>(
       return this.data.length
     },
     toJSON() {
-      return {}
+      return new Map(this.data)
     },
     forEach(cb) {
       this.data.forEach((p) => {
--- a/src/utils/proxyMap.ts
+++ b/src/utils/proxyMap.ts
@@ -92,7 +92,6 @@ export function proxyMap<K, V>(
     [Symbol.iterator]() {
       return this.entries()
     },
-    __proto__: Map.prototype,
   })
 
   Object.defineProperties(map, {
--- a/src/utils/proxyMap.ts
+++ b/src/utils/proxyMap.ts
@@ -92,6 +92,7 @@ export function proxyMap<K, V>(
     [Symbol.iterator]() {
       return this.entries()
     },
+    __proto__: Map.prototype,
   })
 
   Object.defineProperties(map, {
--- a/src/utils/proxySet.ts
+++ b/src/utils/proxySet.ts
@@ -60,7 +60,7 @@ export function proxySet<T>(initialValues?: Iterable<T> | null): Set<T> {
       return 'Set'
     },
     toJSON() {
-      return {}
+      return new Set(this.data)
     },
     [Symbol.iterator]() {
       return this.data[Symbol.iterator]()
--- a/src/utils/proxySet.ts
+++ b/src/utils/proxySet.ts
@@ -76,6 +76,7 @@ export function proxySet<T>(initialValues?: Iterable<T> | null): Set<T> {
       // array.entries returns [index, value] while Set [value, value]
       return new Set(this.data).entries()
     },
+    __proto__: Set.prototype,
   })
 
   Object.defineProperties(set, {
--- a/src/utils/proxySet.ts
+++ b/src/utils/proxySet.ts
@@ -76,7 +76,6 @@ export function proxySet<T>(initialValues?: Iterable<T> | null): Set<T> {
       // array.entries returns [index, value] while Set [value, value]
       return new Set(this.data).entries()
     },
-    __proto__: Set.prototype,
   })
 
   Object.defineProperties(set, {
--- a/tests/proxyMap.test.tsx
+++ b/tests/proxyMap.test.tsx
@@ -103,12 +103,6 @@ const inputValues = [
 
 describe('features parity with native Map', () => {
   initialValues.forEach(({ name, value }) => {
-    it(`is an instance of Map`, () => {
-      const map = proxyMap(value as any)
-
-      expect(map).toBeInstanceOf(Map)
-    })
-
     it(`Support Map operations on ${name}`, () => {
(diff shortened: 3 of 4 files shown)


chiefGui:
I introduced a new commit (https://github.com/pmndrs/valtio/pull/567/commits/d8951dc23812c8f93765b4625da39f48ec7f34d2) to the table.

It basically returns accurate data through the `toJSON` method of the proxied `Set`/`Map`. Previously, an empty, literal `{}` object was being returned, and although it satisfies the final, serialized data of `JSON.stringify()`, it doesn't work if your `JSON.stringify()` has a custom `replacer` function.

_Before the commit:_
```ts
const proxiedSet = proxySet([1, 2, 3])

JSON.stringify(proxiedSet, (_, v) => {
  console.log(v instanceof Set) // false
  console.log(v) // `v` here is the empty, literal {} object.
})
```

_After the commit:_
```ts
const proxiedSet = proxySet([1, 2, 3])

JSON.stringify(proxiedSet, (_, v) => {
  console.log(v instanceof Set) // true
  console.log(v) // `v` here is the native Set version of the state.
})
```

The examples above just display a demonstration for `proxySet`, but my commit satisfies both `proxySet` and `proxyMap`.

chiefGui:
I've just reverted my first commit via https://github.com/pmndrs/valtio/pull/567/commits/419f5cd373b69f1b02437ba54a0b980ff671442f. It turns out that applying `__proto__: <Set|Map>.prototype` caused all sorts of (negative) side effects that I wasn't expecting at first, meaning `proxy<Set|Map>() instanceof <Set|Map>` will be returning `false` even after the merge of this Pull Request because the patch introducing the functionality is currently out of scope.

I'd like to keep this PR open though&mdash;it still resolves the issue brought up at https://github.com/pmndrs/valtio/pull/567#issuecomment-1273615992.

Nonetheless, after some more thought, @dai-shi: do you think `valtio` wants its `proxySet`/`proxyMap` to be understood as an instance of `Set`/`Map`? I mean, I can see the outright benefits of it&mdash;specially when it comes to the interoperability with third-party libraries we don't have control of, such as the mentioned `transport-javascript` case&mdash;but as I observed, that may cause unexpected issues if `valtio` lacks support for a more internal functionality a native `Set` or `Map` has. I mean, it may mean `valtio` would have to have a more accurate fidelity when it comes to `Set`/`Map` behavior and internals, which can be an unsustainable thing in the long term.

dai-shi:
> do you think `valtio` wants its `proxySet`/`proxyMap` to be understood as an instance of `Set`/`Map`?

Maybe not. proxySet and proxyMap are just "emulations" for Set and Map. So, `instanceof` returning `false` would probably be good for being explicit.

@fkhadra What do you think?

dai-shi (approved):
Fixing `toJSON` seems good to me.

fkhadra:
> > do you think `valtio` wants its `proxySet`/`proxyMap` to be understood as an instance of `Set`/`Map`?
> 
> Maybe not. proxySet and proxyMap are just "emulations" for Set and Map. So, `instanceof` returning `false` would probably be good for being explicit.
> 
> @fkhadra What do you think?

I tend to agree that returning `false` is the way to go. Although, I understand your argument @chiefGui when it comes to interoperability with third-party libraries. I'm afraid that at some point we cannot guarantee 100% feature parity with native Set/Map.

fkhadra (approved):
BTW, thanks for fixing the json implementation

chiefGui:
> BTW, thanks for fixing the json implementation

You got it @fkhadra!

--- 

By the way, yeah - just ignore my `instanceof` solution. After some exhaustive debugging on my end, it seems it brings more problems than it solves.

chiefGui:
Oh, and before I forget...

> I'm afraid that at some point we cannot guarantee 100% feature parity with native Set/Map.

Yes. And after re-reading valtio's documentation, I realised you explicitly mention `proxySet/proxyMap` are _emulations_. It was my bad trying to introduce the `instanceof` thing.

If anyone ever stumbles upon this matter in the future, one "trick" I did to be able to recognize a `value: any` is a `Set` or a `Map` that would include sets and maps from valtio is...

Instead of:
```ts
function doSomething(data: any) {
  if (data instanceof Set) {
    // do something...
  }
}
```

I'd do:
```ts
function doSomething(data: any) {
  if (data[Symbol.toStringTag] === 'Set') {
    // do something...
  }
}
```

dai-shi:
I hope to merge and release this with some other changes. Meanwhile, you can use the codesandbox build.

PR #573, opened 2022-10-15
fix(react): a workaround for missing named exports

### LINKED ISSUE #572: Next.js app fails to compile due to strict named exports (missing `experimental_use`)
This error is thrown during Next.js compilation with the latest version of Valtio.

```
file:///Users/jared/Code/apps/potent/node_modules/.pnpm/valtio@1.7.3_react@18.2.0/node_modules/valtio/esm/index.mjs:1
import { useRef, useCallback, experimental_use, useEffect, useMemo, useDebugValue } from 'react';
                              ^^^^^^^^^^^^^^^^
SyntaxError: Named export 'experimental_use' not found. The requested module 'react' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'react';
const { useRef, useCallback, experimental_use, useEffect, useMemo, useDebugValue } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:123:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:189:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:533:24)
    at async importModuleDynamicallyWrapper (node:internal/vm/module:438:15) {
  page: '/'
}
```

The normal mitigation for this in Webpack is to use the following config:
```ts
    config.module = {
      ...config.module,
      parser: {
        ...config.module.parser,
        javascript: {
          ...config.module.parser.javascript,
          exportsPresence: false,
        },
      },
    };
```

However this doesn't seem to work with Next. Maybe they use a version of Webpack that doesn't support it yet.

Love that you stay on the bleeding edge with React features! Was wondering if you would consider refactoring your import code to support Next e.g.

```ts
import pkg from 'react';
const { useRef, useCallback, experimental_use, useEffect, useMemo, useDebugValue } = pkg;
```

Otherwise, I'm happy to use `patch-package` as a last resort :)

dai-shi:
close #572

### DIFF
--- a/src/react.ts
+++ b/src/react.ts
@@ -1,7 +1,6 @@
 /// <reference types="react/experimental" />
 
-import {
-  experimental_use as use,
+import ReactExports, {
   useCallback,
   useDebugValue,
   useEffect,
@@ -23,6 +22,7 @@ import { snapshot, subscribe } from './vanilla'
 import type { INTERNAL_Snapshot as Snapshot } from './vanilla'
 
 const { useSyncExternalStore } = useSyncExternalStoreExports
+const use = ReactExports.experimental_use
 
 // customized version of affectedToPathList
 // we need to avoid invoking getters
--- a/src/react.ts
+++ b/src/react.ts
@@ -21,8 +21,8 @@ import useSyncExternalStoreExports from 'use-sync-external-store/shim'
 import { snapshot, subscribe } from './vanilla'
 import type { INTERNAL_Snapshot as Snapshot } from './vanilla'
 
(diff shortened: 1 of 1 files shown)


nornagon:
FWIW, this seems to cause issues when substituting preact/hooks for react, as preact/hooks has no default export:

```
The requested module 'preact/hooks' does not provide an export named 'default'
```

dai-shi:
That's really unsolvable from the library perspective. We hope React provides ESM, and meanwhile Preact provides CJS with `default` everything.

PR #593, opened 2022-11-22
bug: unrelated changes to state cause component to re-render

jluxenberg:
A component "App" which calls "useSnapshot" but does not reference any fields in the snapshot will re-render when an item in the proxy is updated.

Expected result:
- "App" should not re-render when "text" field is updated in state

Actual result:
- "App" re-renders on any change to "text" field in state

I don't have a fix for this yet, just a test to illustrate the issue.


dai-shi:
It's intentional/expected behavior.
https://github.com/pmndrs/valtio/issues/594#issuecomment-1324399272

PR #607, opened 2022-12-19
feat: adds reaction for automatic watch behavior

benjamind:
## Summary

I've been trying to add valtio to my workflow within [lit](https://lit.dev/) web components in a similar manner to my usage of `mobx` with the [lit-mobx](https://github.com/adobe/lit-mobx) library. i.e. transparently without the need to manually subscribe, or use the watch get function.

For mobx this was achieved using their [Reaction pattern](https://mobx.js.org/reactions.html#reaction), which allows us to have a data tracking function (in this case the lit [update](https://lit.dev/docs/api/ReactiveElement/#ReactiveElement.update) lifecycle method which is invoked just before render and is used to track the variables to respond to, and then an effect method which we use to request an update to the components. This split between tracking / effect allows for a more fine grained control of behavior.

I've implemented this by refactoring the watch method a little bit, and adding a basic automatic tracking of accessed proxies via the proxy `get` implementation. This seems to work pretty well, allowing for a very natural usage of properties without having to wrap each proxy access in a `get` call as we do with `watch` today. This makes adding observability via valtio to a lit element very natural since we can make it aware of observable properties through a mixin without having to rewrite the internal render functions.

As an example of usage within a valtio adapted lit element base class:

```ts
export class ExampleElement extends LitElement {
  @property({ attribute: false })
  public counter?: Counter;

  public render(): TemplateResult {
    return html` Count is ${this.counter ? this.counter.count : "Not Set"} `;
  }
}
export class ValtioExampleElement extends ValtioReaction(ExampleElement) {};
```

We can see how this element could simply be based on regular LitElement, and will update when reference equality of the `counter` property changes, or could be upgraded to support a Valtio proxy on its Counter property simply by mixing in the `lit-valtio` library, with WebComponents this can be done by just subclassing the default implementation of the element. This approach has proved pretty popular with the web components community since it allows you to build your components very agnostic of the state library.

## Check List


benjamind:
Could be yeah. Though as you can see it shares much of the same implementation so I thought it would save payload this way.

How would you suggest I extend the default proxy function with the tracking set? I did have this implemented as an extension using buildProxyFunction but then of course all proxies need to switch to using that modified proxy function. Or am I missing something?

dai-shi:
> but then of course all proxies need to switch to using that modified proxy function. 

Yes, it's unfortunate DX-wise.

I'd expect something like this if it were a separate package.
```js
import {
  proxy, // is an extended proxy function with tracking capability
  reaction,
} from 'valtio-reaction'
```

So, maybe:
```js
import {
  proxyForReaction as proxy,
  reaction,
} from 'valtio/utils'
```

---

> I thought it would save payload this way.

I think  it's true only if people use both `watch` and `reaction`, which is not very likely.
I'm not so against modifying `watch` assuming it's probably not widely used. (I think we can even drop it if the new one covers the use case.)

benjamind:
How would you feel about a small additional to vanilla to achieve the tracking behavior, with the bulk of the rest of the reaction implementation in a util? (have updated the PR with this approach, and handled the case you called out in discussion thread).

This would then keep it compatible with other extensions of proxy and not force users to make a choice at model creation time.

dai-shi:
https://github.com/pmndrs/valtio/discussions/609#discussioncomment-4463224

> That's why adding `tracking` in vanilla.ts is unacceptable.

We should explore how double proxies work and then if unstable_buildProxyFunction works better instead. I know it's sub-optimal. In the end, the last resort might be adding a hook for get handler (but this is against the philosophy and the purpose of unstable_buildProxyFunction. That's like a major change for v2 in the far future.)

benjamind:
Can you elaborate on what you mean by double proxies?

dai-shi:
> Can you elaborate on what you mean by double proxies?

I mean, wrapping with another proxy. You may not like it, but it's like how `useSnapshot` does.
```js
// src/utils/reaction.ts

export const proxyForReaction = (initialValue) => {
  const p = proxy(initialValue);
  return new Proxy(p, {
    get(...) { ... }
  });
};
```

benjamind:
Closing while rethinking the approach :-)

PR #614, opened 2023-01-03
fix(build): Clean up peer dependencies

### LINKED ISSUE #613: vite in peerDependencies and peerDependenciesMeta causes issues in pnpm workspace
## Summary

Valtio's package.json lists _vite_ in its peerDeps. This is unnecessary, and it has the unwanted side effect that pnpm - as a documented behavior - adds vite in valtio's symlink folder name whenever vite deps are also present in a workspace package. The problem is that depending of which of react or vite are present in a specific package, the valtio folder name is non-deterministic, which leads to double instances, which again leads to a Symbol-related bug in valtio.

## Link to reproduction

It's an intranet app, can't link it.

## Check List

- [x] I've already opened a [discussion](https://github.com/pmndrs/valtio/discussions) before opening this issue.

Here is an exact, reproducible scenario which leads to a runtime bug:

1. Create a pnpm workspace with 2 packages: _app_ and _common_
2. Make _common_ depend on vite and react
3. Make _app_ depend on react, but not on vite (app is using common's build interface)
4. Also make both _app_ and _common_ depend on valtio
5. Use valtio in _app or common_ (it doesn't matter) like this: `const store = proxyWithComputed({}, {})`
6. Result: valtio will throw a runtime error saying that it can't determine the target as a proxy object.

The explanation of the bug is as follows:

Because valtio has peer deps which are varyingly present in _app_ and _common_ (app has react, common has react + vite), there will be 2 separate valtio folders in the pnpm store: `valtio@ver+react@ver` (for _app)_ and `valtio@ver+react@ver+vite@ver` (for _common)._ When there's a call to _proxyWithComputed_, vite resolves its import path `valtio/utils` to one of the folders, but then _proxyWithComputed_ imports `valtio`, which vite resolves to the other folder. This leads to the circumstance that _proxyWithComputed_ doesn't recognize the javascript Symbol on the proxy which `valtio` produced.

I couldn't determine why Vite resolves one of the import specifiers (`valtio/utils`) to one of the folders and the other (`valtio`) to the other, so it might be possibly a Vite bug here, but I concluded that in the meantime, valtio's peer deps config seems unreasonable, and it hijacks the pnpm behavior.

There's a workaround for the problem, by adding a `.pnpmfile.cjs` which cancels valtio's peerDeps configs:

```js
function readPackage(pkg, context) {
  if (pkg.name === 'valtio') {
    pkg.peerDependencies = {}
    pkg.peerDependenciesMeta = {}
  }

  return pkg
}

module.exports = {
  hooks: {
    readPackage
  }
}
```

Note that it's necessary to empty both _peerDependencies_ and _peerDependenciesMeta._

sarimarton:
Clean up peer dependencies which are not actually peer deps and they cause non-deterministic pnpm behavior with projects consuming valtio. All the removed peer deps are already listed in devDeps which they seem to belong to.

## Related Issues

Fixes #613.

## Check List

### DIFF
--- a/package.json
+++ b/package.json
@@ -217,29 +217,9 @@
     "date-fns": "2.27.0"
   },
   "peerDependencies": {
-    "@babel/helper-module-imports": ">=7.12",
-    "@babel/types": ">=7.13",
-    "aslemammad-vite-plugin-macro": ">=1.0.0-alpha.1",
-    "babel-plugin-macros": ">=3.0",
-    "react": ">=16.8",
-    "vite": ">=2.8.6"
+    "react": ">=16.8"
   },
   "peerDependenciesMeta": {
-    "vite": {
-      "optional": true
-    },
-    "aslemammad-vite-plugin-macro": {
-      "optional": true
-    },
-    "@babel/helper-module-imports": {
-      "optional": true
-    },
-    "@babel/types": {
-      "optional": true
-    },
-    "babel-plugin-macros": {
-      "optional": true
-    },
     "react": {
       "optional": true
     }


dai-shi (approved):
Thought eslint complains, but doesn't seem so.
I think removing those is reasonable. Let's see if it causes some real problems.

PR #624, opened 2023-01-11
fix(vanilla): use case with refed proxies

dai-shi:
`ref()` is usually used for non plain objects, but what if we use it for valtio proxy objects.
It might be more intuitive to cut the structure with a ref'ed proxy.


dai-shi:
As #626 will hit a new minor version, let's put this into it too. I'm still skeptical about the use case, but it doesn't harm. (only a little size increase)

PR #625, opened 2023-01-13
(website): Feature/dark mode

roguesherlock:
I feel like it was long overdue. Finally got around to it!

4bdec02cb1a250b2512024991f22b110


dai-shi (approved):
Thanks! Looks good to me.

roguesherlock:
One thing I forgot to ask, should we also switch the codesandbox embeds to dark by default?

RikuVan (approved):

roguesherlock:
I've now updated to automatically also update the codesandbox embeds with theme change! 

Only caveat is that users may loose state if/when theme changes. But I think that is fine cause these embeds are only for examples. 

let me know what you guys think!

PR #626, opened 2023-01-14
feat(react/utils): Add useProxy() hook

sarimarton:
## Related Issues

Follow-up on https://github.com/pmndrs/valtio/discussions/620

## Summary

- [x] Adds useProxy() hook.
- [x] Deprecates valtio/macro and valtio/vacro/vite.
- [ ] Update readme

## Check List

### DIFF
--- a/src/react-utils/useProxy.ts
+++ b/src/react-utils/useProxy.ts
@@ -1,0 +1,37 @@
+import { useSnapshot } from '../react'
+import { useLayoutEffect } from 'react'
+
+/**
+ * useProxy
+ *
+ * Takes a proxy and returns a new proxy which you can use in both react render
+ * and in callbacks. For the best ergonomics, you can export a custom hook from
+ * your store, so you don't have to figure out a separate name for the hook
+ * reference. E.g.:
+ *
+ * export const store = proxy(initialState)
+ * export const useStore = () => useProxy(store)
+ * // in the component file:
+ * function Cmp() {
+ *   const store = useStore()
+ *   return <button onClick={() => {store.count++}}>{store.count}</button>
+ * }
+ *
+ * @param proxy
+ * @returns A new proxy which you can use in the render as well as in callbacks.
+ */
+export const useProxy = <T extends object>(proxy: T) => {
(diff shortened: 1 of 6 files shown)


src/react-utils/useProxy.ts
```diff
@@ -20,18 +20,20 @@ import { useLayoutEffect } from 'react'
...
-export const useProxy = <T extends object>(proxy: T) => {
-  const snapshot = useSnapshot(proxy)
+export function useProxy<T extends object>(proxy: T): T {
+  const snapshot = useSnapshot(proxy) as T
 
   let isRendering = true
 
   useLayoutEffect(() => {
+    // This is an intentional hack
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
sarimarton, line 30: Does the linter really error on this? Not passing the dep array is legit usage.

dai-shi: Yes, it does. What it says is useRef instead. It's legit, but not best practice in general?

sarimarton: Ah yeah, I see the warning, sorry. It's funny. I couldn't help figuring out a linter compliant version:

```tsx
const isRendering = { val: true }

useLayoutEffect(() => {
  isRendering.val = false
})
```

Now that you can call a hack 😜

dai-shi: Not worth increasing the bundle size. It's a hack anyway for either case, from the react perspective.

dai-shi:
https://codesandbox.io/s/react-typescript-forked-ix38xu

Deprecation message works.
image

dai-shi:
Hmm, I want to reorganize entry points. Will do it in a new PR along with updating readme.

dai-shi (approved):
Let's merge this. I will probably change `react-utils` to `react/utils` and then export from `utils` too.

src/react-utils/useProxy.ts
sarimarton: Of course, I was just kidding. Thanks @dai-shi for publishing 🥳 and all the great work!

PR #629, opened 2023-01-15
fix(build): do not use rollup alias

dai-shi:
Just noticed alias is not necessary.
Let's update terser plugin too.


dai-shi:
> Just noticed alias is not necessary.

Maybe? The output JS is identical, but *.d.ts import path is changed.

PR #641, opened 2023-01-28
Add all wiki pages

roguesherlock:
This adds all [wiki pages](https://github.com/pmndrs/valtio/wiki) to the `docs` folder verbatim. Once this is merged in, I'll update wiki pages to add links to the respective files.

### DIFF
--- a/docs/api/utils/derive.mdx
+++ b/docs/api/utils/derive.mdx
@@ -35,7 +35,7 @@ derive(
 )
 ```
 
-# underive
+## underive
 
 #### stop evaluating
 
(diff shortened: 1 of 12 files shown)


dai-shi (approved):
LGTM!

RikuVan (commented):
It is nice to have these added. TBH though there is overlap with other docs and the I feel the best version would be not to have this catch all but have it integrated with the rest of the docs 🤷

dai-shi:
@RikuVan Do you think it's okay to merge this PR and leave the refinement for someone?

roguesherlock:
I agree. I think the plan was to just move these verbatim for now so that we have all documentation at just one place and then refine it over time

dai-shi:
Let's merge this.
@roguesherlock Thanks for your work. Any followup work is appreciated.

roguesherlock:
@dai-shi sweet. I've updated all wiki pages to point to new docs now. I'll see if I can refine these docs next week!

PR #645, opened 2023-01-29
docs: Clarify behavior of watch.

stephenh:
## Summary

Just clarifying the non-usage tracking nature of `watch`.

## Check List

### DIFF
--- a/docs/api/advanced/subscribe.mdx
+++ b/docs/api/advanced/subscribe.mdx
@@ -14,9 +14,11 @@ You can access state outside of your components and subscribe to changes.
 ```jsx
 import { subscribe } from 'valtio'
 
-// Subscribe to all changes to state (and its children)
+const state = proxy({ count: 0 })
+
+// Subscribe to all changes to the state proxy (and its child proxies)
 const unsubscribe = subscribe(state, () =>
-  console.log('state has changed to', state)
+  console.log('state.count has changed to', state.count)
 )
 // Unsubscribe by calling the result
 unsubscribe()
--- a/docs/api/advanced/subscribe.mdx
+++ b/docs/api/advanced/subscribe.mdx
@@ -18,7 +18,7 @@ const state = proxy({ count: 0 })
 
 // Subscribe to all changes to the state proxy (and its child proxies)
 const unsubscribe = subscribe(state, () =>
-  console.log('state.count has changed to', state.count)
+  console.log('state has changed to', state)
 )
 // Unsubscribe by calling the result
 unsubscribe()
--- a/docs/api/advanced/subscribe.mdx
+++ b/docs/api/advanced/subscribe.mdx
@@ -14,7 +14,7 @@ You can access state outside of your components and subscribe to changes.
 ```jsx
 import { subscribe } from 'valtio'
 
-// Subscribe to all state changes
+// Subscribe to all changes to state (and its children)
 const unsubscribe = subscribe(state, () =>
   console.log('state has changed to', state)
 )
--- a/docs/api/utils/watch.mdx
+++ b/docs/api/utils/watch.mdx
@@ -9,7 +9,11 @@ description: 'watch for changes.'
 
 #### subscription via a getter
 
-This utility provides a specialized from of subscription. Proxy objects are tracked with a `get` function passed to the callback. Notice the callback will run eagerly as soon as a proxy is tracked, even if not yet mutated.
+This utility supports subscribing to multiple proxy objects (unlike `subscribe` which listens to only a single proxy). Proxy objects are tracked with a `sub` function passed to the callback.
+
+The callback will rerun anytime a `sub`-d proxy (or child proxy of that proxy) has any of its keys mutated, regardless of the keys accessed within the `callback`'s code (i.e. `watch` does not implement the usage tracking of `useSnapshot`, and the return value of `sub` is the same proxy passed in, and not a snapshot).
+
+Also note the callback will run once immediately when `watch` is called, even if not yet mutated, to establish the initial proxy subscriptions.
 
 ```js
 import { proxy } from 'valtio'
@@ -18,9 +22,12 @@ import { watch } from 'valtio/utils'
 const userState = proxy({ user: { name: 'Juuso' } })
 const sessionState = proxy({ expired: false })
 
-watch((get) => {
-  const expired = get(sessionState).expired
-  const name = get(userState).user.name
+watch((sub) => {
+  // Call sub to subscribe
+  sub(sessionState);
+  const expired = sessionState.expired
+  // Or call it inline
+  const name = sub(userState).user.name
   console.log(`${name}'s session is ${expired ? 'expired' : 'valid'}`)
 })
 // 'Juuso's session is valid'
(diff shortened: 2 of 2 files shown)


dai-shi (commented):
Thanks for the suggestion!
I actually like the idea of renaming `get` in docs. But not `sub`, because it reminds me of `subscribe` function which returns unsubscribe function. Do you have any other idea?

stephenh:
> But not sub, because it reminds me of subscribe function which returns unsubscribe function.

Ah yeah, that's fair; wdyt about `add`?

Because it's just adding the proxy to the list of proxies to watch...

...totally non-serious musing, but fwiw I'm wondering if a better API for watch would have been:

```js
watch(proxy, () => { ...logic... });
watch([proxy1, proxy2], () => { ... });
```

Because I think it makes it more clear that we're passing `proxy` solely for the purpose of getting it subscribed, and we don't need to be returned/passed back a special/decorated version of the `proxy`.

...granted, maybe someday `get` ends up returning a snapshot. :-) :shrug:

dai-shi:
Yeah, let's leave it as `get` for now. I understand the original confusion, but it's fairly fine. `sub` or `add` seems more confusing.

> ```js
> watch(proxy, () => { ...logic... });
> ```

Is it just same as `subscribe`?

> ```js
> watch([proxy1, proxy2], () => { ... });
> ```

It's actually much easier to implement. `watch` is more capable than that. It allows to use `get` conditionally, and will auto subscribe/unsubscribe on demand.

(tbh, if `watch` isn't widely used, we could simply drop it. it feels suboptimal. we could add a new util with usage tracking. might be as a third-party package. see also #609 discussion.)

stephenh:
> sub or add seems more confusing.

Okay, changed back to `get`.

Do the rest of the doc updates look okay?

> It allows to use get conditionally, and will auto subscribe/unsubscribe on demand.

Right, I know my proposal was less flexible, i.e. it could not conditionally watch _only_ `proxy1` or `proxy2`, but I liked the clarity of "param1 --> the proxies I'm watching, refire whenever they mutate"; "param2 --> the code that runs when that happens".

As-is, with the `get` interspersed in the code block, it made me jump to the conclusion that `watch` had mobx/`useSnapshot`-style usage tracking where the properties accessed by the `get` return values really mattered.

Moving to a two-arg style, which you're right is exactly like `subscribe`, imo just makes the mental model clearer as to what is happening.

All that said, of course/yeah totally fine to keep as-is. Maybe someday `get` starts returning snapshots and then the current API will make a ton of sense. :-)

> https://github.com/pmndrs/valtio/discussions/609

Fwiw I didn't really follow the proposed `reaction` / `track` API, but :shrug: that's fine...

Per your comment:

> That's why adding tracking in vanilla.ts is unacceptable.

I'll probably use valtio from the convenience of `useSnapshot` anyway, so I don't think it really matters to me, but fwiw personally I think usage based tracking is a really core component to a mature observer library, so would nudge for snapshots to eventually make their way into vanilla, but, again, probably doesn't really matter/totally up to you.

Thanks for the discussion!

docs/api/advanced/subscribe.mdx
```diff
@@ -14,9 +14,11 @@ You can access state outside of your components and subscribe to changes.
 ```jsx
 import { subscribe } from 'valtio'
 
-// Subscribe to all state changes
+const state = proxy({ count: 0 })
+
+// Subscribe to all changes to the state proxy (and its child proxies)
 const unsubscribe = subscribe(state, () =>
-  console.log('state has changed to', state)
+  console.log('state.count has changed to', state.count)
```
dai-shi, line 21: what was the confusing part before this change?

docs/api/utils/watch.mdx
```diff
@@ -9,7 +9,11 @@ description: 'watch for changes.'
 
 #### subscription via a getter
 
-This utility provides a specialized from of subscription. Proxy objects are tracked with a `get` function passed to the callback. Notice the callback will run eagerly as soon as a proxy is tracked, even if not yet mutated.
+This utility supports subscribing to multiple proxy objects (unlike `subscribe` which listens to only a single proxy). Proxy objects are subscribed with a `sub` function passed to the callback.
```
dai-shi, line 12:
```suggestion
This utility supports subscribing to multiple proxy objects (unlike `subscribe` which listens to only a single proxy). Proxy objects are subscribed with a `get` function passed to the callback.
```

dai-shi (commented):
The rest seems good.

docs/api/advanced/subscribe.mdx
stephenh: Iirc I think I had some other edit in mind while typing that, but now forget + yeah, not necessary, so reverted.

dai-shi:
> it made me jump to the conclusion that `watch` had mobx/`useSnapshot`-style usage tracking

I see. I feel like I'm failing to explain the basic conception. Valtio is built with very different mental model. It's not like observer/observable practice.
What valtio core (valtio/vanilla) does is mutable state -> immutable state conversion. It's close to the mental model of Immutable.js or Immer.

Only, useSnapshot (in valtio/react) does tracking. And, it confuses, I know. So, while typical observer pattern does the whole task altogether, valtio split the task into two parts. It was born because of the immutability requirement of React, but I notice valtio/vanilla itself is kind of new approach.

Anyway, it's the design decision, valtio core doesn't do tracking, does only one thing. Adding tracking-capable watch in `valtio/tracking` or somewhere else might make sense in that sense. (but then, it confuses what is "vanilla"..., hmm maybe it's fine there.)

It's probably doc issue, like it should clearly say, something is different from say mobx. We could name the hook as `useSnapshotWithTracking`, but it's too long. At the same time, we want people to use it without knowing too much in detail.

Thanks for your feedback!

---

> Maybe someday `get` starts returning snapshots and then the current API will make a ton of sense.

On second thought, for one reason, I like the idea, because current watch and subscribe is too close in their capability. Just conditional subscription. Adding tracking will expand its usage.

The remaining concern is it's very hard to implement. We did once in proxyWithComputed actually, but the code is hardly readable (now we recommend combining `proxy-memoize`). So, probably, while `subscribeWithTracking` is fairly easy to implement (it's like useSnapshot), `watchWithTracking` will be very very complicated. Not sure if it's worth, given that we don't know how many people would use it.

So, in conclusion, I'd suggest someone to try it as a third party library for a start.

dai-shi (approved):
Thanks for your contribution!

stephenh:
> I feel like I'm failing to explain the basic conception. 

Ah yeah, sorry, that's my fault; I do get that valtio is fundamentally different, I was just attempting to articulate how the current watch/get API, to me, had encouraged me as a still-newbie to assume "it's like mobx", even though yep I get that it's not.

> It's probably doc issue, like it should clearly say, something is different from say mobx

My hope is to contribute some docs in that regard, once I deeply understand valtio, but I'm only about ~1/3rd of the way there. :-) 

> because current watch and subscribe is too close in their capability. Just conditional
> subscription. Adding tracking will expand its usage.

Ah nice! Right! I think that's a great way of rationalizing the additional complexity (but increased power/utility to users).

Thanks for accepting the PR!

PR #646, opened 2023-01-30
chore(tests): migrate to vitest

Aslemammad:
## Related Issues

#643


Aslemammad:
The tests are passing, I do not know what is happening with types in the lower versions.

tests/basic.test.tsx
```diff
@@ -1,6 +1,6 @@
 import { StrictMode, useEffect, useRef, useState } from 'react'
-import { expect, it, jest } from '@jest/globals'
 import { fireEvent, render, waitFor } from '@testing-library/react'
+import { vi } from 'vitest'
```
dai-shi, line 3: Can we keep importing `expect` and `it` from `vitest`?
And, will it avoid adding types in tsconfig.json?

tests/computed.test.tsx
```diff
@@ -64,7 +57,7 @@ describe('proxyWithComputed', () => {
   })
 
   it('computed getters and setters', async () => {
-    const computeDouble = jest.fn((x: number) => x * 2)
+    const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 60: How is `computeDouble` typed without `: number`?

vitest.config.ts
```diff
@@ -0,0 +1,23 @@
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  define: {
+    __DEV__: true,
+  },
```
dai-shi, line 6: We no longer have `__DEV__`. You probably need to merge `main` here?

.eslintrc.json
```diff
@@ -47,10 +49,6 @@
     "@typescript-eslint/no-use-before-define": "off",
     "@typescript-eslint/no-empty-function": "off",
     "@typescript-eslint/no-explicit-any": "off",
-    "jest/consistent-test-it": [
-      "error",
-      { "fn": "it", "withinDescribe": "it" }
-    ],
```
dai-shi, line 53: Can we keep this? (Unless there's an alternative without jest plugin.)

.eslintrc.json
```diff
@@ -31,6 +31,8 @@
     }
   },
   "rules": {
+    "import/namespace": "off",
+    "import/named": "off",
```
dai-shi, line 35: curious what was the error.

package.json
```diff
@@ -58,10 +58,10 @@
     "eslint": "eslint --fix '*.{js,json}' '{src,tests}/**/*.{ts,tsx}'",
     "eslint:ci": "eslint '*.{js,json}' '{src,tests}/**/*.{ts,tsx}'",
     "pretest": "tsc --noEmit",
-    "test": "yarn node --experimental-vm-modules $(yarn bin jest)",
-    "test:ci": "yarn node --experimental-vm-modules $(yarn bin jest)",
-    "test:dev": "yarn node --experimental-vm-modules $(yarn bin jest) --watch --no-coverage",
-    "test:coverage:watch": "yarn node --experimental-vm-modules $(yarn bin jest) --watch",
+    "test": "vitest",
```
dai-shi, line 61: So happy to see the experimental thing is gone!

dai-shi (commented):
nice work!
added some comments for a start.

tests/basic.test.tsx
Aslemammad: Yes, it will avoid. I did it, but it seems to show some testing-library errors `TestingLibraryElementError: Found multiple elements with the text: button`

.eslintrc.json
Aslemammad: it's related to vitest ./utils import! so i solved it with thsi as suggested.

package.json
Aslemammad: Same with me! I've been happy for a while.

.eslintrc.json
Aslemammad: I don't think it'd work with vitest, does it?

tests/basic.test.tsx
Aslemammad: Fixed

dai-shi: Nice!!

tests/computed.test.tsx
dai-shi: I assume it becomes `any`. Can we keep `: number`?

.eslintrc.json
dai-shi: I think it does.

.eslintrc.json
dai-shi: I'll revisit it later.

tests/macro.test.ts
```diff
@@ -1,5 +1,10 @@
 import plugin from 'babel-plugin-macros'
 import BabelPluginTester from 'babel-plugin-tester'
+import { describe, it, expect } from 'vitest'
+
+globalThis.describe = describe
```
dai-shi, line 5: should we add ts-expect-error?

tests/computed.test.tsx
Aslemammad: Yes, I did it!

tests/macro.test.ts
Aslemammad: done

.eslintrc.json
Aslemammad: Did it with eslint-plugin-vitest

dai-shi: 🎉

tests/computed.test.tsx
```diff
@@ -32,7 +25,7 @@ const sleep = (ms: number) =>
 
 describe('proxyWithComputed', () => {
   it('simple computed getters', async () => {
-    const computeDouble = jest.fn((x: number) => x * 2)
+    const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 28:
```suggestion
    const computeDouble = vi.fn((x: number) => x * 2)
```

dai-shi (commented):
So far, it looks good to me.

@arjunvegda Would you like to take a look? (We don't merge this yet.)

tests/computed.test.tsx
dai-shi: @Aslemammad Please do the same for others.

Aslemammad: sure.

arjunvegda:
> So far, it looks good to me.
> 
> @arjunvegda Would you like to take a look? (We don't merge this yet.)

Nice stuff @Aslemammad! 

I'm running into the [same exact issue](https://github.com/pmndrs/jotai/actions/runs/4601683015/jobs/8129835682?pr=1839) on Jotai 😅 . The tests for TS versions `3.9.7` and `3.8.3` are failing.

It appears that those TS versions are ignoring the `skipLibCheck` 😟. I tried a few tricks like `strict: false`, `exclude: ['node_modules/**]`, and adding `// @ts-ignore` but no luck.

I'll continue digging!

tests/computed.test.tsx
```diff
@@ -178,7 +171,7 @@ describe('proxyWithComputed', () => {
 
 describe('DEPRECATED addComputed', () => {
   it('simple addComputed', async () => {
-    const computeDouble = jest.fn((x: number) => x * 2)
+    const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 174:
```suggestion
    const computeDouble = vi.fn((x: number) => x * 2)
```

tests/computed.test.tsx
```diff
@@ -248,7 +241,7 @@ describe('DEPRECATED addComputed', () => {
   })
 
   it('nested emulation with addComputed', async () => {
-    const computeDouble = jest.fn((x: number) => x * 2)
+    const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 244:
```suggestion
    const computeDouble = vi.fn((x: number) => x * 2)
```

tests/derive.test.tsx
```diff
@@ -10,7 +10,7 @@ const sleep = (ms: number) =>
   })
 
 it('basic derive', async () => {
-  const computeDouble = jest.fn((x: number) => x * 2)
+  const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 13:
```suggestion
  const computeDouble = vi.fn((x: number) => x * 2)
```

tests/derive.test.tsx
```diff
@@ -42,7 +42,7 @@ it('basic derive', async () => {
 })
 
 it('derive another proxy', async () => {
-  const computeDouble = jest.fn((x: number) => x * 2)
+  const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 45:
```suggestion
  const computeDouble = vi.fn((x: number) => x * 2)
```

tests/derive.test.tsx
```diff
@@ -80,7 +80,7 @@ it('derive another proxy', async () => {
 })
 
 it('derive with self', async () => {
-  const computeDouble = jest.fn((x: number) => x * 2)
+  const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 83:
```suggestion
  const computeDouble = vi.fn((x: number) => x * 2)
```

tests/derive.test.tsx
```diff
@@ -117,14 +117,14 @@ it('derive with self', async () => {
 })
 
 it('derive with two dependencies', async () => {
-  const computeSum = jest.fn((x: number, y: number) => x + y)
+  const computeSum = vi.fn((x, y) => x + y)
```
dai-shi, line 120:
```suggestion
  const computeSum = vi.fn((x: number, y: number) => x + y)
```

tests/derive.test.tsx
```diff
@@ -190,7 +190,7 @@ it('async derive', async () => {
 })
 
 it('nested emulation with derive', async () => {
-  const computeDouble = jest.fn((x: number) => x * 2)
+  const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 193:
```suggestion
  const computeDouble = vi.fn((x: number) => x * 2)
```

tests/derive.test.tsx
```diff
@@ -255,13 +255,13 @@ it('derive with array.pop', async () => {
 })
 
 it('basic underive', async () => {
-  const computeDouble = jest.fn((x: number) => x * 2)
+  const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 258:
```suggestion
  const computeDouble = vi.fn((x: number) => x * 2)
```

tests/getter.test.tsx
```diff
@@ -1,10 +1,10 @@
 import { StrictMode } from 'react'
-import { expect, it, jest } from '@jest/globals'
 import { fireEvent, render, waitFor } from '@testing-library/react'
+import { expect, it, vi } from 'vitest'
 import { proxy, useSnapshot } from 'valtio'
 
 it('simple object getters', async () => {
-  const computeDouble = jest.fn((x: number) => x * 2)
+  const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 7:
```suggestion
  const computeDouble = vi.fn((x: number) => x * 2)
```

tests/getter.test.tsx
```diff
@@ -46,7 +46,7 @@ it('simple object getters', async () => {
 })
 
 it('object getters returning object', async () => {
-  const computeDouble = jest.fn((x: number) => x * 2)
+  const computeDouble = vi.fn((x) => x * 2)
```
dai-shi, line 49:
```suggestion
  const computeDouble = vi.fn((x: number) => x * 2)
```

dai-shi (changes requested):
still remaining.

Aslemammad:
@arjunvegda Thank you so much, yea, I didn't find any solution neither.

@dai-shi I didn't have access to my main computer that's why I didn't type those functions :)

dai-shi:
> I didn't have access to my main computer

No need to rush. Take your time.
When you get a chance to access to your computer, please merge `main`.

arjunvegda:
> @arjunvegda Thank you so much, yea, I didn't find any solution neither.

I think I managed to make [it work](https://github.com/pmndrs/jotai/actions/runs/4663276091/jobs/8254459966) 😅. I did the following

1. Use the older versions of `@testing-library/user-event` and `@testing-library/react` that are compatible with TS 3.9.7 and 3.9.8. See [jotai/commits/b73e4310803315defebc124f90d0f9a9aaef4376](https://github.com/pmndrs/jotai/pull/1839/commits/b73e4310803315defebc124f90d0f9a9aaef4376)
2. Remove types for `vitest` as there is no way to have TS skip them (probably a bug in 3.9.x ?). See  [jotai/commits/b73e4310803315defebc124f90d0f9a9aaef4376](https://github.com/pmndrs/jotai/pull/1839/commits/b73e4310803315defebc124f90d0f9a9aaef4376)
3. Declare an empty module for `vitest` within the `types.d.ts`. See [jotai/commits/0c628591c05476fc3820e737ca160b1bd719ca08](https://github.com/pmndrs/jotai/pull/1839/commits/0c628591c05476fc3820e737ca160b1bd719ca08)
4. Keep `parse5` module, as vitest relies on it. See [jotai/commits/d0edaf6891450ad4a8eec0e69c74b358f930d724](https://github.com/pmndrs/jotai/pull/1839/commits/d0edaf6891450ad4a8eec0e69c74b358f930d724)

It's not great, but it does the job for now. Let me know what you think!

Aslemammad:
Nice job! Thank you so much!

@dai-shi Wdyt?

dai-shi:
> Nice job! Thank you so much!
> 
> @dai-shi Wdyt?

I just reviewed https://github.com/pmndrs/jotai/pull/1839. Seems like they are reasonable hacks.

package.json
```diff
@@ -58,10 +58,10 @@
...
+    "test:ci": "vitest",
+    "test:dev": "vitest --no-coverage",
+    "test:coverage:watch": "vitest --coverage --watch",
```
dai-shi, line 64: Let's simplify this. https://github.com/pmndrs/jotai/pull/1839#discussion_r1166148724
```suggestion
    "test": "vitest --ui --coverage",
    "test:ci": "vitest",
```

tests/watch.test.tsx
```diff
@@ -1,12 +1,12 @@
-import { describe, expect, it, jest } from '@jest/globals'
+import { describe, expect, it, vi } from 'vitest'
+import { proxy } from 'valtio'
 import { watch } from 'valtio/utils'
-import { proxy } from 'valtio/vanilla'
```
dai-shi, line 3: Does this mean we are failing to alias `valtio/vanilla`?

vitest.config.ts
```diff
@@ -0,0 +1,24 @@
...
+    },
+    environment: 'jsdom',
+    dir: 'tests',
+    alias: [
+      {
+        find: /^valtio$/,
+        replacement: './src/index.ts',
+      },
+      {
+        find: /^valtio(.*)$/,
```
dai-shi, line 19: Isn't it like this?
```suggestion
        find: /^valtio\/(.*)$/,
```

dai-shi (commented):
Overall looks good.

tests/watch.test.tsx
Aslemammad: Yes, it does not work with the alias trick. @arjunvegda You may be able to help.

arjunvegda: That's an interesting one! 😅 Based on the [type of `alias`](https://github.com/vitejs/vite/blob/21c61cbbbe1add738f3fc4e8e80de2c73b1fc5a8/packages/vite/src/types/alias.d.ts#L52-L60), it appears to be using rollup to resolve the paths, so would this do the trick? 🤔 


```diff
// vitest.config.ts
+import path from 'path'
import { defineConfig } from 'vitest/config'

export default defineConfig({
  test: {
   name: 'valtio',
    setupFiles: './tests/setup.ts',
    coverage: {
      reporter: ['text', 'json', 'html', 'text-summary'],
      reportsDirectory: './coverage/',
    },
    environment: 'jsdom',
    dir: 'tests',
    alias: [
      {
        find: /^valtio$/,
-      replacement: './src/index.ts', 
+      replacement: path.resolve(__dirname, './src/index.ts'),
      },
      {
        find: /^valtio\/(.*)$/,
-       replacement: './src/$1.ts',
+       replacement: path.resolve(__dirname, './src/$1.ts'),
      },
    ],
  },
})

```

Aslemammad: I guess this did the trick, thank you so much.

dai-shi: hmm, but it worked for `valtio`, doesn't it? anyway, feel free to give a try.

Aslemammad: Yes, it's working. pushed a new change.

vitest.config.ts
```diff
@@ -0,0 +1,25 @@
...
+    environment: 'jsdom',
+    dir: 'tests',
+    alias: [
+      {
+        find: /^valtio$/,
+        replacement: './src/index.ts'
+      },
+      {
+        find: /^valtio\/(.*)$/,
+        replacement: path.resolve(__dirname, './src/$1.ts')
```
dai-shi, line 21: Can I just give a try with reverting this?
```suggestion
        replacement: './src/$1.ts'
```

If that doesn't work let's revert.

dai-shi (approved):
I don't know why `__dirname` is necessary, and it seems weird, but can't be helped.

dai-shi:
@arjunvegda Could you have a closer look to find why 21db220 doesn't work? what's the difference from jotai and zustand...

dai-shi:
Hmm? It looks like it's loading two instances. 🤔 

image

dai-shi:
Never mind. The last commit fixes the test, and it's probably because it somehow points to a different file.

arjunvegda:
> Hmm? It looks like it's loading two instances. 🤔

Ah! This explains it. if I change the import in [utils/watch.ts#L1](https://github.com/pmndrs/valtio/blob/main/src/vanilla/utils/watch.ts#L1) to `valtio/vanilla` it works as expected 🤔 

It appears that vite is resolving modules differently based on how we import the paths (relative vs absolute)?

```diff
- import { subscribe } from '../../vanilla.ts'
+ import { subscribe } from 'valtio/vanilla'
```

arjunvegda:
> @arjunvegda Could you have a closer look to find why [21db220](https://github.com/pmndrs/valtio/commit/21db220b7952787f660e76c06456ea951d338d6e) doesn't work? what's the difference from jotai and zustand...

Jotai and Zustand do not appear to have conflicting imports within test files, so it is working fine. However, we could reproduce the issue in those libs too. 

This will cause modules to be duplicated and the test will fail.

```ts
import { expect, it } from 'vitest'
import { getDefaultStore } from 'jotai/vanilla/store'
import { getDefaultStore as getDefaultStoreFromVanilla } from 'jotai/vanilla'

it('should match the default store', () => {
  const storeA = getDefaultStoreFromVanilla()
  const storeB = getDefaultStore()
  expect(storeA).toBe(storeB)
})
```

This could be fixed by deduping imports using [@rollup/plugin-node-resolve](https://www.npmjs.com/package/@rollup/plugin-node-resolve) in `vite.config.ts` though

dai-shi:
@arjunvegda Thanks for investigating. Things are clear now. I think for now what we have is fine.

sheremet-va:
> @arjunvegda Thanks for investigating. Things are clear now. I think for now what we have is fine.

This happens because relative paths in aliases are used. Vitest doesn't resolve them relative to root. Each time "valtio" is mentioned it will try to load it relative to the same file and fail. But Vitest fallbacks to the original replacement. This creates two different instances in the modules graph. The fix is to **never** use relative paths in aliases.

The code you have here will probably break in the next Vitest release where we fixed this.

Aslemammad:
> The code you have here will probably break in the next Vitest release where we fixed this.
@sheremet-va Thank you for the explanation. We'll try removing it in the next release of Vitest. cc @dai-shi

dai-shi:
So, always to use `__dirname`? That's also clear (as it's consistent).

sheremet-va:
> So, always to use `__dirname`? That's also clear (as it's consistent).

Yes. Or `import.meta.url`.

PR #647, opened 2023-01-31
docs: Doc updates on classes.

stephenh:
Just adding things that personally I have learned/think is useful to know so far.

## Check List

### DIFF
--- a/docs/api/advanced/snapshot.mdx
+++ b/docs/api/advanced/snapshot.mdx
@@ -21,13 +21,14 @@ import { proxy, snapshot } from 'valtio'
 const store = proxy({ name: 'Mika' })
 const snap1 = snapshot(store) // an efficient copy of the current store values, unproxied
 const snap2 = snapshot(store)
-console.log(snap1 === snap2) // true
+console.log(snap1 === snap2) // true, no need to re-render
+
 store.name = 'Hanna'
 const snap3 = snapshot(store)
-console.log(snap1 === snap3) // false
+console.log(snap1 === snap3) // false, should re-render
 ```
 
-#### Copy on Write
+## Copy on Write
 
 Even though snapshots are a deep copy of the entire state, they use a lazy copy-on-write mechanism for updates, so in practice are quick to maintain.
 
@@ -67,7 +68,7 @@ console.log(s1.books[1] === s2.books[1]) // false
 
 Even though this example only reused one of the four existing snapshot instances, it shows that the cost of maintaining snapshots is based on the _depth_ of your state tree (which is typically low, like author to book to book reviews is three levels), and not the _breadth_ (1000s of books).
 
-#### Classes
+## Classes
 
 Snapshots main the original objects' prototypes, so methods and getters work, and correctly evaluate against the snapshot's frozen state.
 
@@ -100,7 +101,7 @@ Note that the results of getters and methods are not cached, and are re-evaluate
 
 This should be fine, because the expectation is that they execute very quickly (faster than the overhead of caching them would be worth) and are also deterministic, so the return value is based only on the already-frozen snapshot state. That said, if you do need to cache a getter's particularly expensive calculation, you can use <a href="/docs/api/utils/proxyWithComputed">`proxyWithComputed`</a>.
 
-#### Vanilla JavaScript
+## Vanilla JavaScript
 
 In VanillaJS, `snapshot` is not necessary to access proxied object values, inside or outside of subscribe. However, it is useful, for example, to keep a serializable list of unproxied objects or check if objects have changed. It also resolves promises.
 
--- a/docs/api/advanced/snapshot.mdx
+++ b/docs/api/advanced/snapshot.mdx
@@ -9,18 +9,20 @@ description: 'Create a snapshot of current state'
 
 snapshot takes a proxy and returns an immutable object, unwrapped from the proxy.
 
-Immutability is achieved by deeply copying & freezing the object. In sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
+Immutability is achieved by _efficiently_ deep copying & freezing the object (see the "Copy on Write" section for details).
+
+Briefly, in sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same previous snapshot object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
 
 ```js
 import { proxy, snapshot } from 'valtio'
 
 const store = proxy({ name: 'Mika' })
-const unwrappedStore = snapshot(store) // unwrapped store is a copy of the original object, unproxied
-const anotherUnwrappedStore = snapshot(store)
-console.log(unwrappedStore === anotherUnwrappedStore) // true
+const snap1 = snapshot(store) // an efficient copy of the current store values, unproxied
+const snap2 = snapshot(store)
+console.log(snap1 === snap2) // true
 store.name = 'Hanna'
-const yetAnotherUnwrappedStore = snapshot(store)
-console.log(unwrappedStore === yetAnotherUnwrappedStore) // false
+const snap3 = snapshot(store)
+console.log(snap1 === snap3) // false
 ```
 
 #### Copy on Write
--- a/docs/api/advanced/snapshot.mdx
+++ b/docs/api/advanced/snapshot.mdx
@@ -33,10 +33,10 @@ For example, if we have a nested object of:
 const author = proxy({
   firstName: 'f',
   lastName: 'l',
-  books: [{ title: 't1' }, { title: 't2' }]
+  books: [{ title: 't1' }, { title: 't2' }],
 })
 
-const s1 = snapshot(author);
+const s1 = snapshot(author)
 ```
 
 The first `snapshot` call creates four new instances:
@@ -48,20 +48,20 @@ The first `snapshot` call creates four new instances:
 When we mutate the 2nd book:
 
 ```js
-author.books[1].title = 't2b';
-const s2 = snapshot(author);
+author.books[1].title = 't2b'
+const s2 = snapshot(author)
 ```
 
 Then the next `s2` will have a new copy of the 2nd book, but reuse the snapshot of the 1st book.
 
 ```js
-console.log(s1 === s2); // false
-console.log(s1.books === s2.books); // false
-console.log(s1.books[0] === s2.books[0]); // true
-console.log(s1.books[1] === s2.books[1]); // false
+console.log(s1 === s2) // false
+console.log(s1.books === s2.books) // false
+console.log(s1.books[0] === s2.books[0]) // true
+console.log(s1.books[1] === s2.books[1]) // false
 ```
 
-Even though this example only reused one of the four existing snapshot instances, it shows that the cost of maintaining snapshots is based on the *depth* of your state tree (which is typically low, like author to book to reviews is three levels), and not the *breadth* (1000s of books).
+Even though this example only reused one of the four existing snapshot instances, it shows that the cost of maintaining snapshots is based on the _depth_ of your state tree (which is typically low, like author to book to reviews is three levels), and not the _breadth_ (1000s of books).
 
 #### Classes
 
@@ -78,16 +78,16 @@ class Author {
   }
 }
 
-const state = proxy(new Author());
-const snap = snapshot(state);
+const state = proxy(new Author())
+const snap = snapshot(state)
 
 // the snapshot has the Author prototype
-console.log(snap instanceof Author); // true
+console.log(snap instanceof Author) // true
 
-state.firstName = 'f2';
+state.firstName = 'f2'
 
 // Invocations use the snapshot's state, still 'f'
-console.log(snap.fullName()); // 'f l'
+console.log(snap.fullName()) // 'f l'
 ```
 
 #### Vanilla javascript
--- a/docs/api/advanced/snapshot.mdx
+++ b/docs/api/advanced/snapshot.mdx
@@ -7,20 +7,87 @@ description: 'Create a snapshot of current state'
 
 # snapshot
 
-#### A snapshot takes a proxy and returns a immutable object, unwrapped from the proxy.
+snapshot takes a proxy and returns an immutable object, unwrapped from the proxy.
 
-Immutability is achieved by deeply freezing the object. In sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
+Immutability is achieved by deeply copying & freezing the object. In sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
 
 ```js
 import { proxy, snapshot } from 'valtio'
 
 const store = proxy({ name: 'Mika' })
-const unwrappedStore = snapshot(store) // unwrapped store is the original object, unproxied
+const unwrappedStore = snapshot(store) // unwrapped store is a copy of the original object, unproxied
 const anotherUnwrappedStore = snapshot(store)
 console.log(unwrappedStore === anotherUnwrappedStore) // true
 store.name = 'Hanna'
 const yetAnotherUnwrappedStore = snapshot(store)
 console.log(unwrappedStore === yetAnotherUnwrappedStore) // false
+```
+
+#### Copy on Write
+
+Even though snapshots are a deep copy of the entire state, they use a lazy copy-on-write mechanism so in practice are quick to maintain.
+
+For example, if we have a nested object of:
+
+```js
+const author = proxy({
+  firstName: 'f',
+  lastName: 'l',
+  books: [{ title: 't1' }, { title: 't2' }]
+})
+
+const s1 = snapshot(author);
+```
+
+The first `snapshot` call creates four new instances:
+
+- one for the author,
+- one for the books array, and
+- two for the book objects.
+
+When we mutate the 2nd book:
+
+```js
+author.books[1].title = 't2b';
+const s2 = snapshot(author);
+```
+
+Then the next `s2` will have a new copy of the 2nd book, but reuse the snapshot of the 1st book.
+
+```js
+console.log(s1 === s2); // false
+console.log(s1.books === s2.books); // false
+console.log(s1.books[0] === s2.books[0]); // true
+console.log(s1.books[1] === s2.books[1]); // false
+```
+
+Even though this example only reused one of the four existing snapshot instances, it shows that the cost of maintaining snapshots is based on the *depth* of your state tree (which is typically low, like author to book to reviews is three levels), and not the *breadth* (1000s of books).
+
+#### Classes
+
+Snapshots main the original objects' prototypes, so methods and getters work, and correctly evaluate against the snapshot's frozen state.
+
+```js
+import { proxy, snapshot } from 'valtio'
+
+class Author {
+  firstName = 'f'
+  lastName = 'l'
+  fullName() {
+    return `${this.firstName} ${this.lastName}`
+  }
+}
+
+const state = proxy(new Author());
+const snap = snapshot(state);
+
+// the snapshot has the Author prototype
+console.log(snap instanceof Author); // true
+
+state.firstName = 'f2';
+
+// Invocations use the snapshot's state, still 'f'
+console.log(snap.fullName()); // 'f l'
 ```
 
 #### Vanilla javascript
--- a/docs/api/advanced/snapshot.mdx
+++ b/docs/api/advanced/snapshot.mdx
@@ -11,7 +11,9 @@ snapshot takes a proxy and returns an immutable object, unwrapped from the proxy
 
 Immutability is achieved by _efficiently_ deep copying & freezing the object (see the "Copy on Write" section for details).
 
-Briefly, in sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same previous snapshot object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
+Briefly, in sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same previous snapshot object is returned. This allows for shallow comparison in render functions, preventing spurious renders.
+
+Snapshots also throw promises, making them work with React Suspense.
 
 ```js
 import { proxy, snapshot } from 'valtio'
@@ -27,7 +29,7 @@ console.log(snap1 === snap3) // false
 
 #### Copy on Write
 
-Even though snapshots are a deep copy of the entire state, they use a lazy copy-on-write mechanism so in practice are quick to maintain.
+Even though snapshots are a deep copy of the entire state, they use a lazy copy-on-write mechanism for updates, so in practice are quick to maintain.
 
 For example, if we have a nested object of:
 
@@ -47,14 +49,14 @@ The first `snapshot` call creates four new instances:
 - one for the books array, and
 - two for the book objects.
 
-When we mutate the 2nd book:
+When we mutate the 2nd book, and take a new `snapshot`:
 
 ```js
 author.books[1].title = 't2b'
 const s2 = snapshot(author)
 ```
 
-Then the next `s2` will have a new copy of the 2nd book, but reuse the snapshot of the 1st book.
+Then `s2` will have a new copy of the 2nd book, but reuse the existing snapshot of the unchanged 1st book.
 
 ```js
 console.log(s1 === s2) // false
@@ -63,7 +65,7 @@ console.log(s1.books[0] === s2.books[0]) // true
 console.log(s1.books[1] === s2.books[1]) // false
 ```
 
-Even though this example only reused one of the four existing snapshot instances, it shows that the cost of maintaining snapshots is based on the _depth_ of your state tree (which is typically low, like author to book to reviews is three levels), and not the _breadth_ (1000s of books).
+Even though this example only reused one of the four existing snapshot instances, it shows that the cost of maintaining snapshots is based on the _depth_ of your state tree (which is typically low, like author to book to book reviews is three levels), and not the _breadth_ (1000s of books).
 
 #### Classes
 
@@ -98,7 +100,7 @@ Note that the results of getters and methods are not cached, and are re-evaluate
 
 This should be fine, because the expectation is that they execute very quickly (faster than the overhead of caching them would be worth) and are also deterministic, so the return value is based only on the already-frozen snapshot state. That said, if you do need to cache a getter's particularly expensive calculation, you can use <a href="/docs/api/utils/proxyWithComputed">`proxyWithComputed`</a>.
 
-#### Vanilla javascript
+#### Vanilla JavaScript
 
 In VanillaJS, `snapshot` is not necessary to access proxied object values, inside or outside of subscribe. However, it is useful, for example, to keep a serializable list of unproxied objects or check if objects have changed. It also resolves promises.
 
--- a/docs/api/advanced/snapshot.mdx
+++ b/docs/api/advanced/snapshot.mdx
@@ -86,7 +86,9 @@ console.log(snap instanceof Author) // true
 
 state.firstName = 'f2'
 
-// Invocations use the snapshot's state, still 'f'
+// Invocations use the snapshot's state, e.g. this is still 'f' because
+// inside `fullName`, `this` will be the frozen snapshot instance and not
+// the mutable state proxy
 console.log(snap.fullName()) // 'f l'
 ```
 
--- a/docs/api/advanced/snapshot.mdx
+++ b/docs/api/advanced/snapshot.mdx
@@ -70,7 +70,7 @@ Even though this example only reused one of the four existing snapshot instances
 
 ## Classes
 
-Snapshots main the original objects' prototypes, so methods and getters work, and correctly evaluate against the snapshot's frozen state.
+Snapshots maintain the original objects' prototypes, so methods and getters work, and correctly evaluate against the snapshot's frozen state.
 
 ```js
 import { proxy, snapshot } from 'valtio'
@@ -99,7 +99,7 @@ console.log(snap.fullName()) // 'f l'
 
 Note that the results of getters and methods are not cached, and are re-evaluated on every call.
 
-This should be fine, because the expectation is that they execute very quickly (faster than the overhead of caching them would be worth) and are also deterministic, so the return value is based only on the already-frozen snapshot state. That said, if you do need to cache a getter's particularly expensive calculation, you can use <a href="/docs/api/utils/proxyWithComputed">`proxyWithComputed`</a>.
+This should be fine, because the expectation is that they execute very quickly (faster than the overhead of caching them would be worth) and are also deterministic, so the return value is based only on the already-frozen snapshot state.
 
 ## Vanilla JavaScript
 
--- a/docs/api/advanced/snapshot.mdx
+++ b/docs/api/advanced/snapshot.mdx
@@ -92,6 +92,10 @@ state.firstName = 'f2'
 console.log(snap.fullName()) // 'f l'
 ```
 
+Note that the results of getters and methods are not cached, and are re-evaluated on every call.
+
+This should be fine, because the expectation is that they execute very quickly (faster than the overhead of caching them would be worth) and are also deterministic, so the return value is based only on the already-frozen snapshot state. That said, if you do need to cache a getter's particularly expensive calculation, you can use <a href="/docs/api/utils/proxyWithComputed">`proxyWithComputed`</a>.
+
 #### Vanilla javascript
 
 In VanillaJS, `snapshot` is not necessary to access proxied object values, inside or outside of subscribe. However, it is useful, for example, to keep a serializable list of unproxied objects or check if objects have changed. It also resolves promises.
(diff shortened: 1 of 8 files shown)


docs/api/advanced/snapshot.mdx
```diff
@@ -7,22 +7,89 @@ description: 'Create a snapshot of current state'
 
 # snapshot
 
-#### A snapshot takes a proxy and returns a immutable object, unwrapped from the proxy.
```
stephenh, line 10: The sentence seemed really long for a `####` header, so just removed the header notation...

docs/api/advanced/snapshot.mdx
```diff
@@ -7,22 +7,89 @@ description: 'Create a snapshot of current state'
...
 
-Immutability is achieved by deeply freezing the object. In sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
+Immutability is achieved by deeply copying & freezing the object. In sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
 
 ```js
 import { proxy, snapshot } from 'valtio'
 
 const store = proxy({ name: 'Mika' })
-const unwrappedStore = snapshot(store) // unwrapped store is the original object, unproxied
+const unwrappedStore = snapshot(store) // unwrapped store is a copy of the original object, unproxied
```
stephenh, line 18: Pedantically the snapshot is not literally the `{ name: 'Mika' }` instance passed to `proxy`.

docs/api/advanced/snapshot.mdx
```diff
@@ -7,22 +7,89 @@ description: 'Create a snapshot of current state'
...
 const anotherUnwrappedStore = snapshot(store)
 console.log(unwrappedStore === anotherUnwrappedStore) // true
 store.name = 'Hanna'
 const yetAnotherUnwrappedStore = snapshot(store)
 console.log(unwrappedStore === yetAnotherUnwrappedStore) // false
 ```
 
+#### Copy on Write
```
stephenh, line 26: I realize that line 12 is already saying the same thing, but dunno, personally it took me awhile to understand / get it, so feel this is useful. Happy to drop/edit.

docs/api/advanced/subscribe.mdx
```diff
@@ -35,6 +35,12 @@ subscribe(state.arr, () => console.log('state.arr has changed to', state.arr))
 state.arr.push('world')
 ```
 
-## Codesandbox demo in VanillaJS
+#### no usage tracking
+
```
stephenh, line 39: Basically a copy/paste of the disclaimer I added to `watch` in the last PR

docs/api/advanced/subscribe.mdx
```diff
@@ -35,6 +35,12 @@ subscribe(state.arr, () => console.log('state.arr has changed to', state.arr))
...
+`subscribe` currently does not implement the fine-grained usage tracking of `useSnapshot`, so the callback will rerun anytime the subscribed proxy (or one of its child proxy) has any fields mutated, regardless of the fields accessed within the `callback`'s code.
+
+Basically, valtio's subscriptions are fired on each new version of a proxy, which aligns to exactly when new `snapshot`s need to be taken, which is separate from whether specific views on a given snapshot need to re-render.
```
stephenh, line 42: Not sure if this page is the best place for it, but this subscriptions --> on new version --> snapshots need to know when a new version happens has a nice "aha" for me, about why valtio only does fine-grained / usage-based logic "at the very end" against snapshots and not sooner in the subscription lifecycle.

docs/api/basic/useSnapshot.mdx
```diff
@@ -8,7 +8,7 @@ description: 'Create a local snapshot that catches changes.'
 # useSnapshot
 
 Create a local snapshot that catches changes. This hook actually returns a wrapped snapshot in a proxy for
-render optimization instead of a plain object, which is what <a href="/docs/advanced/snapshot">`snapshot()`</a> returns.
+render optimization instead of a plain object, which is what <a href="/docs/api/advanced/snapshot">`snapshot()`</a> returns.
```
stephenh, line 11: Fixed a link

docs/api/utils/proxyWithComputed.mdx
```diff
@@ -7,36 +7,24 @@ description: ''
 
 # proxyWithComputed
 
-#### add a computation to a proxy
+#### add cached computation to a proxy
```
stephenh, line 10: I know `proxyWithComputed` doesn't _have_ to be used for caching, but that seems like the primary use case, so snuck this in...happy to drop.

docs/api/utils/proxyWithComputed.mdx
```diff
@@ -48,16 +36,60 @@ const state2 = proxyWithComputed(
...
 
-The last use case fails to infer types in TypeScript
-[#192](https://github.com/pmndrs/valtio/issues/192).
+The last use case fails to infer types in TypeScript, see [#192](https://github.com/pmndrs/valtio/issues/192).
+
+#### Disclaimer on `proxy-memoize`
+
+Be careful not to overuse `proxy-memoize` because `proxy-memoize` and `useSnapshot` do similar optimization and double optimization may lead to less performance.
+
+The similarity is that both `proxy-memoize` and `useSnapshot` keep caches of data; the difference is that getters invoked against `snapshot`s or `useSnapshot`s are not cached and will re-invoke the getter/re-calculate the result each time they are accessed; whereas `proxy-memoize` will use the `snap` argument to calculate the getter result only once per snapshot. )
```
stephenh, line 54: Fwiw this is my understanding of things, happy to update.

docs/api/advanced/snapshot.mdx
```diff
@@ -7,22 +7,95 @@ description: 'Create a snapshot of current state'
 
 # snapshot
 
-#### A snapshot takes a proxy and returns a immutable object, unwrapped from the proxy.
+snapshot takes a proxy and returns an immutable object, unwrapped from the proxy.
 
-Immutability is achieved by deeply freezing the object. In sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
+Immutability is achieved by deeply copying & freezing the object. In sequential snapshot calls, when the values in the proxy have not changed, a pointer to the same object is returned. This allows for shallow comparison in render functions, preventing spurious renders. Snapshots also throw promises, making them work with React Suspense.
```
dai-shi, line 12: we want to emphasize that the copying is optimized and most efficient.

docs/api/advanced/snapshot.mdx
dai-shi: What you pass to `proxy()` is "initializing value" and it's different from `store` too.
Again, `snapshot()` return an immutable object which is different from `store`, but it's super optimized copying. I'm afraid by simply saying `a copy of`, it feels like `snapshot()` does deep copying on each call.

docs/api/advanced/snapshot.mdx
```diff
@@ -7,22 +7,95 @@ description: 'Create a snapshot of current state'
...
 const anotherUnwrappedStore = snapshot(store)
 console.log(unwrappedStore === anotherUnwrappedStore) // true
 store.name = 'Hanna'
 const yetAnotherUnwrappedStore = snapshot(store)
 console.log(unwrappedStore === yetAnotherUnwrappedStore) // false
 ```
 
+#### Copy on Write
+
+Even though snapshots are a deep copy of the entire state, they use a lazy copy-on-write mechanism so in practice are quick to maintain.
```
dai-shi, line 28: yay!

docs/api/advanced/snapshot.mdx
```diff
@@ -7,22 +7,95 @@ description: 'Create a snapshot of current state'
...
+
+```js
+console.log(s1 === s2) // false
+console.log(s1.books === s2.books) // false
+console.log(s1.books[0] === s2.books[0]) // true
+console.log(s1.books[1] === s2.books[1]) // false
+```
+
+Even though this example only reused one of the four existing snapshot instances, it shows that the cost of maintaining snapshots is based on the _depth_ of your state tree (which is typically low, like author to book to reviews is three levels), and not the _breadth_ (1000s of books).
+
```
dai-shi, line 65: ☝️ very nice example.

docs/api/advanced/snapshot.mdx
```diff
@@ -7,22 +7,95 @@ description: 'Create a snapshot of current state'
...
+  fullName() {
+    return `${this.firstName} ${this.lastName}`
+  }
+}
+
+const state = proxy(new Author())
+const snap = snapshot(state)
+
+// the snapshot has the Author prototype
+console.log(snap instanceof Author) // true
```
dai-shi, line 85: I didn't know `instanceof` works like that. 😄

docs/api/advanced/snapshot.mdx
```diff
@@ -7,22 +7,95 @@ description: 'Create a snapshot of current state'
...
+// Invocations use the snapshot's state, e.g. this is still 'f' because
+// inside `fullName`, `this` will be the frozen snapshot instance and not
+// the mutable state proxy
+console.log(snap.fullName()) // 'f l'
+```
+
+Note that the results of getters and methods are not cached, and are re-evaluated on every call.
+
+This should be fine, because the expectation is that they execute very quickly (faster than the overhead of caching them would be worth) and are also deterministic, so the return value is based only on the already-frozen snapshot state. That said, if you do need to cache a getter's particularly expensive calculation, you can use <a href="/docs/api/utils/proxyWithComputed">`proxyWithComputed`</a>.
+
```
dai-shi, line 98: ☝️ well written. 👏

docs/api/advanced/subscribe.mdx
dai-shi: it's a bit wordy here, because no usage tracking in subscribe is by design.

docs/api/advanced/subscribe.mdx
```diff
@@ -35,6 +35,12 @@ subscribe(state.arr, () => console.log('state.arr has changed to', state.arr))
 state.arr.push('world')
 ```
 
-## Codesandbox demo in VanillaJS
+#### no usage tracking
+
+`subscribe` currently does not implement the fine-grained usage tracking of `useSnapshot`, so the callback will rerun anytime the subscribed proxy (or one of its child proxy) has any fields mutated, regardless of the fields accessed within the `callback`'s code.
```
dai-shi, line 40: such that it will _never_ be implemented.

docs/api/advanced/subscribe.mdx
dai-shi: yeah, this is a nice description and we should have it somewhere, here or before.

docs/api/basic/useSnapshot.mdx
```diff
@@ -90,6 +90,15 @@ const { profile } = useSnapshot(state)
 return <div>{profile.name}</div>
 ```
 
+#### Dev Mode Debug Values
+
+In dev mode, `useSnapshot` uses React's `useDebugValue` to output a list of fields that were accessed during rendering, i.e. which specific fields will trigger re-render when the tracking proxy changes.
```
dai-shi, line 95: Thanks. We should have added this section.
It would be super nice if we add some screenshots.

docs/api/hacks/getVersion.mdx
```diff
@@ -7,7 +7,13 @@ description: ''
 
 # getVersion
 
-In Valtio, updates to proxied objects are tracked internally with a version number. Therefore `getVersion` helper can be used to check if a proxied object has been updated. This is not typically useful or recommended to use in application code.
+In Valtio, updates to proxied objects are tracked internally with a version number.
+
+Every mutation to a proxy increases a global version number, and assigns the just mutated proxy, and any parent proxies (which automatically subscribe to their child proxies), to the latest version number.
```
dai-shi, line 12: fwiw, the reason we use the global number is to detect loop, that support circular structure. If we don't need to support circular structure, local version number just works.

docs/api/utils/proxyWithComputed.mdx
dai-shi: For a historical reason, we show the usage with proxy-memoize.
But, the primary use case is without proxy-memoize, so no caching.
proxy-memoize should be used if caching improves performance (which can be rare).

docs/api/utils/proxyWithComputed.mdx
dai-shi: The similarity is usage tracking (= proxy-compare). caching is kind of by-product.
I'm not sure what you are trying say for the difference.

docs/api/utils/proxyWithComputed.mdx
```diff
@@ -48,16 +36,60 @@ const state2 = proxyWithComputed(
...
+
+#### Compared to Classes
+
+Using `proxyWithComputed` is similar in spirit to using classes, e.g. the above `proxyWithComputed` could be written as:
```
dai-shi, line 58: I'm not sure if I agree. Feels misleading. I would remove this section.

docs/api/utils/proxyWithComputed.mdx
```diff
@@ -48,16 +36,60 @@ const state2 = proxyWithComputed(
...
+const state = proxy(new Author())
+```
+
+The major difference is that:
+
+- `proxyWithComputed` allows using `memoize` to cache getters with course-grained dependency tracking
+
+  "Course-grained" insofar as the getter will be recalculated whenever any field in the snapshot changes, i.e. it has a version bump due to any mutation. This in contract with `useSnapshot`, which implements "fine-grained" dependency tracking based on the fields specifically accessed within the JSX.
+
+#### Compared to object getters
```
dai-shi, line 80: Same here.

I would like to have a comparison between classes and object getters.

docs/api/utils/watch.mdx
```diff
@@ -78,8 +78,8 @@ If you remove the setTimeout in the example above, 'Juuso's session is expired'
 
 #### no usage tracking
 
-`watch` currently does not implement the usage tracking of `useSnapshot`, so the callback will rerun anytime a watched proxy (or child proxy) has any of its keys mutated, regardless of the keys accessed within the `callback`'s code.
+`watch` currently does not implement the usage tracking of `useSnapshot`, so the callback will rerun anytime a watched proxy (or child proxy) has any fields mutated, regardless of the fields accessed within the `callback`'s code.
 
-And the return value of `sub` is just the proxy itself, not a snapshot.
+And the return value of `add` is just the proxy itself, not a snapshot.
```
dai-shi, line 83: it's `get`. 😄

dai-shi (commented):
Thanks for your contribution!
Here's the first review!

docs/api/advanced/snapshot.mdx
stephenh: > What you pass to proxy() is "initializing value" and it's different from store too.

Right, but imo the comment as-is, "unwrapped store is the original object" to me insinuated it is literally the same instance as the initializing value, which it is not.

I've edited a bit (will push the edit soon...) and thrown in an "efficient" to see if that helps? :-)

docs/api/advanced/subscribe.mdx
stephenh: Agreed I like this description but feel this isn't 100% the best place for it; leaving it here for now...

docs/api/utils/proxyWithComputed.mdx
stephenh: Hm, thanks for pointing this out, b/c I'd assumed this caching pattern was the primary reason to use `proxyWithComputed` over object getters--but if that's not the case, what actually is the primary use case?

And is that a use case that cannot be served by object getters/setters? I know from an implementation perspective, the object getter will be invoked with `this` as the proxy, where as `proxyWithComputed` will have `get` invoked with a snapshot, but I'm struggling to guess a use case / situation where that would matter.

docs/api/advanced/snapshot.mdx
stephenh: Sounds good; added an "efficient" and "see copy on write section later" to hopefully highlight that.

docs/api/advanced/subscribe.mdx
stephenh: Ah yeah, I think I see what you mean: per that other paragraph I added, I understand why valtio's core `addListener` + subscription primitives must never be usage-based, b/c snapshots must be re-created on any field change.

That said, when I wrote this, I was thinking that technically `subscribe`-the-exported-method _could_ learn to do usage tracking, w/o changing the other internals...

Granted, the API would change b/c the callback would need to accept a snapshot.

But, maybe at that point `subscribe` would be too much like a `watch`-that-learned-to-do-usage-tracking...

So, yeah, that makes sense; I'll edit this to remove the insinuation that `subscribe`-the-public-api could be usage based someday.

docs/api/utils/proxyWithComputed.mdx
stephenh: Oh man, I see; I thought `proxy-memoize` was just some generic memoization project from NPM that the examples happened to use.

But I get it now, it is literally a sister project, also built on `proxy-compare`, that does snapshot-ish style lazy recalcs.

That is nifty; will update the docs with that learning. Let me know if the updated description is right.

...fwiw, it seems like maybe `proxyWithComputed` should learn to use "real"/native Valtio snapshots, and pass those into the computed's gets, so that it can do the lazy recalc itself? Just thinking that would solve the "double copies" problem? Granted, it would mean all computeds are cached... (unless there was an option the user passed to turn caching on/off for each getter.) Dunno, just talking out loud.

docs/api/advanced/subscribe.mdx
stephenh: Okay, I rewrote this denote subscribe will always stay non-usage-based, b/c its primary purpose is to drive snapshot creation.

docs/api/hacks/getVersion.mdx
stephenh: Ah! That's cool. I noticed it handles cycle really well.

I probably won't work that tidbit into the docs, but I might make an in-source comment for it... :-)

docs/api/utils/proxyWithComputed.mdx
stephenh: This seems like the biggest thing I'm still struggling with: what are your recommendations of "when to use a class?" vs. "when to use proxyWithComputed?"

I can't tell if the difference is just personal style (some people just don't like classes, which is fine), or if there is something else...

At first, I assumed the big plus of `proxyWithComputed` was that you could use it with `proxy-memoize` to get getter caching, so that's what I wrote up as my best guess.

But since getter caching is not, per your other comment, the primary use case for `proxyWithComputed`, I'm failing to guess/observe what the trade-off would be.

So, dunno, I think that's the shortest way I can ask it: when do you recommend classes, vs. when do you recommend proxyWithComputed?

Then I'll work on rewriting this to reflect that answer. Thank you!

docs/api/advanced/snapshot.mdx
dai-shi: efficient deep copying sounds good!

docs/api/utils/proxyWithComputed.mdx
dai-shi: Good point. OK, the benefit or `proxy-memoize` is understood. Let's see how proxyWithComputed works without `proxy-memoize`.

(btw, I wanted to drop proxyWithComputed when I added derive, but with this unique capability, I left proxyWithComputed. In general, I don't recommend proxyWithComputed, it's very hard to understand.)

Object getters and class getters are just functions. So, let's compare pure functions and proxyWithComputed.

https://codesandbox.io/s/unruffled-cdn-o82gup?file=/src/App.tsx

The thing is, as you see, snapshot works differently.
Let me know if something is unclear.

dai-shi: Nice catch! Yeah, I missed that one. Let's drop proxyWithComputed, and just add a guide in docs how to use object getters (and setters), with optionally `proxy-memoize`.

> // do we need a snapshot(this) call for this to work? not sure...

Yes, we definitely do.

dai-shi: #650

docs/api/utils/watch.mdx
stephenh: Gah! Habit. :-) Fixed!

stephenh:
@dai-shi okay, I've rebased this, I believe addressed all of your feedback, and also made some additions to the new Computed Properties guide. Let me know if I missed any comments, or any other changes you'd like. Thanks!

docs/guides/computed-properties.mdx
```diff
@@ -49,14 +113,20 @@ const memoizedDoubled = memoize((snap) => snap.count * 2)
...
 ## Note about using `this`
 
-You can use `this` instead of `state` in object getters and setters.
-You should be familiar with how JS `this` works,
-as `this` can point to proxy state or snapshot object depending on usage.
+You can use `this` inside of getters and setters, however you should be familiar with how JS `this` works: basically that `this` is whatever object you invoked the call against.
+
+So if you call `state.doubled`, then `this` will be the `state` proxy.
+
+And if you call `snap.doubled`, then `this` will be the snapshot object (except for object getters & setters, where the value is literally copied during the snapshot process, so object getters & setters are never invoked on snapshots, and for them `this` is _always_ the proxy).
```
stephenh, line 130: Fwiw @dai-shi I don't meant to just re-start the debate here, but writing this made me realize that if class getters were cached like object getters, it would let us say that `this` is _always_ the proxy object.

(Because snapshots would never have/invoke getters/setters, as the getter is invoked on the proxy during the snapshot process, and then is just a dumb value in the snapshot.)

Which, dunno, seems like a pretty big simplification for users, and might let us reduce the warnings / "advanced usage only" disclaimers that are currently in the docs?

docs/guides/computed-properties.mdx
```diff
@@ -6,40 +6,104 @@ description: 'Use object getters and setters'
 
 # Computed Properties
 
-Valtio itself doesn't provide a functionality for computed properties,
-but you can emulate it with object getters and setters.
+In Valtio you can use object & class getters and setters to create computed properties.
 
-⚠️ Note: This is a very advanced usage.
+⚠️ Note: This is advanced usage due to the nuances around `this` behavior; see the "Note about this `this`" section below.
```
dai-shi, line 11: I didn't only mean about `this`. Getters are advanced in general.

docs/guides/computed-properties.mdx
```diff
@@ -6,40 +6,104 @@ description: 'Use object getters and setters'
...
 
-## Simple getter
+## Simple object getter
 
 ```js
 const state = proxy({
   count: 1,
   get doubled() {
-    return state.count * 2
+    return this.count * 2
```
dai-shi, line 19: we avoid teaching `this` in the first place. only add a note that they could do it.

dai-shi (commented):
Reviewed it again.

docs/api/advanced/snapshot.mdx
```diff
@@ -7,23 +7,101 @@ description: 'Create a snapshot of current state'
...
+console.log(s1.books[0] === s2.books[0]) // true
+console.log(s1.books[1] === s2.books[1]) // false
 ```
 
-#### Vanilla javascript
+Even though this example only reused one of the four existing snapshot instances, it shows that the cost of maintaining snapshots is based on the _depth_ of your state tree (which is typically low, like author to book to book reviews is three levels), and not the _breadth_ (1000s of books).
+
+## Classes
+
+Snapshots main the original objects' prototypes, so methods and getters work, and correctly evaluate against the snapshot's frozen state.
```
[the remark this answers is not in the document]
stephenh: Ah thanks, I forgot the `tain` of maintain. :-)

docs/api/advanced/snapshot.mdx
```diff
@@ -7,23 +7,101 @@ description: 'Create a snapshot of current state'
...
+
+// Invocations use the snapshot's state, e.g. this is still 'f' because
+// inside `fullName`, `this` will be the frozen snapshot instance and not
+// the mutable state proxy
+console.log(snap.fullName()) // 'f l'
+```
+
+Note that the results of getters and methods are not cached, and are re-evaluated on every call.
+
+This should be fine, because the expectation is that they execute very quickly (faster than the overhead of caching them would be worth) and are also deterministic, so the return value is based only on the already-frozen snapshot state. That said, if you do need to cache a getter's particularly expensive calculation, you can use <a href="/docs/api/utils/proxyWithComputed">`proxyWithComputed`</a>.
```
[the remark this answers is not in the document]
stephenh: Ah right, missed updating this; I'll just drop the last "That said" sentence.

docs/guides/computed-properties.mdx
stephenh: I will restore to unblock the rest of the PR, but getters are advanced? Like getters specifically in Valtio are advanced, or like fundamentally the JS/TS getter language feature itself is advanced?

dai-shi: I believe getters in JS are advanced. IMO, classes (prototypes) are advanced too. Or, read "advanced" not for beginners, instead.

stephenh: Also, even though I do generally disagree (I like getters :-), I do in good faith want to hear & would like to include in the docs your rationale for how/why they are considered advanced.

Otherwise, as a newbie when I read a super-succinct statement like "This is an advanced usage <period>", my mind goes to "great...but why?" or "okay, I guess I believe you, but what specifically is the footgun I need to be careful of?".

And so besides just the direct "this is advanced" statement, I would genuinely appreciate hearing the rationale, so readers can have that insight into the "why" of your thought process, instead of having to take it at face value or guess as to the reasons.

And if it really is "getters are an advanced JS language feature", that's still good, b/c I can include that why/rationale in the docs, and users can then best evaluate if that rationale applies to them/their teams.

Thanks!

docs/guides/computed-properties.mdx
stephenh: I will revert in my next push, but fwiw if I were to read `return state.count` in a getter in basically any code review, I would immediately assume a) it was a bug, and b) it doesn't work.

Granted, I would be wrong (it does work), b/c JS can actually capture the `const state` and eval `state.count` from within `get doubled()`.

But, dunno, I just really want to push/sanity check that in _any_ language (JS, TS, Java, C#, etc), in _any_ framework (mobx, ember), a getter that does _not_ do `return this.whatever` is a bug and so, dunno, this being Valtio's up-front / recommended usage "looks weird" and I'm worried will cause more confusion than the "avoiding `this`" intention hopes to avoid.

Nonetheless, will revert.

docs/guides/computed-properties.mdx
stephenh: (Sorry, didn't see your reply, that sounds good!)

dai-shi: Good point. Thanks for sharing your point.

I think getters as well as proxies are advanced in JS, because it changes the default behavior. People expect plain objects to work like this.
```js
obj.foo = 2;
console.log(obj.foo) // ---> 2
```
getters and proxies can break that.

fwiw, I originally disliked proxies. Now, when I develop libraries with proxies, I pay much attention not to break JS behavior. That's one of the reason, `proxy()` in valtio doesn't have "get" handler.

JS functions are usually invoked with `fn()`. Just accessing a property invoking a function is very counter-intuitive.

In addition, valtio has notion of proxy/snapshot separation, which complicates things. state.foo can be a getter, but snap.foo won't invoke the getter...

So, I think people would need to know the clear understanding of getters and valtio. I recommend beginners to use just plain objects (no getters, no prototypes).

docs/api/advanced/subscribe.mdx
stephenh: Okay, I will remove for now; I'm planning on a writing "how/why valtio works" doc for my next docs PR sometime soon (a little bit of "how is valtio different than mobx"/etc), so I can bring it back as part of that.

docs/guides/computed-properties.mdx
```diff
@@ -6,23 +6,39 @@ description: 'Use object getters and setters'
 
 # Computed Properties
 
-Valtio itself doesn't provide a functionality for computed properties,
-but you can emulate it with object getters and setters.
+In Valtio you can use object & class getters and setters to create computed properties.
 
-⚠️ Note: This is a very advanced usage.
+⚠️ Note: Getters in JavaScript are a more advanced feature of the language, so Valtio recommends using them with caution. That said, if you are a more advanced JavaScript programmer, they should work as you expect; see the "Note about this `this`" section below.
```
stephenh, line 11: @dai-shi Just updated this section + addressed your other comments, should be ready for re-review. Thanks!

docs/guides/computed-properties.mdx
```diff
@@ -6,23 +6,39 @@ description: 'Use object getters and setters'
...
 
-## Simple getter
+## Simple object getter
 
 ```js
 const state = proxy({
   count: 1,
   get doubled() {
+    // We recommend `state.count` for clarity to beginners, but `this.count` works too
```
stephenh, line 19: @dai-shi disclaimer I still kept a `this.count` in a comment :see_no_evil: but will remove if you call me on it. :-)

docs/guides/computed-properties.mdx
stephenh: > Just accessing a property invoking a function is very counter-intuitive.

Yeah, I actually do see what you mean...

I like that they hide whether "is `obj.foo` stored or calculated?" from callers, as that's an implementation detail only I should care about; but admittedly even that is a technical/non-beginner rationale.

I dunno, given getters were added to JavaScript in ES5 / 2009, I think you might be shouting into the wind in terms of taking a principled stance against them. :-)

> I recommend beginners to use just plain objects (no getters, no prototypes

Makes sense.

> fwiw, I originally disliked proxies.

Ha, yeah that is ironic given your, what, 5+ proxy-based libraries.

...maybe you will like getters someday! :-D

dai-shi: well, I personally like how valtio deals object getters. I would use it, especially now that the usaeg is clear without proxyWithComputed.

But, this concept/mental model is unique (the caching is by-product of snapshot). Unless users know getters and valtio in deep, they can't really be sure how to get benefit.

Last but not least, you are far smarter than the average.

docs/guides/computed-properties.mdx
dai-shi: Thanks so much for taking my preference into account. Really appreciated.

docs/guides/computed-properties.mdx
dai-shi: I'm fine with this. Thanks.

docs/guides/computed-properties.mdx
dai-shi: just a minor comment:

> let us say that this is always the proxy object.

it would for getters and setters, but not methods in general.

dai-shi (approved):
Okay, it looks good. Thanks for working on this and discussion with patience. This is huge.

May I merge this, or should wait?

stephenh:
> it would for getters and setters, but not methods in general.

Good point; I've just taken the "always" text out for now. I agree addressing methods would be useful, but probably as a follow up PR.

I kinda wonder if maybe snapshots should just have all methods dropped from the type; that would be possible with TypeScript mapped types, dropping all methods/functions, so only properties & getters would be invocable on the snapshot.

I'm also kind of wondering about some sort of `function.bind` / apply magic that would try to invoke object/class methods with `this` always bound to the `proxy`. So even `snap.clickMe()` would have `this` be the proxy. Not sure if that is possible or desirable, but just an idea.

> May I merge this, or should wait?

Yep, feel free to merge. Thanks!

dai-shi:
> I kinda wonder if maybe snapshots should just have all methods dropped from the type; that would be possible with TypeScript mapped types, dropping all methods/functions, so only properties & getters would be invocable on the snapshot.

What do you mean?
This is still useful:
```js
const state = proxy({
  firstName: '...',
  lastName: '...',
  getFullName() {
    return this.firstName + ' ' + this.lastName
  }
})
```

Just dropping types doesn't help for JS users anyway. I mean we want the same capability for JS and TS.

> I'm also kind of wondering about some sort of `function.bind` / apply magic that would try to invoke object/class methods with `this` always bound to the `proxy`. So even `snap.clickMe()` would have `this` be the proxy. Not sure if that is possible or desirable, but just an idea.

Yes, for `withClass` util, I thing `.bind()` could be a nice magic. But, we need to know, if a method is for mutation, or just getting values. (If we disallow methods for getting values, and only recommends getters, it might be possible. Unfortunately, it's too counter-intuitive.)

stephenh:
> This is still useful:

Ah yeah, a non-getter method; that's true, the wrinkle with methods is that we have to trust the user to not make any side-effecting calls, i.e. they shouldn't write:

```js
const state = proxy({
  firstName: '...',
  lastName: '...',
  handleNewName(input) {
    this.firstName = ...
  }
})
const snap = snapshot(state);
const onClick = () => snap.handleNewName(...)
```

And so, dunno, it seems a little safer to just say "snapshots never have might-be-side-effecting methods", and nudge users to rewrite `getFullName` as a getter b/c it will better communicate "you really can't modify the state here".

Ah yep, exactly as you said, "we disallow methods for getting values".

I dunno, personally I find "methods can have side effects so snapshots don't have them" to be very intuitive, but agreed that would not be immediately obvious to beginners.

> I thing .bind() could be a nice magic

Nice, yeah I'll play around with it a bit when I get a chance.

dai-shi:
Continuing the last topic a little bit.

While I kept saying `this` is advanced, I don't mean to drop it. So, I'd expect this to work:
```js
const state = proxy({
  name: 'valtio',
  setName(input) {
    this.name = input
  },
  getName() {
    return this.name
  }
})

state.setName('react') // expected to work as `state.name = 'react'`

snapshot(state).setName('react') // runtime error as `snap.name = 'react'`
//              ^ that can be `never` in TypeScript, but there's no way to guess if a method contains mutation

state.getName() // just works

snap(state).getName() // should work too as `snap.name`
```

Furthermore, this is valid, right?
```js
const state = proxy({
  name: 'valtio',
  counter: 0,
  get theName() {
    ++this.counter
    return this.name
  }
})
```

stephenh:
> While I kept saying this is advanced, I don't mean to drop it

Great!

Yep, everything you outlined makes sense to me, in terms of both making sense intuitively + being how Valtio behaves today.

PR #650, opened 2023-02-04
feat(utils): deprecate proxyWithComputed

dai-shi:
Ref: https://github.com/pmndrs/valtio/pull/647#discussion_r1096397705


docs/guides/computed-properties.mdx
```diff
@@ -0,0 +1,60 @@
...
+  get doubled() {
+    return state.count * 2
+  },
+  set doubled(newValue) {
+    state.count = newValue / 2
+  },
+})
+```
+
+Be careful: You shouldn't use `this` in setters.
```
stephenh, line 41: Just poking around at this example:

```js
  it('can use object setters', () => {
    const state = proxy({
      count: 1,
      get doubled() {
        return this.count
      },
      set doubled(v: number) {
        this.count = v / 2
      },
    })

    // seems to work again the state itself
    expect(getVersion(state)).toBe(2)
    state.count = 2
    expect(getVersion(state)).toBe(3)
    state.doubled = 6
    // version ticks twice, once for "set double" and once for "set count"
    expect(getVersion(state)).toBe(5)

    // what about a snapshot?
    const snap = snapshot(state)
    expect(snap.count).toBe(3)
    expect(Reflect.ownKeys(snap)).toEqual(['count', 'doubled'])
    // @ts-expect-error snap.doubled should not compile
    expect(() => (snap.doubled = 8)).toThrowError('Cannot assign')
  })
```

It seems like the behavior of using `this.count` in a setter is not that bad? Like sure it only works in the state, and it blows up in the snapshot; but that's expected, and the typing makes it a compile error to try and call the setter anyway (which is really neat).

Afaict `this` basically is `state`, or `snap`, so, dunno, seems like maybe it's fine to use? Or am I missing something subtle?

dai-shi: Yeah, not that bad. But, I'm sure you are so experienced with `this`.

> the typing makes it a compile error 

this is convincing. let me tweak the comment.

dai-shi:
Merging this.
@stephenh Please feel free to help improving the docs: wording, examples, etc.

stephenh:
@dai-shi cool! I'll working on rebasing my docs PR on top of this. Thanks!

PR #654, opened 2023-02-05
refactor(vanilla): simplify handling object descriptor

dai-shi:
Thought this would be more consistent with https://github.com/pmndrs/valtio/pull/653#issuecomment-1416911665.

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -287,6 +287,8 @@ const buildProxyFunction = (
         key
       ) as PropertyDescriptor
       const hasValue = 'value' in desc
+      // `delete desc.value` is required, because otherwise
+      // we can't set a new value in the `if (hasValue)` block below.
       delete desc.value
       Object.defineProperty(baseObject, key, desc)
       if (hasValue) {
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -245,9 +245,7 @@ const buildProxyFunction = (
           value = getUntracked(value) || value
         }
         let nextValue = value
-        if (Object.getOwnPropertyDescriptor(target, prop)?.set) {
-          // do nothing
-        } else if (value instanceof Promise) {
+        if (value instanceof Promise) {
           value
             .then((v) => {
               value.status = 'fulfilled'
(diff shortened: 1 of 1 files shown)


dai-shi:
@stephenh Would you review this please?

src/vanilla.ts
```diff
@@ -288,9 +286,10 @@ const buildProxyFunction = (
         initialObject,
         key
       ) as PropertyDescriptor
-      if (desc.get || desc.set) {
-        Object.defineProperty(baseObject, key, desc)
-      } else {
+      const hasValue = 'value' in desc
+      delete desc.value
```
stephenh, line 290: @dai-shi just curious, by why `delete desc.value` and then reset it?

I had noticed when first diving into this that `proxyObject[key] = initialObject[key]` invoked the handler's `set` and did the whole "version"/"notifyUpdate"s process for every key on `initialObject`.

My curiosity was whether, since we've _just_ created the proxy, there should defacto be no subscribers, maybe we could skip all of that?

I might be missing something else important that the `handler.set` does, not sure.

src/vanilla.ts
```diff
@@ -245,9 +245,7 @@ const buildProxyFunction = (
           value = getUntracked(value) || value
         }
         let nextValue = value
-        if (Object.getOwnPropertyDescriptor(target, prop)?.set) {
```
stephenh, line 248: Nice! I had came across this code and couldn't really piece together why it was necessary; i.e. why we wanted to avoid the promise/child proxy logic just b/c this was a setter.

I am a little morbidly curious why it was necessary in the 1st place, but that's not super important, and great to see it go away.

stephenh (approved):
@dai-shi looks good to me!

Small disclaimer I am still in the "barely following along" stage of understanding and not "ah I immediately see why this is better" stage. :-)

dai-shi:
> Small disclaimer I am still in the "barely following along" stage of understanding and not "ah I immediately see why this is better" stage. :-)

Sure, totally fine. (And, there are only few who barely follow along.)

src/vanilla.ts
dai-shi: I wanted to avoid handling things if setter exists, because we don't know what the setter does. It can be something really unexpected. But, anyway, this would simplify it. Not really sure what happens...

src/vanilla.ts
dai-shi: AFAIR, some tests failed.

stephenh: Ah sure, it looks like it sets up child proxies (just glancing at the `subscribe.test.tsx` failure). That makes sense.

Maybe leave an in-source hint for future maintainers/newbies? LIke

```
// Unset and reset the value to ensure child proxies are wrapped
const hasValue = 'value' in desc
delete desc.value
```

dai-shi: 399f92d

stephenh: Great, thanks!

stephenh (approved):

PR #656, opened 2023-02-05
fix(vanilla): improve snapshot creation ignoring object setters in prototype

dai-shi:
ref: https://github.com/pmndrs/valtio/pull/653#issuecomment-1418045702

- [x] impl
- [x] tests
- [x] update proxy-compare

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -131,7 +131,9 @@ const buildProxyFunction = (
 
   versionHolder = [1, 1] as [number, number],
 
-  proxyFunction = <T extends object>(initialObject: T): T => {
+  proxyFunction = function proxyFunction<T extends object>(
+    initialObject: T
+  ): T {
     if (!isObject(initialObject)) {
       throw new Error('object required')
     }
@@ -261,7 +263,7 @@ const buildProxyFunction = (
             })
         } else {
           if (!proxyStateMap.has(value) && canProxy(value)) {
-            nextValue = proxy(value)
+            nextValue = proxyFunction(value)
           }
           const childProxyState =
             !refSet.has(nextValue) && proxyStateMap.get(nextValue)
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -108,23 +108,26 @@ const buildProxyFunction = (
     markToTrack(snap, true) // mark to track
     snapCache.set(target, [version, snap])
     Reflect.ownKeys(target).forEach((key) => {
+      if (Object.hasOwn(snap, key)) {
+        // Only the known case is Array.length so far.
+        return
+      }
       const value = Reflect.get(target, key)
+      const desc: PropertyDescriptor = {
+        value,
+        enumerable: true,
+      }
       if (refSet.has(value as object)) {
         markToTrack(value as object, false) // mark not to track
-        snap[key] = value
       } else if (value instanceof Promise) {
-        Object.defineProperty(snap, key, {
-          get() {
-            return handlePromise(value)
-          },
-        })
+        delete desc.value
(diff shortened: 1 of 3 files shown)


dai-shi:
@stephenh how do you like this?
The test is failing, and I found a subtle bug in proxy-compare.

src/vanilla.ts
```diff
@@ -108,30 +108,35 @@ const buildProxyFunction = (
     markToTrack(snap, true) // mark to track
     snapCache.set(target, [version, snap])
     Reflect.ownKeys(target).forEach((key) => {
```
stephenh, line 110: @dai-shi this PR in general looks great!

I copy/pasted in the tests from #653 to see what happens, and what I'm observing is that:

1) For _object_ getters, this PR is actually not necessary, b/c new snapshots have a pure/empty prototype, so the original `snap[key] = value` line finds nothing on `Object.prototype`, and so does a regular storage of `double: 2` directly on `snap` (no invocation of the proxy's object setter).

2) For _class_ getters, the snapshot comes in with `Counter.prototype` and so `snap[key] = value` _would_ get routed to the `Counter.double` setter method, and so this PR avoids that, and so yeah _would_ define  `double: 2` directly on `snap`.

However, the reason I say "would" is that `Reflect.ownKeys` still only sees `count` as the field it should copy over. If I do a super-hacky:

```
    ;[...Reflect.ownKeys(target), 'doubled'].forEach((key) => {
```

On this line, then this loops sees `doubled`, and so `defineProperty(snap, doubled, 4)` runs, and then we end up with the same behavior of object getters, the `snap.doubled` is the primitive 4 value.

So, that's what I've observed. I'll defer to you on whether you think this PR as-is is still an improvement/simplification (even though afaict it doesn't behaviorally change either object or class getter behavior, which could be fine!), or if you want to nudge it further to pick up `doubled` (and align the object/class getter behavior). 

Thanks!

dai-shi: This PR itself doesn't change the behavior. My hope is it would make `withClass` util possible.

stephenh: > My hope is it would make withClass util possible.

Gotcha. Fwiw I think `proxyWithClass` is what I would prefer b/c it handles the case of "author has some books" and wanting all authors + books to get the same modified behavior.

With `withClass`, I think the caller would have to `withClass` each of the books. I guess `withClass` could recursively `withClass` its contents, but then as new books are added (`author.books.push(new Book())` I think `withClass` would be missed.

So, in a good way, I think `proxyWithClass` is a cleaner/easier way to ensure the upgraded behavior is gotten by the entire proxy tree.

dai-shi: I'm fine with `proxyWithClass`. Even so, it can be implemented easily, without `unstable_buildProxyFunction`, if this PR is merged, or not? I didn't confirm. Oh, I see. You want it recursively. Then, this PR might not help.

stephenh: > You want it recursively. Then, this PR might not help.

Right, although it would help the `proxyWithClass`'s custom `createSnapshot` be closer to the real `createSnapshot` (the main change just being the `ownKeys` line), so if you wanted to merge it anyway / think this is a good improvement, I think that'd be great.

But also np to close out, but still great to see, as I'll probably use this PR's `createSnapshot` in the `proxyWithClass`'s version.

dai-shi: > (the main change just being the ownKeys line)

Okay, it's a bit unfortunate for my goal, but yeah, I will merge it anyway.
Maybe there's another way to avoid overriding `createSnapshot`? don't know yet.

PR #664, opened 2023-02-13
fix(vanilla): improve performance with configurable property on snapshot

### LINKED ISSUE #519: Bottleneck on large objects
Hello @dai-shi I made an experiment of having 5000 form fields with different state managers and found out that Valtio isn't performing well:

Jotai
https://codesandbox.io/s/form-test-jotai-05chcf?file=/src/App.tsx

Zustand
https://codesandbox.io/s/form-test-zustand-u53m0s

Valtio(increase SIZE to 5000)
https://codesandbox.io/s/form-test-valtio-b61ldg

dai-shi:
and some refactors.

close #519 again.

### DIFF
--- a/src/react.ts
+++ b/src/react.ts
@@ -34,6 +34,10 @@ const useAffectedDebugValue = (
   useDebugValue(pathList.current)
 }
 
+// This is required only for performance.
+// Ref: https://github.com/pmndrs/valtio/issues/519
+const targetCache = new WeakMap()
+
 type Options = {
   sync?: boolean
 }
@@ -162,5 +166,10 @@ export function useSnapshot<T extends object>(
     useAffectedDebugValue(currSnapshot, currAffected)
   }
   const proxyCache = useMemo(() => new WeakMap(), []) // per-hook proxyCache
-  return createProxyToCompare(currSnapshot, currAffected, proxyCache)
+  return createProxyToCompare(
+    currSnapshot,
+    currAffected,
+    proxyCache,
+    targetCache
+  )
 }
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -116,6 +116,9 @@ const buildProxyFunction = (
       const desc: PropertyDescriptor = {
         value,
         enumerable: true,
+        // This is intentional to avoid copying with proxy-compare.
+        // It's still non-writable, so it avoids assigning a value.
+        configurable: true,
       }
       if (refSet.has(value as object)) {
         markToTrack(value as object, false) // mark not to track
@@ -123,7 +126,14 @@ const buildProxyFunction = (
         delete desc.value
         desc.get = () => handlePromise(value)
       } else if (proxyStateMap.has(value as object)) {
-        desc.value = snapshot(value as object, handlePromise)
+        const [target, ensureVersion] = proxyStateMap.get(
+          value as object
+        ) as ProxyState
+        desc.value = createSnapshot(
+          target,
+          ensureVersion(),
+          handlePromise
+        ) as Snapshot<T>
       }
       Object.defineProperty(snap, key, desc)
     })
@@ -134,9 +144,7 @@ const buildProxyFunction = (
 
   versionHolder = [1, 1] as [number, number],
 
-  proxyFunction = function proxyFunction<T extends object>(
-    initialObject: T
-  ): T {
+  proxyFunction = <T extends object>(initialObject: T): T => {
     if (!isObject(initialObject)) {
       throw new Error('object required')
     }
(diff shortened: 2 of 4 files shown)


src/vanilla.ts
```diff
@@ -116,14 +116,24 @@ const buildProxyFunction = (
       const desc: PropertyDescriptor = {
         value,
         enumerable: true,
+        // This is intentional to avoid copying with proxy-compare.
+        // It's still non-writable, so it avoids assigning a value.
+        configurable: true,
```
dai-shi, line 121: This allows to change the configuration of the property, but it's probably not often done.
By the way, we already removed Object.freeze in #656, so assigning a new property was possible.
After all, non-writable is just to avoid misusage of snapshots.

src/vanilla.ts
```diff
@@ -116,14 +116,24 @@ const buildProxyFunction = (
...
       } else if (proxyStateMap.has(value as object)) {
-        desc.value = snapshot(value as object, handlePromise)
+        const [target, ensureVersion] = proxyStateMap.get(
+          value as object
+        ) as ProxyState
+        desc.value = createSnapshot(
+          target,
+          ensureVersion(),
+          handlePromise
+        ) as Snapshot<T>
```
dai-shi, line 136: Thought this is more self contained.

src/vanilla.ts
```diff
@@ -134,9 +144,7 @@ const buildProxyFunction = (
 
   versionHolder = [1, 1] as [number, number],
 
-  proxyFunction = function proxyFunction<T extends object>(
-    initialObject: T
-  ): T {
+  proxyFunction = <T extends object>(initialObject: T): T => {
```
dai-shi, line 147: function naming wasn't necessary.

src/vanilla.ts
```diff
@@ -320,10 +328,10 @@ const buildProxyFunction = (
     versionHolder,
   ] as const
 
-const [proxyFunction] = buildProxyFunction()
+const [defaultProxyFunction] = buildProxyFunction()
```
dai-shi, line 331: renaming just to avoid confusion.

dai-shi:
Finally fixed! https://codesandbox.io/s/form-test-valtio-forked-85ozhn?file=/src/App.tsx

src/vanilla.ts
stephenh: Ah right; shoot--even though this was not doing an `Object.freeze`, the `configurable: false` + `writable: false` defaults kicked in the "I'm going to copy now" behavior.

Just wondering, I was going to ask in the proxy-compare PR, but what do you think about, instead of copying, just failing the proxy-compare `createProxy` call if we detect non-configureable+non-writeable properties, and ask the caller to fix their properties (either by doing a copy on their side, or adjusting their descriptors)?

I'm thinking that making the copy might be too implicit/too magical, and maybe most callers would be like this code, which didn't realize it was asking `createProxy` to do a copy, and it was actually be pretty easy to throw in the "oh sure, `configurable: true`" on the caller end?

src/react.ts
```diff
@@ -34,6 +34,10 @@ const useAffectedDebugValue = (
   useDebugValue(pathList.current)
 }
 
+// This is required only for performance.
+// Ref: https://github.com/pmndrs/valtio/issues/519
+const targetCache = new WeakMap()
```
stephenh, line 39: Just for my understanding, doesn't the `configurable: true` fix to `createSnapshot` mean that snapshots shouldn't be getting copied anymore, hence the `targetCache` shouldn't be necessary?

Or are you thinking there could be non-snapshot objects that we come across that might be copied and/or lets keep it for good measure just given how important avoiding the copies is?

src/vanilla.ts
stephenh: Also, wdyt of adding a unit test that covers `snapshot`s properties are configurable? 

...or maybe make a snapshot, make the compare-proxy, call `getUntracked` and make sure it's the same instance as the snapshot? Oh, except `getUntracked` is smart enough to, when a copy is made, it still returns the non-copy anyway.

I liked the assert-again-`getUntracked` idea b/c it would be more black-box and assert what we really care about (compare-proxy isn't making copies), but that's fine, maybe just asserting against the descriptor is good enough.

...or maybe call compare-proxy with a `targetCache` and assert the map stays empty, which would imply no copies were made.

stephenh (approved):
Makes sense!

src/react.ts
dai-shi: I hoped so, but `needsToCopyTarget` function is very slow for #519 case, compared to `Object.isFrozen`.

stephenh: Oh jeez, I see, even merely evaluating "do I need to copy?" is itself expensive... :-| 

That makes sense...

dai-shi:
> I liked the assert-again-getUntracked idea b/c it would be more black-box and assert what we really care about (compare-proxy isn't making copies)

@stephenh Can you help with the idea of black-box tests? (Open a PR?)

Merging this for now.

PR #671, opened 2023-02-16
chore(tests): Add test that we probably avoid copies.

stephenh:

#664 

## Summary

Add a regression test for #664 .

Granted, this doesn't truly ensure proxy-compare doesn't make a copy, but it still seems worthwhile to prevent accidental regressions.

## Check List


dai-shi (approved):
> Granted, this doesn't truly ensure proxy-compare doesn't make a copy, but it still seems worthwhile to prevent accidental regressions.

I agree. It's nice to have. We will know if something is changed.

PR #672, opened 2023-02-16
fix(vanilla): Improve Snapshot type so that it respects builtin objects

octet-stream:
## Summary

This PR improves Snapshot type so that it does not affect builtin object, like Date.

Examples for old and new behaviours can be found in TS playground: for TS [3.5.1](https://www.typescriptlang.org/play?ts=3.5.1#code/C4TwDgpgBAggdiAYgVzgY2ASwPZygXigAoA6MgQwCcBzAZwC4pyEBtAXQEoCA+JhAKH6hIsWgCUIAMwJQA3vyiKoAEmUA3cgBss2CZMbBKyCPwC+g4dBgB3cpmAQAJgB4AKr0KuoEAB4O4jrRQAAqU2AC2mLQQzphwkhCUUACqvAD8KVCMroIA9LlQAMoQ0AAWwMBgDPnU9qXIAEYkaBG5YOEBlLS5Gto4uQ2a2A25ACwAnACM5OQAzACso5MATPMADONrCcvkk-Prk5MQhwDsGw3zh5vzEPOzubSUaD3MmJqa5CTAtADEADLLNYAWgBAA4hOBoAB5TSOQpwchVUrYYBuDwKJReXz+QKwBAodA6OAYpRKDI5Umk7LePwQAJBGDiKQkymk8ks1mKanYum40IRKIxZggbgczls2C2exONFi8VcuSUCDkRy4TQgKAsADSUDiUAA1hAQNhpK42IwYXCEUiUW5tWxuOYISJ4YjaMjUe4CCysbT6aI9HKoOzxdy-XywpFos5UPq4NhrHBRfKyZK7A4XO4g1SoL6cUFZCxDSBGLRDHFqOaoLH44nzCnWRlZEqVWqNdrdXhiybc1XXTbgMs7VqHfWGzm87ygjWE3B2NmGxkW6q4Oqitb3Sih64WHBkOEGokHfPx6HcxZIeu3R7t+jSZP-YzA+KQ5yw-mQpHBTG4HHZ8nFzTaVMwAlN3ynOQiyNUtyzgStGBnOsF0bRVlRXNcOz1btTSrHcRzHU8slzGkP0Quc2GQ+UlzQttc13fdD0oNgT0IylsgvEQAHU6ktftN2AGReI3D1nHkUlaAiCACQwHA4CIDgYMoCsWTAL9okYfkoxiPcD0SUCoHCRFGAAWURZwyyUuCABpq1-WskxZaJgEYYpUQsit9OlcIGFgShKHIEBRKDTBHEU5TxSwYBNAgMK4KDRwIGAOxNB89y4KgAAfKA93eIM0BbDMYGcqAABFyAcINkDARxyqcIrGDKirSVMUUWo46BuOAUoADkIGsPiPRkAbbTEpQJPCKTUBk3B5MYNRsBClS1Jiz8BWjHTGP0wywBMsy0uoGyyP0pyXMS8zYOoTyHG8xgYD8gKgvFELYuoINIuil74sS5LUouzLsuQXLxXy5VCuKxqTHFKqarBhrapZNq2v4fIoAASWkY1kAAcneKAlTAD40GgAADMjiagaw6igYnhXJrroDgbEoAcMsKbeTQoDAchaFoZ1oFR2hgmwAArbA0RkB9cULYsXqrMjTGDZmjGgRhJC0aJ2rR2gIaF0WZAF3WxYh0UgA) and for TS [4.9.5](https://www.typescriptlang.org/play?ts=4.9.5#code/C4TwDgpgBAggdiAYgVzgY2ASwPZygXigAoA6MgQwCcBzAZwC4pyEBtAXQEoCA+JhAKH6hIsWgCUIAMwJQA3vyiKoAEmUA3cgBss2CZMbBKyCPwC+ggPQWoAZQjQAFsGBgGV6pmAPkAIxJpsAFsLMEC4ABNKWgsNbRwLH01sHwsAFgBOAEZycgBmAFZUzIAmfIAGdLLJCGLyTPzyzMyIJoB2Cp98psr8iHzci1pKNBjmTE1NchJgWgBiABlisoBaRYAOIXBoAHlNcJs4clcHbGAAHgAVXnwFJQuoCAAPYAgI2lgEFHQdOFulJQA-FALn9-lBGPcni83qI9KCwYDgfCEeDgQ9nq9wu8AAqUIKYWgQM7MEDcZEoxEwADu5E8EHClzJFIpjFklAg5HCuE0ICgLAA0lBMHgANYQEDYaQXNiMXb7Q7HU6XAVsbjmTYiA5HWgnc5XAigyEYmEwcRSclQIEglloqGYnF4wIEomoEVwbBUuBM5mKIHU2kvBlXC3-CHo6FYuQsMUgRi0QzC6gyqCu92e8w+hFAtkcrlwHl8wXCqAxyXA5NaxV6lVqkMosN2mGpj1wdh1n1A9mc7m8ys6pUXFhwZCBHwQSiqtuZ5kQwTCaAAdU8Djlfd1MlXCv753k-1oQQgXwwODgRA4cYTcGooLAjudjFx+MJZ2Ho-H3sUgSOjAAskczvGlCJgANCmcBui2H5QISwCMHY5yAYmUF0oEDCwJQlDkCAZy7iimDhBeQFXhaWDAJoECEYmFrhBAwC0poaGIVeUAAD5QMOEwWmgXaBjAsFQAAIuQLwWsgYDhMJ9J8YwQkif8phkgpc5bFAS5eAAchAVJrqcMg6TuoL7oEh6oMeuBnowajYPhN53oSD52USr5jpQUFfmAv7-kx1Cgc2npQTBcG0QBl7UMhLyoYwMAYVhOEWvhlHERSpHkYl14UjRdHjIxoWsexyCcRS3Ecrx-GySYFJiRJpUyZJoJKUp-BWFAACS0gSsgADkExQOyYCTGg0AAAZ+XAQ1QFSy5QENJLjV40BwFCUAvPGE3jJoUBgOQtC0Bq0AtbQ2LYAAVtgjIyEaEbvLI0bimlyajaYlrLUY0CMJIWiEspIgHeVR2nTIB3-Wd5VkkAA). As you can see from the `OldSnapshot` example, it iterates through every member of `Date`, `Map` and `Set` object and converts these objects to readonly POJOs for some reason.

The solution might not be perfect though, so I would suggest adding some tests for typings (or maybe you already have and I just won't be able to find any, so please let me know).

Also, I wouldn't be able to run tests, because this project is hardly relying on Yarn, so will see if the tests are passing.

And lastly, I would like to ask if you can expose both `Options` (for useSnapshot hook) and `Snapshot<T>` types, so that people might implement extensions upon this hook and rely on public API.

## Check List


dai-shi (approved):
The change looks reasonable!

dai-shi:
Thanks for your suggestions!

>  so I would suggest adding some tests for typings

Would you be able to help on it?

> this project is hardly relying on Yarn

That's true, but I think you can run normal tests with `npm test`. (CI tests are really complicated.)

> if you can expose both Options (for useSnapshot hook) and Snapshot<T> types,

As you might guess, the first one can be extracted fairly easily, and the second one is intentional.

> people might implement extensions upon this hook and rely on public API.

If they rely on internal types, I'd suggest to pin the valtio version and use it. We may really change them in the future without notice.

octet-stream:
> Would you be able to help on it?

Yup, I think I can write bunch of tests for Snapshot type using something like [`tsd`](https://www.npmjs.com/package/tsd). But I am have no idea how to get this done for patches for TS 3.4

> That's true, but I think you can run normal tests with npm test. (CI tests are really complicated.)

Nope, it doesn't work and I have to install Yarn. And maybe I will, it's just that I have to do this for such small thing. Not a big problem, I think. But having scripts running using npm might be more convenient, as it is the default (and I personally could've run tests via pnpm).

> As you might guess, the first one can be extracted fairly easily, and the second one is intentional.

And so I did, yes, but it would be just much more handful to have useSnapshot `Options` to be part of the public API.

> If they rely on internal types, I'd suggest to pin the valtio version and use it. We may really change them in the future without notice.

Alright then :)

dai-shi:
> > Would you be able to help on it?
> 
> Yup, I think I can write bunch of tests for Snapshot type using something like [`tsd`](https://www.npmjs.com/package/tsd). But I am have no idea how to get this done for patches for TS 3.4

We can simply drop such tests for very old TS. Please make it a new test file.

> > That's true, but I think you can run normal tests with npm test. (CI tests are really complicated.)
> 
> Nope, it doesn't work and I have to install Yarn. And maybe I will, it's just that I have to do this for such small thing. Not a big problem, I think. But having scripts running using npm might be more convenient, as it is the default (and I personally could've run tests via pnpm).

Oops, my bad. My recent change requires yarn for test.
Yeah, we may consider improving those build/test scripts in the future and I hope to make it work with `npm` too.

octet-stream:
Added some tests for Snapshot type, ended up using [`ts-expect`](https://www.npmjs.com/package/ts-expect). They will be run through regular `tsc --noEmit` command.

.gitignore
```diff
@@ -25,3 +25,7 @@ yarn-error.log*
 # misc
 .DS_Store
 .idea
+
+# pnpn & npm lockfiles
+pnpm-lock.yaml
+package-lock.json
```
octet-stream, line 31: I also added these two lines, so I don't have to use `--no-lockfile` flag every time.

src/vanilla.ts
```diff
@@ -15,17 +15,15 @@ type Op =
...
   ? T
-  : T extends Promise<any>
+  : T extends Promise<unknown>
   ? Awaited<T>
-  : {
-      readonly [K in keyof T]: Snapshot<T[K]>
-    }
+  : T extends unknown[]
+  ? readonly Snapshot<T[number]>[]
+  : T extends { [key: string]: unknown }
```
stephenh, line 24: Fwiw classes don't extend `{ [key: string]: unknown }`, so this will stop `Snapshot` from recursing into classes.

Which is a good thing for types like `Date`, but doesn't technically match `snapshot`s runtime behavior, which will snapshot anything that passes `canProxy`.

Currently `canProxy` just special cases `Date` / `Number` / etc., and I think the mapped type will probably have to do that as well (?) if it wants to exactly line up the type system behavior to the runtime behavior.

(That is typically how I've seen this done, like a `type Primitive = string | number | Date | ...` and then `T extends Primitive ? T : ...recurse...`, but I was actually hoping this would work b/c it's much smaller/more succinct)

tests/typings/Snapshot.test-d.ts
```diff
@@ -0,0 +1,95 @@
+import type { TypeEqual } from 'ts-expect'
+import { expectType } from 'ts-expect'
+import { INTERNAL_Snapshot as Snapshot } from '../../src/vanilla'
+
+// Converts object properties to readonly
+expectType<
+  TypeEqual<
```
stephenh, line 7: This is really neat, I hadn't seen `ts-expect` before and am going to start using it. Thanks! :-)

src/vanilla.ts
octet-stream: Hmm. Yes, you're right. Thanks for pointing this out! I've tried it in TS playground and it doesn't pass this test.
From the top of my mind I was thinking about something like this test:
```ts
type IsConstructable<T> = T extends new (...args: any[]) => any ? true : false
```

But seem to work for `typeof MyClass`. I'll try and find a better solution, or we just keep it as you've suggested. See [here](https://www.typescriptlang.org/play?ts=4.9.5#code/C4TwDgpgBAkgzgYQPYDs7AE4FcDGwCGARgDYQA8AKgHxQC8UFUEAHsBCgCZxQoQDuUABQA6UfgwBzOAC4o+FCADaAXQCUdGvJBQA-FExZosgGb5icCAChLOYvjjcAqhYxQA3paheoxgJYZ0ADl8AFsIAEJZdAxfFAlPbzsg0IiozFj4gF9rHFR0KCwXOh5+KGcIDEFVa1BIKAARfDZ4ZDQDPCJSYpa89oIScka2Kkta6AQkuB627A6B7sRe2f7SMjGkYzKXEbHYNvkcCGno3BXoemO+zvJ1zcKKkaA).

tests/typings/Snapshot.test-d.ts
octet-stream: There's also similar stuff in [`@type-challanges/utils`](https://www.npmjs.com/package/@type-challenges/utils) package, but I'm not sure if it intended for tests too (probably not) x)

src/vanilla.ts
octet-stream: Not to mention some of the builtin objects technically constructable: like Date, Map and Set

dai-shi (changes requested):
So, this needs to be changed to support custom classes?
Feels like we never reach an ideal solution. false positives are better than false negatives.

octet-stream:
> Feels like we never reach an ideal solution. false positives are better than false negatives.

Can't agree with you more.
So I've added `SnapshotIgonre` union type with the list of objects that Snapshot will not turn into `readonly`. Please let me know if I've missed something.

octet-stream:
Out of curiosity: Why GitHub does not let me to request review from both of you at the same time? x)

dai-shi (commented):
I don't know if there are any edge cases uncovered. Seems reasonable. (And makes me feel Snapshot type will be INTERNAL forever.)

I'll wait for @stephenh 's review.

octet-stream:
Wellp, looks like I need yarn after all :)

octet-stream:
Wait, which yarn version do I need? Is it 1.x?

dai-shi:
yes, it's classic yarn.

octet-stream:
Okay, I have no idea why tests for old ts versions are failing. It doesn't seem like I've changed anything significant since yesterday, and yesterday's builds were passing the tests.

Also, please add contribution guidelines of some sort, just to clarify how to develop for valtio.

tests/typings/Snapshot.test-d.ts
```diff
@@ -0,0 +1,107 @@
+import type { TypeEqual } from 'ts-expect'
+import { expectType } from 'ts-expect'
+import { INTERNAL_Snapshot as Snapshot } from '../../src/vanilla'
```
dai-shi, line 3: I think this breaks the old ts tests.

dai-shi:
```suggestion
import { INTERNAL_Snapshot as Snapshot } from 'valtio/vanilla'
```
might help.

Also, I'm not sure if `tests/*/*.ts` is supported. Suppose it does, can you name the file `tests/typings/snapshot.test.ts`?

octet-stream: I'm not sure if I understand you correctly. These files tests as regular `tsc --noEmit`, not via `jest`. According to [`tsconfig.json`](https://github.com/pmndrs/valtio/blob/e159c821672c6e3914346131fe2aec64dad8f35f/tsconfig.json#L17), `tsc` should go through this file as expected. Why do I need to rename it to `*.test.ts`? Also, if I do so, I have to ignore it in jest.

stephenh:
> Feels like we never reach an ideal solution.

Fwiw @dai-shi if the "gah classes" gives you too much grief, I think a potential compromise is requiring users to opt-in via a base class, like:

```ts
// in valtio
abstract class BaseStore {
  [hiddenSymbol]: ...
}

// in apps, opt-in to "treat my class as a store"
class User extends BaseStore {}
```

I know this sort of "forced base class" is a trigger for OO purists, _but_ it would let both the runtime / `canProxy` do an `target instanceof BaseStore` and mapped type / `Snapshot<T>` do an `target extends BaseStore` and know 100% for sure the user wants their data to be proxified/snapshotted/etc.

I.e. I believe it'd let us remove the "disallow list" of WeakMap+WeakSet+other special things we think of, and move to just "is it a basestore (or the vanilla object literal)"? 

I dunno, personally I'd like to avoid that, and try to keep just "vanilla classes" working out-of-the-box. But just mentioning it as a potential middle ground for someday.

tests/typings/Snapshot.test-d.ts
```diff
@@ -0,0 +1,107 @@
+import type { TypeEqual } from 'ts-expect'
+import { expectType } from 'ts-expect'
+import { INTERNAL_Snapshot as Snapshot } from 'valtio/vanilla'
+
+// Converts object properties to readonly
```
stephenh, line 5: @octet-stream do you mind add an explicit test for classes, i.e. add in:

```
class User { firstName = "u1" }
```

And show that it does get recursively `Snapshot`-ified?

stephenh (approved):
Small ask for a `class User`-or-similar test to make sure it really works, but my initial read is that it looks good! Thanks @octet-stream !

tests/typings/Snapshot.test-d.ts
octet-stream: Sure. I've added a test case for classes.

octet-stream:
Fixed Snapshot type tests for TS 3.7. It appears that `import type` was introduced in [`3.8`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export).

dai-shi (approved):
LGTM.

Are you hesitant renaming the new file to `tests/typings/snapshot.test.ts`?

octet-stream:
> Are you hesitant renaming the new file to tests/typings/snapshot.test.ts?

No, but this will require configuration for jest, I think. It should be ignored by it.

octet-stream:
Didn't catch this one. For some reason VSCode decided not to show ESLint errors for me :D

dai-shi:
> No, but this will require configuration for jest, I think. It should be ignored by it.

Oh, I wasn't aware that. Which part is bad? We have `tests/snapshot.test.ts`, so the `typings` subdir could be something, but not that one...

octet-stream:
> Which part is bad?

The part where jest will likely fail trying to run through this file :)
These test files should not be executed in *runtime*. They only meant for `tsc` typeschecks.
But technically we could just ignore the whole `typings` directory for jest, I guess. Doesn't really matter where these tests placed, or how they named. They're just regular TypeScript files.

I can add configuration for jest and rename the file as you ask. If you insist :)

dai-shi:
> The part where jest will likely fail trying to run through this file :)

Ah, I thought the other way around.
Well, we run those type only tests with jest, so, it should be fine. Does it cause CI errors?

octet-stream:
> Well, we run those type only tests with jest, so, it should be fine. Does it cause CI errors?

If jest will ignore test file without any test in it, then it should be fine. I'm not familiar with jest that much, but, for instance, AVA will fail in that case. So I just assumed it might have same behaviour, because it seem reasonable.

octet-stream:
Google says it will fail. So in that case we need to either ignore these files (as I suggested - the whole `tests/typings` directory), or add a noop test, or add [todo test](https://jestjs.io/docs/api#testtodoname).

dai-shi:
This is how it's done in jotai.
https://github.com/pmndrs/jotai/blob/main/tests/vanilla/types.test.tsx
So, it's a noop test?

octet-stream:
It does seem like it. I don't see any use of jest's `expect` function. But they look like they're regular jest tests. I can rewrite these test like this, if you want.

dai-shi:
> Not sure though why do use have [`Component`](https://github.com/pmndrs/jotai/blob/11add6285012fed1c9b25928e3d7ade694e3dd00/tests/vanilla/types.test.tsx?rgh-link-date=2023-02-18T09%3A04%3A32Z#L38) at the end of each test.

I think some of eslint rules complain otherwise.

dai-shi:
> I can rewrite these tests like this, if you want.

Yeah, please try it for now. We'll revisit later with other repos. Or maybe with #646.

I may consider including typing tests in `tests/snapshot.test.ts` for now, as it works with ts 3.7.5. That makes all test files flat.

octet-stream:
> I may consider including typing tests in tests/snapshot.test.ts for now, as it works with ts 3.7.5. That makes all test files flat.

I've merged typings tests file into `tests/snapshot.test.ts`.

tests/snapshot.test.ts
```diff
@@ -44,6 +45,136 @@ it('should not change snapshot with assigning same object', async () => {
   expect(snap1).toBe(snap2)
 })
 
+it('converts object properties to readonly', () => {
```
dai-shi, line 48: Cool. Let's add `describe` something like:
```suggestion
describe('snapshot typing', () => {
  it('converts object properties to readonly', () => {
```

dai-shi (approved):
Thanks for your contribution!!

tests/snapshot.test.ts
octet-stream: Dome.

PR #673, opened 2023-02-16
fix(vanilla): proxy can be created from snapshot

dai-shi:

Fixes #670 

## Summary

#654 broke one use case described in #670. (tbh, this isn't intended use case).
When copying from an initial object to a proxy base object, we should make it writable to make the proxy object mutable.

## Check List

### DIFF
--- a/tests/snapshot.test.ts
+++ b/tests/snapshot.test.ts
@@ -42,3 +42,10 @@ it('should not change snapshot with assigning same object', async () => {
   const snap2 = snapshot(state)
   expect(snap1).toBe(snap2)
 })
+
+it('should create a new proxy from a snapshot', async () => {
+  const state = proxy({ c: 0 })
+  const snap1 = snapshot(state)
+  const state2 = proxy(snap1)
+  expect(state2.c).toBe(0)
+})
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -299,14 +299,14 @@ const buildProxyFunction = (
         initialObject,
         key
       ) as PropertyDescriptor
-      const hasValue = 'value' in desc
-      // `delete desc.value` is required, because otherwise
-      // we can't set a new value in the `if (hasValue)` block below.
-      delete desc.value
-      Object.defineProperty(baseObject, key, desc)
-      if (hasValue) {
+      if ('value' in desc) {
         proxyObject[key as keyof T] = initialObject[key as keyof T]
+        // We need to delete desc.value because we already set it,
+        // and delete desc.writable because we want to write it again.
+        delete desc.value
+        delete desc.writable
       }
+      Object.defineProperty(baseObject, key, desc)
     })
     return proxyObject
   }


octet-stream:
Looks good to me, and that does the trick.

> tbh, this isn't intended use case

Well, as alternative, I would suggest to just have a way to convert the snapshot back to POJO (recursively). Is there any way to get this done? Maybe I missed something, again. Like, MobX has this [`.toJS`](https://mobx.js.org/api.html#tojs) utility that does exactly that.

dai-shi:
I said not intended, but I think it's valid.

What you pass to `proxy()` is an object to initialize the proxy object. It's basically used and thrown away (but we keep the reference, which is tricky. I might re-consider it in the future version.)

>  a way to convert the snapshot back to POJO (recursively).

Do you mean a snapshot to a new plain object, that's basically deep cloning. ex. `lodash.cloneDeep()`.

octet-stream:
> Do you mean a snapshot to a new plain object, that's basically deep cloning. ex. lodash.cloneDeep().

Yep, so that I can use it to create a new proxy. If the current way isn't correct (but I see that you said it just not intended).

In my case, I don't really care much of what happens with nested objects, just needed to put part of the state to context for easy access from within its descendants.

dai-shi:
> > Do you mean a snapshot to a new plain object, that's basically deep cloning. ex. lodash.cloneDeep().
> 
> Yep, so that I can use it to create a new proxy. If the current way isn't correct (but I see that you said it just not intended).

When you create a new proxy, it will copy the `initialObject` recursively anyway, so this PR would be better as of now (instead of cloning on your end.)
So, it is valid.

> In my case, I don't really care much of what happens with nested objects, just needed to put part of the state to context for easy access from within its descendants.

My question, in your use case, is if you can pass the child proxy via context, instead of creating a new proxy (which is isolated from the original proxy. maybe it's intentional.)

octet-stream:
> My question, in your use case, is if you can pass the child proxy via context, instead of creating a new proxy (which is isolated from the original proxy. maybe it's intentional.)

Is this valid? I don't quite understand, maybe. Documentation says that I need to use [snapshots](https://github.com/pmndrs/valtio#react-via-usesnapshot) within render function. Does the component react on changes in state object?

dai-shi:
It's valid, but very tricky, so it's hard to educate.

```jsx
const Component = () => {
  useSnapshot(state.child) // use it here to subscribe
  return (
    <MyContext.Provider value={state.child}>
      ...
```

> I need to use [snapshots](https://github.com/pmndrs/valtio#react-via-usesnapshot) within render function.

We've changed the word over time, maybe it's still not ideal.
snapshot objects should be used in a component render function body, but if you pass things to child components (via props or contexts), they should be primitive values or state objects, not snapshot objects.

tests/snapshot.test.ts
```diff
@@ -60,3 +60,10 @@ it('should not cause proxy-compare to copy', async () => {
   const cmp = createProxy(snap1, new WeakMap())
   expect(getUntracked(cmp)).toBe(snap1)
 })
+
+it('should create a new proxy from a snapshot', async () => {
+  const state = proxy({ c: 0 })
+  const snap1 = snapshot(state)
+  const state2 = proxy(snap1)
+  expect(state2.c).toBe(0)
```
stephenh, line 68: @dai-shi with this approach, this assertion fails:

```
expect(state2).toBe(state);
```

Wdyt of teaching `proxy` to realize "I'm being passed a snapshot" and then using some sort of `getOriginalObject` / `getUntracked` approach to just recover the existing proxy?

It seems like it'd be nice to get back the same store identity.

Or, alternatively, just throw an error and say "you passed a snapshot to proxy...don't do that".

stephenh:
> snapshot objects should be used in a component render function body, but
> if you pass things to child components (via props or contexts), they
> should be primitive values or state objects, not snapshot objects.

Fwiw I wrote up my current understanding of the behavior / your recommendation's @dai-shi  here:

https://github.com/pmndrs/valtio/pull/676

Happy to edit if I misunderstood.

tests/snapshot.test.ts
octet-stream: > Or, alternatively, just throw an error and say "you passed a snapshot to proxy...don't do that".

Wouldn't the be a breaking change?

stephenh: Yes, I'm just thinking if @dai-shi thinks the "proxying a snapshot" is enough of a unintended use case / probably not what the user really wanted to do, it'd be better to just throw and guide them to a better pattern, i.e. specifically passing the store around directly (which I believe is the recommended approach).

(Or, maybe we do the 1st option, which is that "proxying the snapshot" does become the canonical way of getting "back to the store" from the snapshot, which again AFAIU is unintended/accidental behavior today.)

dai-shi: Maybe this the misguided. I think we should support creating a new proxy from a frozen (incl. non-configurable&non-writable) object. And, a snap is just an example of it. And we support it.

It's definitely not about recovering the original proxy.

stephenh: > I think we should support creating a new proxy from a frozen

Ah sure; is that what @octet-stream wants to do? I might have misread the thread, but I thought the idea was to get back to the original store.

I guess Nick's comment "so that I can use it (the snapshot) to create a new proxy" does explicitly ask for going from snapshot --> "new store", but the intervening comments made me jump to the conclusion Nick was running into the "how do I get a store in a child that was passed a snapshot" wrinkle.

And not literally "I want a completely separate / 2nd additional store that is 100% detached from the first store".

But I could have made the wrong assumption.

stephenh: From @octet-stream 's example, he almost certainly wants the original store; he's using a `useSnapshot` against a root list of notes:

```
  const {items} = useNotesStateSnapshot()

  return (
    <ul>
      {items.map((note) => (
        <li key={note.id}>
          <NoteStateContextProvider data={note}>
            <NoteCard />
          </NoteStateContextProvider>
        </li>
      ))}
    </ul>
```

Passing each note (a compare proxy) into a child context and doing `proxy(data)`:

```
  const StateContextProvider: FC<ProviderProps<T>> = ({data, children}) => {
    // Use `useMemo` instead of `useRef` with `valtio`, so that we can reload page state if the `data` is changed
    const state = useMemo(() => proxy(data), [data])
```

Which I just really have to assume the desire is "get me the note[0] model from the original store, so I can mutate it in my child if I want":

This is why I think calling `proxy(snapshotFromSnapshot)` or `proxy(snapshotProxyFromUseSnapshot)` should either:

a) throw a "this is probably not what you meant to do" un-intended use error, or 
b) return the original underlying store that creates the `snapshot` snapshot or `useSnapshot 

Instead of the current behavior which, AFAIU creates a 2nd store that is 100% disconnected (different data, different subscribers) from the 1st store and just surely is not what the user wanted.

stephenh:
> As alternative, I would suggest to just have a way to convert the snapshot
> back to POJO (recursively). Is there any way to get this done?

@octet-stream per this ask of "just give me a POJO", FWIW this is what `snapshot` returns. Just a 100% POJO/disconnected copy of the store's data.

However, if you're using `useSnapshot`, then those are also "snapshots" but wrapped in a proxy.

If you have a provided-by-`useSnapshot` snapshot-proxy, IMO @dai-shi it would be desireable to be able to pass that to `snapshot` and basically "unwrap it" back to the non-access tracking / pure-POJO snapshot.

Granted, the user should know "don't render against this", b/c they will lose access tracking, but if they wanted to put the snapshot on the wire somewhere, they probably would prefer using the `snapshot`-based POJO and not `useSnaphot`-based POJO proxy.

Granted, @dai-shi perhaps your suggestion is for the user to just always have the original store available, and they can `snapshot` it directly if they want a POJO, instead of trying to recover snapshots (or stores per the comment from) from snapshot-proxies.

Basically given either of a `snapshot` snapshot a `useSnapshot` snapshot-proxy, I think the user should be able to easily "get back" to both the `snapshot` snapshot (from the snapshot-proxy by calling `snapshot` on it) as well as the non-snapshot store (from either the snapshot-proxy or the snapshot-pojo by calling `proxy` on it).

tests/snapshot.test.ts
dai-shi: > he almost certainly wants the original store

It might be the case, and if so, we could do it right? https://github.com/pmndrs/valtio/discussions/657#discussioncomment-4878032

dai-shi:
I feel like allowing snap -> state conversion leads to misconception, but I might be wrong.

> perhaps your suggestion is for the user to just always have the original store available

you get it right. I'd suggest to avoid overusing what `useSnapshot` returns.

tests/snapshot.test.ts
stephenh: Yes. @octet-stream can you confirm if @dai-shi 's linked comment is what you're trying to do?

Fwiw @dai-shi this is what the docs PR I opened tries to clarify for users of useSnapshot; it has your example as-is, explaining that they should pass the store if they want to mutate.

I don't think I included a disclaimer that if they proxy a snapshot, it will make a completely separate store, but that's probably a good idea.

octet-stream:
> From @octet-stream 's example, he almost certainly wants the original store; he's using a useSnapshot against a root list of notes

I believe I said before that I don't really care if that would be original store, or a separate one. For me, creating a new snore from snapshot is just a natural way to pass the store's parts to through `React.Context` to child components, and at the same time subscribe to the original store to track changes for the whole list (in this case).

Within the `NoteStateContextProvider` I can access a note that was passed to it. The child components will react to its changes anyway, so it's ok as long as it work that way. At the same time I could also update the original store, if needed, because all Note-related components also wrapped within `NotesStateContextProvider`, and all subscribed components will react to this change. [Here](https://github.com/octet-stream/next-mikro-orm-trpc-template/blob/e616a99f7ebcefee0a38c2c746dae090a6b179a5/src/component/NoteModal/NoteCreateModal/NoteCreateModal.tsx#L32), for example, I update the list when a new note is added. Because I subscribed to the original store, notes list will be updated as I expect, and for the new note the new `NoteStateContextProvided` with corresponding store is created. But then again, I don't care if that would be a separate store, or a part of the original.

octet-stream:
Also, the main point is still that you broke previous behaviour. I almost certain it still falls into breaking changes category, if you follow semver.

PR #675, opened 2023-02-19
feat: Add isStore.

stephenh:

Related to my `useStore` prototypes (https://github.com/pmndrs/valtio/discussions/662)

## Summary

When working generically with a store, and wanting to handle it recursively, it's useful to ask "is this nested value itself a store?".

Currently this is not necessary for Valtio & proxy-compare, b/c `createSnapshot` _pushes_ "this is special"-ness into proxy-compare's `markToTrack`, but 3rd party libraries/hooks like `useStore` don't have that blessed push-based mechanism, so need a pull-/probed-based mechanism to detect store-ness/special-ness.

Very open to alternative approaches/impls that accomplish the "is this a store?" goal.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -33,6 +33,8 @@ type Snapshot<T> = T extends AnyFunction
  */
 export type INTERNAL_Snapshot<T> = Snapshot<T>
 
+const IS_STORE_SYMBOL = Symbol()
+
 type HandlePromise = <P extends Promise<any>>(promise: P) => Awaited<P>
 
 type CreateSnapshot = <T extends object>(
@@ -242,6 +244,12 @@ const buildProxyFunction = (
         }
         return deleted
       },
+      get(target, prop, receiver) {
+        if (prop === IS_STORE_SYMBOL) {
+          return true
+        }
+        return Reflect.get(target, prop, receiver)
+      },
       set(target: T, prop: string | symbol, value: any, receiver: object) {
         const hasPrevValue = Reflect.has(target, prop)
         const prevValue = Reflect.get(target, prop, receiver)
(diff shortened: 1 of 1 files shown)


dai-shi:
`isStore` should be implemented on your end. Something like this:
```js
const isStore = (x) => getVersion(x) !== undefined;
```
There might be room for improvement in docs.

I think I did it in some of my valtio third-party libraries.

Closing.

src/vanilla.ts
```diff
@@ -242,6 +244,12 @@ const buildProxyFunction = (
         }
         return deleted
       },
+      get(target, prop, receiver) {
```
dai-shi, line 247: Just a note, we want to make sure valtio proxy doesn't trap `get`.

dai-shi:
> There might be room for improvement in docs.

Yeah, we should add this `isStore` recipe in: https://github.com/pmndrs/valtio/blob/main/docs/api/hacks/getVersion.mdx

Another impl:
```js
const isStore = (x) => typeof getVersion(x) !== 'number'`;
```

BTW, I prefer `isProxy` to `isStore`.

stephenh:
> Yeah, we should add this isStore recipe in:

Wdyt about adding `isProxy` but based on your impl, i.e. a helper method that just calls `getVersion`?

Just thinking it's a more ergonomic/more discoverable to users than having to find it as a suggested recipe in the `getVersion` docs and then implement their own.

dai-shi:
It is only for library authors. I'd prefer hiding it in the `getVersion` docs.

stephenh:
Sounds good. Fwiw just in terms of push-vs-pull, it might be neat if instead of `markToTrack`-ing every snapshot, Valtio just did like:

```
markToTrackFunction((obj) => isSnapshot(obj);
```

Which would let proxy-compare ask "should I proxy this object?" by just calling the function.

In theory it'd cut down on the overhead of pushing every snapshot into the mark to track set. :shrug: 

Not a big deal, just mentioning as a potential micro-optimization.

dai-shi:
Yeah, that would be another way.
For now, I avoided it because it's not predictable in terms of performance. People can register a heavy function, and multiple functions (which requires to iterate those functions).
But, let me keep that in mind for the future version of proxy-compare, which I want to simplify further (your suggestion of not supporting frozen object might be in.)

stephenh:
> People can register a heavy function

Good point!

PR #676, opened 2023-02-20
docs(useSnapshot): Clarify parent/child useSnapshot usage.

stephenh:

Potentially clarifies behavior discussion from #673.

## Summary

Documents the `snap.books.map(book, i)` approach that @dai-shi mentioned is current best practice for passing stores to children.

## Check List

### DIFF
--- a/docs/api/basic/useSnapshot.mdx
+++ b/docs/api/basic/useSnapshot.mdx
@@ -9,9 +9,9 @@ description: 'Create a local snapshot that catches changes.'
 
 Create a local snapshot that catches changes.
 
-Normally, Valtio's snapshots (created via <a href="/docs/api/advanced/snapshot">`snapshot()`</a>) are recreated on _any_ change to a store, or any of its child stores.
+Normally, Valtio's snapshots (created via <a href="/docs/api/advanced/snapshot">`snapshot()`</a>) are recreated on _any_ change to a proxy, or any of its child proxies.
 
-However `useSnapshot` wraps the Valtio snapshot in an access-tracking proxy, so your component is render optimized, i.e. it will only re-render if keys that it (or its child components) specifically accessed have changed, and not on every single change to the store.
+However `useSnapshot` wraps the Valtio snapshot in an access-tracking proxy, so your component is render optimized, i.e. it will only re-render if keys that it (or its child components) specifically accessed have changed, and not on every single change to the proxy.
 
 ## Usage
 
@@ -78,7 +78,7 @@ Note if `BookView` is `React.memo`d, the 1st `BookView` will not re-render, b/c 
 
 ### Child Components Making Mutations
 
-The above approach works if `BookView` is read-only; if your child component needs to make mutations, then you'll need to pass the store:
+The above approach works if `BookView` is read-only; if your child component needs to make mutations, then you'll need to pass the proxy:
 
 ```jsx
 function AuthorView() {
@@ -93,7 +93,7 @@ function AuthorView() {
 }
 
 function BookView({ book }) {
-  // book is the store, so we can re-snap it + mutate it
+  // book is the proxy, so we can re-snap it + mutate it
   const snap = useSnapshot(book)
   return <div onClick={() => book.updateTitle()}>{snap.title}</div>
 }
--- a/docs/api/basic/useSnapshot.mdx
+++ b/docs/api/basic/useSnapshot.mdx
@@ -7,12 +7,17 @@ description: 'Create a local snapshot that catches changes.'
 
 # useSnapshot
 
-Create a local snapshot that catches changes. This hook actually returns a wrapped snapshot in a proxy for
-render optimization instead of a plain object, which is what <a href="/docs/api/advanced/snapshot">`snapshot()`</a> returns.
+Create a local snapshot that catches changes.
 
-#### Rule of thumb: read from snapshots in render function, otherwise use the source.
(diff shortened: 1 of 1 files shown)


docs/api/basic/useSnapshot.mdx
```diff
@@ -7,12 +7,17 @@ description: 'Create a local snapshot that catches changes.'
...
+Normally, Valtio's snapshots (created via <a href="/docs/api/advanced/snapshot">`snapshot()`</a>) are recreated on _any_ change to a store, or any of its child stores.
 
-The component will only re-render when the parts of the state you access have changed; it is render-optimized.
+However `useSnapshot` wraps the Valtio snapshot in an access-tracking proxy, so your component is render optimized, i.e. it will only re-render if keys that it (or its child components) specifically accessed have changed, and not on every single change to the store.
+
+## Usage
+
+### Read from snapshots in render, write to the store in callbacks
+
+Snapshots are read-only, so you can render JSX from them, but mutations still need to be made via the store.
```
dai-shi, line 20: we want to emphasize, when reading in a function (like `handleClick`), we should read it from the proxy state.

btw, I don't think we (should) use the term `store` in docs.

docs/api/basic/useSnapshot.mdx
```diff
@@ -26,7 +31,65 @@ function Counter() {
...
+}
+```
+
+If book 2's title is changed, a new `snap` is created and the `AuthorView` and `BookView` components will re-render.
+
+Note if `BookView` is `React.memo`d, the 1st `BookView` will not re-render, b/c the 1st `Book` snapshot will be the same instance, as only the 2nd `Book` was mutated (the root `Author` snapshot will also be updated since the list of `books` has changed).
+
+### Child Components Making Mutations
+
+The above approach works if `BookView` is read-only; if your child component needs to make mutations, then you'll need to pass the store:
```
dai-shi, line 71: btw, we could pass both snap and state.

docs/api/basic/useSnapshot.mdx
```diff
@@ -26,7 +31,65 @@ function Counter() {
...
+
+function BookView({ book }) {
+  // book is a snapshot
+  return <div>{book.title}</div>
+}
+```
+
+If book 2's title is changed, a new `snap` is created and the `AuthorView` and `BookView` components will re-render.
+
+Note if `BookView` is `React.memo`d, the 1st `BookView` will not re-render, b/c the 1st `Book` snapshot will be the same instance, as only the 2nd `Book` was mutated (the root `Author` snapshot will also be updated since the list of `books` has changed).
```
dai-shi, line 67: maybe, we should have a special section about memo and deps. it's very important but missing in docs, I suppose (kind of my bad.)

dai-shi (commented):
I appreciate for your work! It's very very helpful.
I'd like to adjust something before merging.

docs/api/basic/useSnapshot.mdx
stephenh: Sounds good, I pushed an update with a "passing both" example.

docs/api/basic/useSnapshot.mdx
stephenh: Good point, I reworded + gave example of the callback using the proxy state.

> use the term store in docs.

Wdyt of the term state?

My hesitation about the term proxy is, ironically, the same hesitation you had wrt getters: that it seems like an advanced term/concept most users probably won't know about / shouldn't have to know about, as ideally/imo that Valtio uses proxies is just an implementation detail.

Vs. "store", which I think is generally widely used in state management circles for "the store of your state". So it's a more "logical"/concept based term than proxy's physical term.

...I would even pitch that in some Valtio 2.x, maybe `proxy` is renamed to `store`. :-D

("state" I think is too ambigious...do you mean my store state, or my snapshot state? Both the proxy and snapshot have state, it's just that one is mutable state and other immutable state. Hence like the store (mutable) + snapshot (immutable) pair.) 

That said, don't want that terminology suggestion to hold up the PR, so I've replaced all the "store"s with "proxy", unless you find one I've missed.

docs/api/basic/useSnapshot.mdx
stephenh: Yeah, I agree; fine to be a separate PR?

docs/api/basic/useSnapshot.mdx
dai-shi: I use "state" for valtio proxy objects and "snap" or "snapshot" for valtio snapshot objects.
I know "state" can be confusing and too general, though. We can use "proxy state" to be a little more explicit.

docs/api/basic/useSnapshot.mdx
dai-shi: okay.

docs/api/basic/useSnapshot.mdx
```diff
@@ -7,26 +7,116 @@ description: 'Create a local snapshot that catches changes.'
...
+      {snap.books.map((book, i) => (
+        <Book key={book.id} bookProxy={state.books[i]} bookSnapshot={book} />
+      ))}
     </div>
   )
 }
 ```
 
-#### Read only what you need
+There should be no performance difference between these two approaches.
```
dai-shi, line 117: Do you mean between passing snap and passing proxy state?
Both are render optimized, but the latter with React.memo would perform better with large list. Yeah, but without memo, it's basically no difference (the former might be a bit efficient?). Using memo is kind of advanced, because we can't pass snapshot objects to memoed components. So, maybe it's a different story.

dai-shi (approved):
looks good to me. can merge if you are fine.

PR #684, opened 2023-02-27
fix(react/utils): added useProxy parameter for opts #683

### LINKED ISSUE #683: allow `useProxy` to pass `useSnapshot` options
## Summary

`useProxy` only takes the store yet as argument, in case of input-elements a sync pass-through would also be required (ref https://github.com/pmndrs/valtio/#update-synchronously)

## Link to reproduction

https://stackblitz.com/edit/vitejs-vite-6d3nla?file=src%2FApp.jsx&terminal=dev

## Check List

steve-py96:

Fixes #683

## Summary

pass-through `useSnapshot` opts in `useProxy`

## Check List

- [x ] `yarn run prettier` for formatting code and docs

### DIFF
--- a/src/react/utils/useProxy.ts
+++ b/src/react/utils/useProxy.ts
@@ -24,7 +24,7 @@ import { useSnapshot } from 'valtio/react'
  */
 export function useProxy<T extends object>(
   proxy: T,
-  options?: Parameters<typeof useSnapshot>[1]
+  options?: NonNullable<Parameters<typeof useSnapshot>[1]>
 ): T {
   const snapshot = useSnapshot(proxy, options) as T
 
--- a/src/react/utils/useProxy.ts
+++ b/src/react/utils/useProxy.ts
@@ -24,7 +24,7 @@ import { useSnapshot } from 'valtio/react'
  */
 export function useProxy<T extends object>(
   proxy: T,
-  options?: Parameters<typeof useSnapshot<T>>[1]
+  options?: Parameters<typeof useSnapshot>[1]
 ): T {
   const snapshot = useSnapshot(proxy, options) as T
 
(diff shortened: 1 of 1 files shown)


src/react/utils/useProxy.ts
```diff
@@ -19,10 +19,14 @@ import { useSnapshot } from 'valtio/react'
...
  *
  * @param proxy
+ * @param options
  * @returns A new proxy which you can use in the render as well as in callbacks.
  */
-export function useProxy<T extends object>(proxy: T): T {
-  const snapshot = useSnapshot(proxy) as T
+export function useProxy<T extends object>(
+  proxy: T,
+  options?: Parameters<typeof useSnapshot<T>>[1]
```
steve-py96, line 27: for older ts versions probably without <T> for useSnapshot, I'll remove that

src/react/utils/useProxy.ts
```diff
@@ -19,10 +19,14 @@ import { useSnapshot } from 'valtio/react'
...
  *
  * @param proxy
+ * @param options
  * @returns A new proxy which you can use in the render as well as in callbacks.
  */
-export function useProxy<T extends object>(proxy: T): T {
-  const snapshot = useSnapshot(proxy) as T
+export function useProxy<T extends object>(
+  proxy: T,
+  options?: Parameters<typeof useSnapshot>[1]
```
dai-shi, line 27: To be precise, let's do this:
```suggestion
  options?: NonNullable<Parameters<typeof useSnapshot>[1]>
```

dai-shi (approved):
LGTM

src/react/utils/useProxy.ts
steve-py96: I don't mind going safe here ofc 👍🏿  (but since I wanna learn possible edge-cases and got not much experience in OSS (unlike you :P) I ask: Is `null` even a possible arg value here? When I f.e. try to assign null as Parameter it doesn't work)

dai-shi: It's to avoid `undefined`. There's no runtime risk to pass `undefined`. It's just to make the `useProxy` type same as `useSnapshot` type.

Issue #689, opened 2023-03-06
[DOCS]: wrong API url links


devagja:
I have seen that it happens in several parts of the documentation,
I am going to make a list of all of them and then a PR,

Thx for Valtio!

devagja:
Hello,
I have seen that it happens in several parts of the documentation,
I am going to make a list of all of them and then a PR,

thanks for Valtio!

devagja:
https://github.com/pmndrs/valtio/pull/691

PR #691, opened 2023-03-06
docs: fix links url

devagja:
https://github.com/pmndrs/valtio/issues/689

### DIFF
--- a/docs/api/basic/proxy.mdx
+++ b/docs/api/basic/proxy.mdx
@@ -59,7 +59,7 @@ authState.user.name = 'Nina'
 
 #### Promises in proxies
 
-See [`async`](/docs/guides/async) for more details.
+See [`async`](../../guides/async) for more details.
 
 ```jsx
 import { proxy } from 'valtio'
--- a/docs/api/basic/proxy.mdx
+++ b/docs/api/basic/proxy.mdx
@@ -59,7 +59,7 @@ authState.user.name = 'Nina'
 
 #### Promises in proxies
 
-See [`async`](/docs/advanced/async) for more details.
+See [`async`](/docs/guides/async) for more details.
 
 ```jsx
 import { proxy } from 'valtio'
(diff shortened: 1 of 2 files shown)


docs/api/basic/proxy.mdx
```diff
@@ -59,7 +59,7 @@ authState.user.name = 'Nina'
 
 #### Promises in proxies
 
-See [`async`](/docs/advanced/async) for more details.
+See [`async`](/docs/guides/async) for more details.
```
dai-shi, line 62: Can this be
```suggestion
See [`async`](../../guides/async.mdx) for more details.
```

dai-shi (commented):
Thanks for the catch.
We want to be able to browse docs in github.
Can you make all links relative paths?

dai-shi:
Hmm, adding extension broke the website.
@roguesherlock Is it possible to remove `.mdx` extension when deploying website?

roguesherlock:
@dai-shi Currently the renderer we use to parse the markdown doesn't support it but It should be possible to add a plugin to do that for us.

dai-shi:
> @dai-shi Currently the renderer we use to parse the markdown doesn't support it but It should be possible to add a plugin to do that for us.

Oh, that's unfortunate. I wonder if anyone would be interested in supporting it.
Otherwise, we would need to remove the extension for now...

devagja:
I am going to change the routes as at the beginning.
I can research how to fix it but if anyone else can investigate, I'd appreciate it.

dai-shi:
> I am going to change the routes as at the beginning.

If it's about a workaround, can it be relative paths without `.mdx`? It won't fix, but people can add the extension when 404.

devagja:
> If it's about a workaround, can it be relative paths without `.mdx`? It won't fix, but people can add the extension when 404.

sorry for the lapse of time,
I just added the relative paths without the `.mdx`

dai-shi (approved):
Looks good. Thanks!

PR #713, opened 2023-04-17
Allow passthrough of Redux devtools options

sarimarton:
The current 'devtools()' util in Valtio restricts the Redux-devtools option to the 'name' field, and a custom 'enabled' field. But it might be useful to pass other devtools options too to Redux-devtools. My current case, which motivated this, is that I'm having a large data set in the proxy store (as a ref() of course), and any change to the proxy causes a significant lag in redux-devtools. This has been performance measured in Chrome. The solution is to customize redux-devtools' serializer to skip the large data part, as I don't need to see that part. This way I can further debug other perf-related bottlenecks in my heavy app. During this effort, it seemed like Valtio unnecessarily restricted the redux-devtools options object, so this change only lifts this restriction.

Fixes #

## Check List

### DIFF
--- a/src/vanilla/utils/devtools.ts
+++ b/src/vanilla/utils/devtools.ts
@@ -1,5 +1,5 @@
 import { snapshot, subscribe } from '../../vanilla.ts'
-import type {} from '@redux-devtools/extension'
+import type { EnhancerOptions } from '@redux-devtools/extension'
 
 // FIXME https://github.com/reduxjs/redux-devtools/issues/1097
 type Message = {
@@ -10,9 +10,8 @@ type Message = {
 
 const DEVTOOLS = Symbol()
 
-type Options = {
+type Options = EnhancerOptions & {
   enabled?: boolean
-  name?: string
 }
 
 export function devtools<T extends object>(
@@ -67,7 +66,7 @@ export function devtools<T extends object>(
   }
 
   let isTimeTraveling = false
-  const devtools = extension.connect({ name })
+  const devtools = extension.connect({ name, ...options })
   const unsub1 = subscribe(proxyObject, (ops) => {
     const action = ops
       .filter(([_, path]) => path[0] !== DEVTOOLS)
--- a/src/vanilla/utils/devtools.ts
+++ b/src/vanilla/utils/devtools.ts
@@ -1,5 +1,5 @@
-import { snapshot, subscribe } from '../../vanilla.ts'
 import type { EnhancerOptions } from '@redux-devtools/extension'
+import { snapshot, subscribe } from '../../vanilla.ts'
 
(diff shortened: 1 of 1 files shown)


src/vanilla/utils/devtools.ts
```diff
@@ -1,5 +1,5 @@
+import type { EnhancerOptions } from '@redux-devtools/extension'
```
dai-shi, line 1: This produces types in dist. We would need to add it a peer dep.
There is a previous discussion, but the solution is probably copying&pasting types. Or just use `unknown`.

src/vanilla/utils/devtools.ts
```diff
@@ -67,7 +66,7 @@ export function devtools<T extends object>(
   }
 
   let isTimeTraveling = false
-  const devtools = extension.connect({ name })
+  const devtools = extension.connect({ name, ...options })
```
dai-shi, line 69: Shouldn't we remove `enabled` from `options`?

src/vanilla/utils/devtools.ts
sarimarton: Indeed. I would lean towards peer dep, it can't hurt...

src/vanilla/utils/devtools.ts
sarimarton: I agree, I didn't actually figure out exactly why it's there.

dai-shi: `enabled` is used in our lib, right?

dai-shi:
While the idea is good, this PR needs more work and it's incomplete.
Closing as stale for now.
Feel free to add comments.

sarimarton:
Wouldn't it be better merging it in with a @todo flag? I'm not sure what problems need to be solved. Even with no extra typing, just as an escape hatch...

dai-shi:
The two things are really something addressed. Especially, the first one is not acceptable as it caused problems previously. The second one is easy to fix without todo flag.

dai-shi:
I will modify the branch.

src/vanilla/utils/devtools.ts
```diff
@@ -10,10 +10,14 @@ type Message = {
 
 const DEVTOOLS = Symbol()
 
+type EnhancerOptions = Parameters<
+  NonNullable<(typeof window)['__REDUX_DEVTOOLS_EXTENSION__']>['connect']
```
dai-shi, line 14: This idea clicked me.

dai-shi (approved):
Should be good to go.

src/vanilla/utils/devtools.ts
sarimarton: Awesome!

PR #736, opened 2023-06-09
fix: invalidate snap cache when property is removed from object

### LINKED ISSUE #712: Cache key could refer previous value
## Summary

The cache key is referred to the previous value when the shape of references in the cache key is the same.

## Link to reproduction

https://jsfiddle.net/wxq64cdn/

Minimum reproduction cases represent proxy in array proxy.

Here's my debugging context that may help:

It seems to be no cache invalidation when the 'remove' operation for the array.

In minimum repro. case, at 40 lines, the proxy and value are cached keys as a's proxy.

When handle() function is called, the array will empty. After that, the value proxy is updated  (not reference.) and pushed again to a.

The result of the 'a' still refers to the same cache that has been made by 40 lines. Because it refers to the same shape of references that is used as the cache key.

## Check List

Juyeong-Byeon:

related to  
https://github.com/pmndrs/valtio/issues/712

Just like that change [1], I suggest removing snap cache when props are removed from the parent object or array

[1] [This change](https://github.com/pmndrs/valtio/pull/599/files) is removing the listeners for removed props from the parent object or array.

FYI: #712 is an issue that has occurred since version 1.8.0  caused by https://github.com/pmndrs/valtio/commit/ec445576274f6b1c53c86099846592b5550e24fd or https://github.com/pmndrs/valtio/pull/599

Fixes #712

## Summary
 Adding remove snap cache logic to `deleteProperty` function

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -244,6 +244,7 @@ const buildProxyFunction = (
     const handler: ProxyHandler<T> = {
       deleteProperty(target: T, prop: string | symbol) {
         const prevValue = Reflect.get(target, prop)
+        snapCache.delete(target as Object)
         removePropListener(prop)
         const deleted = Reflect.deleteProperty(target, prop)
         if (deleted) {
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -244,7 +244,6 @@ const buildProxyFunction = (
     const handler: ProxyHandler<T> = {
       deleteProperty(target: T, prop: string | symbol) {
         const prevValue = Reflect.get(target, prop)
-        snapCache.delete(target as Object)
         removePropListener(prop)
         const deleted = Reflect.deleteProperty(target, prop)
         if (deleted) {
--- a/tests/snapshot.test.ts
+++ b/tests/snapshot.test.ts
@@ -1,7 +1,12 @@
 import { describe, expect, it } from '@jest/globals'
 import { createProxy, getUntracked } from 'proxy-compare'
 import { TypeEqual, expectType } from 'ts-expect'
-import { INTERNAL_Snapshot as Snapshot, proxy, snapshot } from 'valtio'
+import {
+  INTERNAL_Snapshot as Snapshot,
+  proxy,
+  snapshot,
+  subscribe,
+} from 'valtio'
 
 const sleep = (ms: number) =>
   new Promise((resolve) => {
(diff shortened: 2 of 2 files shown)


src/vanilla.ts
```diff
@@ -244,6 +244,7 @@ const buildProxyFunction = (
     const handler: ProxyHandler<T> = {
       deleteProperty(target: T, prop: string | symbol) {
         const prevValue = Reflect.get(target, prop)
+        snapCache.delete(target as Object)
```
dai-shi, line 247: We never delete the cache elsewhere, so it doesn't feel very nice.
Do you happen to know why deleteProperty doesn't increment the version number?
I think we should increment it if it doesn't.

dai-shi (changes requested):
Would you be able to add a failing test?

src/vanilla.ts
Juyeong-Byeon: Ok, I will try more on that way.
And I found out that although if I increase the target version of deleted property in `deleteProperty` it is still getting the same version in [here](https://github.com/pmndrs/valtio/blob/d1da1b234ce8d4b0460a14d83e16a8cfa5a94bcd/src/vanilla.ts#L105-L114).

Juyeong-Byeon: > Would you be able to add a failing test?
Do you mean that adding a test fails when [this symptom]((https://github.com/pmndrs/valtio/issues/712) has occurred?

dai-shi: ah, yeah. adding a test to represent #712 issue, which fails in the current main branch, and will be resolved with this PR.

dai-shi: > it is still getting the same version

Is it like reading a stale version from somewhere?

tests/snapshot.test.ts
```diff
@@ -69,6 +74,41 @@ it('should create a new proxy from a snapshot', async () => {
...
 })
 
+it('should return updated property snapshot when property is updated after removed from Array and added to array again', async () => {
+  /**
+   * error occurs when
+   * [1]. Remove target item should be in the array
+   * [2]. Remove target item should be nested object
+   * [3]. The array should be subscribed
+   * [4]. The array should pass to snapshot() before the target item removed
+   */
```
Juyeong-Byeon, line 84: @dai-shi I added test and reverted the previous commit.
And wrote the error occurrence conditions.

dai-shi: Thanks. Can you resolve conflicts and merge the main branch?

Juyeong-Byeon: Done! And I will look in to it about another approach that you suggested. 😀

dai-shi: Hi, are you still around?

Juyeong-Byeon: I made some tryies after this but it didn't worked. (ex- force version update if child object removed)
I think It will be hard to fix without breaking something as you said.

Do you have any idea about approach to fixit?

dai-shi: https://github.com/pmndrs/valtio/issues/712#issuecomment-1643355452 is my current understanding.

dai-shi:
Closing as inactive. Let's reopen when there's a progress.

Issue #749, opened 2023-07-09
Snapshots can be mutated by adding or removing properties


pastelmind:
## Summary

Adding or removing properties from a snapshot object or array does not throw, even in strict mode.

```js
'use strict' // Must be enabled in browser scripts / Node CommonJS

const state = proxy([5, 10, 15])
const snap = snapshot(state)

// snap[0] = 100 // Throws
snap[10] = 100 // Does not throw
snap.push(100) // Does not throw
snap.pop() // Does not throw

const state2 = proxy({ foo: 1 })
const snap2 = snapshot(state2)

// snap2.foo = 100 // Throws
snap2.bar = 2 // Does not throw
```

Although it appears that such invalid operations do not affect the proxy store itself, ~~silently swallowing~~ <ins>allowing</ins> these errors is undesirable, since the snapshot (correctly) errors on other types of mutations.

<ins>Edit: The snapshot _can_ be mutated:</ins>

```js
const p = proxy({ foo: 1 });
const s = snapshot(p);
console.log(s); // { foo: 1 }
s.bar = 2;
console.log(s); // { foo: 1, bar: 2 }
delete s.foo;
console.log(s); // { bar: 2 }
```

This seems to be a correctness issue.

## Link to reproduction

https://codesandbox.io/s/sad-gauss-w364gf

## Check List

Please do not ask questions in issues.

- [ ] I've already opened a [discussion](https://github.com/pmndrs/valtio/discussions) before opening this issue, or already discussed in other media.

Please include a minimal reproduction.

- [x] I've added a link to a [typescript playground](https://www.typescriptlang.org/play) or [codesandbox](https://codesandbox.io) with a minimal reproduction.

pastelmind:
Preventing the addition of new properties should be trivial. We can call `Object.preventExtensions()` on the snapshot object here, just before returning it [here](https://github.com/pmndrs/valtio/blob/5d0eca7edc25aa7f17fb8082ef14244a35fc12b2/src/vanilla.ts#L149):

```diff
diff --git a/src/vanilla.ts b/src/vanilla.ts
index 7844994..1e169a0 100644
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -146,6 +146,7 @@ const buildProxyFunction = (
       }
       Object.defineProperty(snap, key, desc)
     })
+    Object.preventExtensions(snap)
     return snap
   },
```

I tested this locally and found no breaking tests. I can submit a PR if you want.

[...]

PR #752, opened 2023-07-10
fix(vanilla): Prevent adding new property on snapshot

pastelmind:

Partially deals with #749 by disallowing property addition
(but does not completely fix it, since we still allow property removal)

## Summary

As a performance optimization, we currently mark [properties of snapshots as configurable](https://github.com/pmndrs/valtio/blob/5d0eca7edc25aa7f17fb8082ef14244a35fc12b2/src/vanilla.ts#L128C10-L130). This allows users to inadvertently add or remove properties on a snapshot object with no warning. Although TypeScript typings do prevent this to a degree, it is not foolproof.

While we currently do not have a robust solution that disallows property removal while maintaining the current level performance, we can prevent addition of new properties on the snapshot by calling `Object.preventExtensions()` on it.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -146,6 +146,7 @@ const buildProxyFunction = (
       }
       Object.defineProperty(snap, key, desc)
     })
+    Object.preventExtensions(snap)
     return snap
   },
 
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -146,8 +146,7 @@ const buildProxyFunction = (
       }
       Object.defineProperty(snap, key, desc)
     })
-    Object.preventExtensions(snap)
-    return snap
+    return Object.preventExtensions(snap)
   },
 
   proxyCache = new WeakMap<object, ProxyObject>(),
--- a/tests/snapshot.test.ts
+++ b/tests/snapshot.test.ts
@@ -1,7 +1,7 @@
 import { createProxy, getUntracked } from 'proxy-compare'
 import { TypeEqual, expectType } from 'ts-expect'
-import { INTERNAL_Snapshot as Snapshot, proxy, snapshot } from 'valtio'
 import { describe, expect, it } from 'vitest'
+import { INTERNAL_Snapshot as Snapshot, proxy, snapshot } from 'valtio'
 
 const sleep = (ms: number) =>
   new Promise((resolve) => {
--- a/tests/snapshot.test.ts
+++ b/tests/snapshot.test.ts
@@ -1,7 +1,7 @@
 import { createProxy, getUntracked } from 'proxy-compare'
 import { TypeEqual, expectType } from 'ts-expect'
-import { describe, expect, it } from 'vitest'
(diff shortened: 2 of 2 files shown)


dai-shi:
Seems like #519 codesandbox works well too.

src/vanilla.ts
```diff
@@ -146,6 +146,7 @@ const buildProxyFunction = (
       }
       Object.defineProperty(snap, key, desc)
     })
+    Object.preventExtensions(snap)
     return snap
```
dai-shi, line 150: Just a little bit compress.
```suggestion
    return Object.preventExtensions(snap)
```

tests/snapshot.test.ts
```diff
@@ -1,7 +1,7 @@
 import { createProxy, getUntracked } from 'proxy-compare'
 import { TypeEqual, expectType } from 'ts-expect'
-import { describe, expect, it } from 'vitest'
 import { INTERNAL_Snapshot as Snapshot, proxy, snapshot } from 'valtio'
+import { describe, expect, it } from 'vitest'
```
dai-shi, line 4:
```suggestion
import { describe, expect, it } from 'vitest'
import { INTERNAL_Snapshot as Snapshot, proxy, snapshot } from 'valtio'
```

dai-shi (approved):
Thanks for your contribution!

dai-shi:
I'll make this to mark #749 as closed as nothing else is actionable. Please open [a new discussion](https://github.com/pmndrs/valtio/discussions/new) for more implementation ideas.

lovetingyuan:
this change cause an error, call ownKeys in Proxy object, `ownKeys target is non-extensible but key is missing from trap result`

pastelmind:
Could you elaborate? Please provide some sample code so we can understand it.

lovetingyuan:
> Could you elaborate? Please provide some sample code so we can understand it.

I use valtio in expo project, I think the problem may be related to `[Hermes](https://reactnative.dev/docs/hermes)` engine? I am not sure.
Here is example repo: https://github.com/lovetingyuan/expo-example

setup development environment for expo, then run `npm run android`, click the detail button

lovetingyuan:
when I add `"jsEngine": "jsc",` in `app.json`, there is no error. So this issue is highly possible related to hermes engine.

lovetingyuan:
I create an issue in hermes repo https://github.com/facebook/hermes/issues/1063, you can take a look.

pastelmind:
If this is an engine bug in Hermes, I'm OK with reverting the change for now while we investigate the issue.

pastelmind:
From https://github.com/pmndrs/valtio/discussions/764 reported by @afkcodes, I hypothesize that Hermes may be implementing `Object.preventExtensions()` in such a way that it also makes existing properties non-writable and non-configurable, a la `Object.freeze()`. This could be wrong ofc. But if true, then we may have no choice but to remove `Object.preventExtensions()`.

afkcodes:
This i  have validated and have found that this is true, i have my state typed, thanks @pastelmind

PR #760, opened 2023-07-12
feat(vanilla): Support class fields (defineProperty)

pastelmind:
## Related Issues or Discussions

Fixes #759 (the feature request at the end)

## Summary

Adds support for ES2022 class fields created with `Object.defineProperty()` semantics. When a configurable, enumerable, and writable (CEW) own property (usually a class field) is created on a proxy object that does not have the property, or has a CEW own property of the same name, treat it like a [[set]] trap.

## Check List

- [x] `yarn run prettier` for formatting code and docs

PR #766, opened 2023-07-24
Revert "feat(vanilla): Support class fields (defineProperty) (#760)"

### LINKED ISSUE #765: defineProperty trap makes property non-CEW in React Native (Hermes engine)
## Summary

This issue describes the bug reported in https://github.com/pmndrs/valtio/pull/752#issuecomment-1644968041 and #764. In both cases, Valtio was added to a React Native project with the Hermes engine enabled. Assigning a new value to a property on a proxy store object marked the property as non-configurable and non-writable, effectively "freezing" the property and preventing future mutations.

The source of the bug was initially attributed to #752. However, I suspect that the actual cause is #760. My own investigations (https://github.com/pmndrs/valtio/discussions/764#discussioncomment-6517634 and https://github.com/pmndrs/valtio/discussions/764#discussioncomment-6521744) indicate that the bug occurs on 1.11.0 (with #760), but not on 1.10.7 (with #752) or 1.10.6. Also, the bug affects proxy objects (affected by #760) rather than their snapshots (affected by #752).

The root cause appears to be a bug in the Proxy implementation of Hermes. I opened a bug report (https://github.com/facebook/hermes/issues/1065) and am waiting for confirmation from the Hermes devs.

Meanwhile, it would be nice if we handled this ourselves. Hermes and React Native are large projects and move slower than the rest of the JS ecosystem. Even if Hermes fixes the problem immediately (if at all), it would take a long time for the fix to propagate to downstream projects. Many people are using React Native today and we need to support them too.

## Link to reproduction

See following discussion comments

- https://github.com/pmndrs/valtio/discussions/764#discussioncomment-6517634
- https://github.com/pmndrs/valtio/discussions/764#discussioncomment-6521744

## Check List

- [x] I've already opened a [discussion](https://github.com/pmndrs/valtio/discussions) before opening this issue, or already discussed in other media.

- [ ] I've added a link to a [typescript playground](https://www.typescriptlang.org/play) or [codesandbox](https://codesandbox.io) with a minimal reproduction.

pastelmind:
This reverts commit c0bda5afd2780fb929656a87bcc9e16bcf189fd8.

Due to [what I presume is a bug in the Hermes engine](https://github.com/facebook/hermes/issues/1065), React Native projects cannot use Valtio 1.11.0. Let's revert #760 to support existing React Native users.

Fixes #765
May fix #764 (but we need confirmation from actual RN users)

## Summary

This reverts #760 (c0bda5afd2780fb929656a87bcc9e16bcf189fd8). The following pattern (described in detail at #759) is no longer explicitly supported:

```js
class ValtioProxy {
  constructor() {
    return proxy(this)
  }
}

class MyStore extends ValtioProxy {
  field1 = 'asdf' // OK
  field2 = []     // This works in v1.11.0.
                  // But when this PR is merged,
                  // the array will no longer be reactive
}

const store = new MyStore()
store.field2.push(1) // Will re-render views in v1.11.0
                     // but not after this PR is merged
```

To work around this, we can explicitly set the field in the constructor, which properly invokes the `set` proxy trap:

```js
class MyStore extends ValtioProxy {
  constructor() {
    super()
    this.field2 = [] // This will be reactive
  }
}
```

Alternatively, you can configure your transpiler to use 'set' semantics for class fields (e.g. [`useDefineForClassFields: false`](https://www.typescriptlang.org/tsconfig#useDefineForClassFields). (Personally I advise against this, because it deviates from the modern ECMAScript spec)

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -86,9 +86,6 @@ const buildProxyFunction = (
     !(x instanceof RegExp) &&
     !(x instanceof ArrayBuffer),
 
-  shouldTrapDefineProperty = (desc: PropertyDescriptor) =>
-    desc.configurable && desc.enumerable && desc.writable,
-
   defaultHandlePromise = <P extends Promise<any>>(
     promise: P & {
       status?: 'pending' | 'fulfilled' | 'rejected'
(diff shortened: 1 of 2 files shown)


dai-shi:
> May fix https://github.com/pmndrs/valtio/discussions/764 (but we need confirmation from actual RN users)

Is it now confirmed with https://github.com/pmndrs/valtio/discussions/764#discussioncomment-6531628 ?

dai-shi (approved):
The change looks good.

pastelmind:
> > May fix #764 (but we need confirmation from actual RN users)
> 
> Is it now confirmed with [#764 (reply in thread)](https://github.com/pmndrs/valtio/discussions/764#discussioncomment-6531628) ?

I believe so.

ivankdev:
Vote for this, when it could be merged?

td-tomasz-joniec:
Is it worth reverting? Maybe the bug will be fixed in Hermes earlier.

dai-shi:
It makes sense to revert as we didn't know such a limitation when we merged and released.
I don't disagree that this is a reasonable addition, but my assumption was it didn't hurt at all.

https://github.com/pmndrs/valtio/pull/760#pullrequestreview-1538331597
> Not sure if there are pitfalls, but this should be harmless for normal (plain objects) usage, except for the bundle size increase.

So, there is an unexpected pitfall and it's harmful for some users.

dominiczaq:
yeah, and there are a lot of ppl stuck in certain RN versions. And there is
no timeframe for fixing this in hermes and even if it's fixed then it would
still require people to upgrade to latest version

PR #771, opened 2023-08-03
docs: fix typo in `getting-started.mdx`

shairez:
## Summary
I think it's suppose to say "reactivity", but I might be wrong

## Check List


dai-shi (approved):
Not 100%sure, but probably so.

PR #777, opened 2023-08-11
fix(utils): improve devtools typing to avoid depending on @redux-devtools/extension

### LINKED ISSUE #776: tsc fails when using functions in valitio/utils
## Summary

I got the following tsc error when using functions in `valtio/utils`.

```ts
import { proxy } from "valtio";
import { derive } from "valtio/utils";

const state = proxy({ count: 0 });
const derived = derive(
  {
    double: (get) => get(state).count * 2,
  },
  { proxy: state }
);
```

```console
$ tsc && vite build
node_modules/valtio/vanilla/utils/devtools.d.ts:1:63 - error TS2339: Property '__REDUX_DEVTOOLS_EXTENSION__' does not exist on type 'Window & typeof globalThis'.
```

This issue occurs with `v1.10.6` (does not occur with `v1.10.5`)
I guess #713 causes this error and the root cause is the same as https://github.com/pmndrs/zustand/issues/1205..

## Link to reproduction

https://codesandbox.io/p/sandbox/modest-curran-ptlrl7

## Check List

fukumasuya:

Fixes #776

## Summary

I found the same issue in zustand(https://github.com/pmndrs/zustand/issues/1205#issuecomment-1221270524), so I added the same workaround(copied the type defs from `@redux-devtools/extension`)
I'm not sure if this way is a desired way, but I did it anyway..

## Check List

### DIFF
--- a/src/vanilla/utils/devtools.ts
+++ b/src/vanilla/utils/devtools.ts
@@ -1,76 +1,6 @@
 import { snapshot, subscribe } from '../../vanilla.ts'
 import type {} from '@redux-devtools/extension'
 
-// Copy types to avoid import type { Config } from '@redux-devtools/extension'
-// https://github.com/pmndrs/valtio/issues/776
-type Action<T = any> = {
-  type: T
-}
-type ActionCreator<A, P extends any[] = any[]> = {
-  (...args: P): A
-}
-type EnhancerOptions = {
-  name?: string
-  actionCreators?:
-    | ActionCreator<any>[]
-    | {
-        [key: string]: ActionCreator<any>
-      }
-  latency?: number
-  maxAge?: number
-  serialize?:
-    | boolean
-    | {
-        options?:
-          | undefined
-          | boolean
-          | {
-              date?: true
-              regex?: true
-              undefined?: true
-              error?: true
-              symbol?: true
-              map?: true
-              set?: true
-              function?: true | ((fn: (...args: any[]) => any) => string)
-            }
-        replacer?: (key: string, value: unknown) => any
-        reviver?: (key: string, value: unknown) => any
-        immutable?: any
-        refs?: any
-      }
-  actionSanitizer?: <A extends Action>(action: A, id: number) => A
-  stateSanitizer?: <S>(state: S, index: number) => S
-  actionsBlacklist?: string | string[]
-  actionsWhitelist?: string | string[]
-  actionsDenylist?: string | string[]
-  actionsAllowlist?: string | string[]
-  predicate?: <S, A extends Action>(state: S, action: A) => boolean
-  shouldRecordChanges?: boolean
-  pauseActionType?: string
-  autoPause?: boolean
-  shouldStartLocked?: boolean
-  shouldHotReload?: boolean
-  shouldCatchErrors?: boolean
-  features?: {
-    pause?: boolean
-    lock?: boolean
-    persist?: boolean
-    export?: boolean | 'custom'
-    import?: boolean | 'custom'
-    jump?: boolean
-    skip?: boolean
-    reorder?: boolean
-    dispatch?: boolean
-    test?: boolean
-  }
-  trace?: boolean | (<A extends Action>(action: A) => string)
-  traceLimit?: number
-}
-type Config = EnhancerOptions & {
-  type?: string
-}
-
 // FIXME https://github.com/reduxjs/redux-devtools/issues/1097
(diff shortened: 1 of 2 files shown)


fukumasuya:
It might be a better solution to move `@redux-devtools/extension` from `devDependencies` to `dependencies` and import `Config` type  from `@redux-devtools/extension` .. But it might seem a little odd that it's in `dependencies` and `redux` is a `peerDependencies` of `@redux-devtools/extension` and devtools imports redux, so valtio users need to install redux manually...

dai-shi (commented):
So, it's basically doing the same as https://github.com/pmndrs/zustand/pull/1207 ?
It looks okay then.

src/vanilla/utils/devtools.ts
```diff
@@ -10,14 +80,10 @@ type Message = {
 
 const DEVTOOLS = Symbol()
 
-type EnhancerOptions = Parameters<
-  NonNullable<(typeof window)['__REDUX_DEVTOOLS_EXTENSION__']>['connect']
->[0]
```
dai-shi, line 15: Though, this looks better. Curious how this is failing?

dai-shi: Oh, I see. We still need to install the package.
I wonder if we can type check on it.

dai-shi: Curious why L124 doesn't cause a type error.

dai-shi: because it disappears, okay.

dai-shi (approved):
208d3c9 fixes #776 reproduction.

fukumasuya:
> So, it's basically doing the same as https://github.com/pmndrs/zustand/pull/1207 ?

Yes. that's right.

I also confirmed https://github.com/pmndrs/valtio/pull/777/commits/208d3c9b1d6ee655d0ac332625204fd0e8477c6d fixes #776. thanks for updating my fix!

But it looks #713 feature does not work with TypeScript.. (tsc fails if we add some parameters that is coming from `@redux-devtools/extensions`.. is that ok for us?

```ts
const state = proxy({ count: 0, text: 'hello' });
devtools(state, { name: 'state name', enabled: true, latency: 1000 }); // 'latency' is coming from `@redux-devtools/extensions`
```

```console
  Overload 1 of 2, '(proxyObject: { count: number; text: string; }, options?: { enabled?: boolean | undefined; name?: string | undefined; } | undefined): (() => void) | undefined', gave the following error.
    Argument of type '{ name: string; enabled: true; latency: number; }' is not assignable to parameter of type '{ enabled?: boolean | undefined; name?: string | undefined; }'.
      Object literal may only specify known properties, and 'latency' does not exist in type '{ enabled?: boolean | undefined; name?: string | undefined; }'.
  Overload 2 of 2, '(proxyObject: { count: number; text: string; }, name?: string | undefined): (() => void) | undefined', gave the following error.
    Argument of type '{ name: string; enabled: boolean; latency: number; }' is not assignable to parameter of type 'string'.

15 devtools(state, { name: 'state name', enabled: true, latency: 1000 });
```

fukumasuya:
It seems if we install `@redux-devtools/extension` and import some function  in this or load the type def file, then `__REDUX_DEVTOOLS_EXTENSION__` is declared and following code works without type errors.., In that case, we need to install `redux` additionally because `redux` is in peerDependencies of `@redux-devtools/extension`..

```ts
const state = proxy({ count: 0, text: 'hello' });
devtools(state, { name: 'state name', enabled: true, latency: 1000 }); // 'latency' is coming from `@redux-devtools/extensions`
```

tsconfig.json
```json
{
  "include": [
    "./node_modules/@redux-devtools/extension/lib/types/index.d.ts",
  ]
}
```

dai-shi:
Hmm, isn't it just enough to install `@redux-devtools/extension`?
In that case, we should update the docs to explicitly import it (both in valtio and zustand.)
```ts
import type {} from '@redux-devtools/extension'
import { devtools } from 'valtio/utlis'
```

fukumasuya:
> Hmm, isn't it just enough to install @redux-devtools/extension?

It seems If we set `skipLibCheck` to `true`, we don't need to install `redux`.
Anyway, https://github.com/pmndrs/valtio/pull/777/commits/208d3c9b1d6ee655d0ac332625204fd0e8477c6d fixes #776 and is fine for me 👍 thanks!

PR #794, opened 2023-10-03
test: add cyclical class test

Gregoor:
Hi there,

thanks for the great library!

I've got a case of mildly cursed cyclical classes, that I don't think I can data model my way around. valtio seems like a great solution to make it work, but I seem to have run into a proxy-ing bug which I've been able to repro here (the test case succeeds without the `proxy()`-wrap).

Now the docs do say

> Not everything can be proxied. Generally, you are safe if it is serializable. Classes can also be proxied. But avoid special objects.

And it's not serializable and it's definitely special (the bad kind of special), but as I said, it also seems necessary for my use case.

So here's my question: would fixing this be possible and in-scope? If so I'd also be happy to see if I can figure it out on my own, though guidance would certainly help.

Cheers!

### DIFF
--- a/tests/class.test.tsx
+++ b/tests/class.test.tsx
@@ -1,6 +1,6 @@
 import { StrictMode, useEffect, useRef } from 'react'
 import { fireEvent, render, waitFor } from '@testing-library/react'
-import { it } from 'vitest'
+import { expect, it } from 'vitest'
 import { proxy, useSnapshot } from 'valtio'
 
 it('simple class without methods', async () => {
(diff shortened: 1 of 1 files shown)


dai-shi:
Thanks for opening up a discussion.

> would fixing this be possible and in-scope?

One thing we can't change is we kind of support classes by not touching prototype chain.
So, leaving `prototype` untouched is the design decision.
tbh, I'm not sure if it's fixable under this constraint.

> If so I'd also be happy to see if I can figure it out on my own, though guidance would certainly help.

I'd encourage you to dig a little deeper. I don't yet know why it doesn't work. We do support cyclical objects, which is a unique feature to my knowledge.

Gregoor:
I did not manage to figure it out and in the scramble I managed to unscramble my data model. So I'll close this PR. Thanks for the words and the lib! Also for Jotai, which I'm going back to now, since it indeed was the only solid lib I found that does cyclical objects.

PR #800, opened 2023-10-19
fix(vanilla): respect property enumetrability

alexander-entin:

Fixes #726

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -125,8 +125,10 @@ const buildProxyFunction = (
         return
       }
       const value = Reflect.get(target, key)
-      const { enumerable = false } =
-        Reflect.getOwnPropertyDescriptor(target, key) as PropertyDescriptor
+      const { enumerable = false } = Reflect.getOwnPropertyDescriptor(
+        target,
+        key
+      ) as PropertyDescriptor
       const desc: PropertyDescriptor = {
         value,
         enumerable,
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -125,8 +125,8 @@ const buildProxyFunction = (
         return
       }
       const value = Reflect.get(target, key)
-      const { enumerable = true } =
-        Reflect.getOwnPropertyDescriptor(target, key) ?? {}
+      const { enumerable = false } =
+        Reflect.getOwnPropertyDescriptor(target, key) as PropertyDescriptor
       const desc: PropertyDescriptor = {
         value,
         enumerable,
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -125,13 +125,13 @@ const buildProxyFunction = (
         return
       }
       const value = Reflect.get(target, key)
-      const { enumerable = false } = Reflect.getOwnPropertyDescriptor(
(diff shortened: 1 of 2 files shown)


src/vanilla.ts
```diff
@@ -125,9 +125,11 @@ const buildProxyFunction = (
         return
       }
       const value = Reflect.get(target, key)
+      const { enumerable = true } =
+        Reflect.getOwnPropertyDescriptor(target, key) ?? {}
```
dai-shi, line 129: I think we are sure that it's not undefined.
```suggestion
      const desc = Object.getOwnPropertyDescriptor(
        target,
        key
      ) as PropertyDescriptor
```

Do we want to make it `true` if it's undefined?

dai-shi (commented):
Please also check CI errors.

src/vanilla.ts
alexander-entin: 1. As for getting property descriptor.
We do not expect `undefined` and all we need is to make TypeScript happy.
For this `as PropertyDescriptor` is indeed clearer than `?? {}`

2. As for using the descriptor just got instead of creating a new one from scratch.
No, this messes `value`, `get`, `set` and fails some tests.

3. As for `enumerable` default value.
It looks like `getOwnPropertyDescriptor` always returns `enumerable` set.
I don't know if this is guaranteed by the spec.
At the same time `defineProperty` has its own defaults.
I think we can assume that passing whatever is returned by `getOwnPropertyDescriptor` to `defineProperty` will do.
So, we don't actually need to provide any default.
Again, we only need to make TypeScript happy.
I'm not familiar with TypeScript enough to propose a type-only solution.
Maybe, using the same default as `defineProperty` is OK:
````javascript
const { enumerable = false } = ...
````

File: src/vanilla.ts [hunk 2]
```diff
@@ -125,9 +125,13 @@ const buildProxyFunction = (
         return
       }
       const value = Reflect.get(target, key)
+      const { enumerable = false } = Reflect.getOwnPropertyDescriptor(
```
dai-shi, line 128: If `undefined` is acceptable by TypeScript, let's go with it. (We just avoid using the knowledge what is the default.)
```suggestion
      const { enumerable } = Reflect.getOwnPropertyDescriptor(
```

dai-shi: Maybe not? https://github.com/pmndrs/valtio/pull/800#discussion_r1366825175 seems to  say so.

alexander-entin:
````javascript
const { enumerable } = Reflect.getOwnPropertyDescriptor(
        target,
        key
      ) as PropertyDescriptor
      const desc: PropertyDescriptor = {
        value,
        enumerable: enumerable as boolean,
        // This is intentional to avoid copying with proxy-compare.
        // It's still non-writable, so it avoids assigning a value.
        configurable: true,
      }
````
As I understand, this is correct from JavaScript view and doesn't rely on `defineProperty` defaults.
`as boolean` silences TypeScript (although, is not strictly correct type-wise)

dai-shi (approved):
LGTM. I think the TS hack is reasonable.

PR #809, opened 2023-11-05
fix(utils): useProxy to avoid re-renders when no props are touched

dai-shi:

Fixes https://github.com/pmndrs/valtio/discussions/808#discussioncomment-7476506

## Summary

Touch a dummy prop so that it avoids re-renders if no other props are touched.
I'm not sure if we should add this hack, or just add a caveat note in docs.

## Check List


dai-shi:
I think this change is probably reasonable, because useProxy only takes care of top level property access.

PR #816, opened 2023-11-13
refactor: revisit prettier config to be minimal

Aslemammad:
> The prettier config was defined back in some time, and prettier has been updated since then.
I think we should follow the default config as much as possible.

https://github.com/pmndrs/zustand/pull/2193


dai-shi (approved):
Thanks for your work!

Issue #507, opened 2022-07-26
Promise support in watch


Akkuma:
https://codesandbox.io/s/breaking-watch-ct246n

If you attempt to use async/await in watch it'll break the cleanup cycle. I added a quick test locally and fooled around with it and can avoid the issue with something along these lines:
```
      // If there's a cleanup, we add this to the cleanups set
      if (cleanupReturn && !(cleanupReturn instanceof Promise)) {
        cleanups.add(cleanupReturn)
      }
 ```
However, I'm not sure that's the desired behavior either as that would also mean you can't run a cleanup function using async/await without additional logic, but maybe that's ok.

Derive handles this scenario already of working with promises, so it feels a little odd that watch is the odd function out when it comes to promises. I suppose one workaround is doing a manual subscribe dance around the promises and their subscriptions.

**Comment by dai-shi** [comment]
Thanks for reporting.
I'm not sure if it's correct to allow async function. `watch` is a wrapper around `subscribe` and `subscribe` doesn't support it, right?
`useEffect` doesn't allow async function either.

> Derive handles this scenario already of working with promises

Does `derive` handles promises? Hm, we don't have cleanup functions but `underive` instead, so its design is different (because we want `derive` to return a proxy).

**Comment by Akkuma** [comment]
Here's derive using async/await https://codesandbox.io/s/derive-with-promises-yx2fkq?file=/src/App.js and here's subscribe using async/await https://codesandbox.io/s/subscribe-with-promises-7pokmq

I think the key difference is that `watch` is the only function that allows for an explicit cleanup function returned from it. The code for `subscribe` doesn't expect or check for any return value from the callback itself I've already figured out an ok alternative to work with `watch`, by deferring async/await to another function something like:
```
watch((get) => {
  const asyncThing = get(state).asyncThing
  const asyncStuff = get(state2).asyncStuff
  
  doAsyncAwait(asyncThing, asyncStuff)
}
```

dai-shi:
https://codesandbox.io/s/derive-with-promises-yx2fkq?file=/src/App.js
This is interesting. Thanks for sharing. It's valid because valtio supports promise values.

https://codesandbox.io/s/subscribe-with-promises-7pokmq
Okay, right. `subscribe` callback returns void. So, it's fine.

`watch` callback should return `cleanup` or `void`.
I guess it's because `watch` does subscribe/unsubscribe automatically. @LXSMNSYC ?
And, cleanup function must not be a promise.

[...]

PR #825, opened 2023-11-18
feat: Support promise in watch util

LiamMartens:
https://github.com/pmndrs/valtio/issues/507

## Summary
Updates the watch util to support a promise callback. Proxies are subscribed to immediately to improve the time to first callback. This way a proxy can be subscribed to before the promise finishes fully.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -357,7 +357,7 @@ export function getVersion(proxyObject: unknown): number | undefined {
 
 export function subscribe<T extends object>(
   proxyObject: T,
-  callback: (ops: Op[]) => void,
+  callback: (ops: Op[]) => void | Promise<void>,
   notifyInSync?: boolean,
 ): () => void {
   const proxyState = proxyStateMap.get(proxyObject as object)
--- a/src/vanilla/utils/watch.ts
+++ b/src/vanilla/utils/watch.ts
@@ -2,12 +2,18 @@ import { subscribe } from '../../vanilla.ts'
 
 type Cleanup = () => void
 type WatchGet = <T extends object>(proxyObject: T) => T
-type WatchCallback = (get: WatchGet) => Cleanup | void | undefined
+type WatchCallback = (
+  get: WatchGet,
+) => Cleanup | void | Promise<void> | Promise<Cleanup> | undefined
 type WatchOptions = {
   sync?: boolean
 }
 
 let currentCleanups: Set<Cleanup> | undefined
+
+function runFnCallback<T extends CallableFunction>(value: T) {
+  value()
+}
 
 /**
  * watch
@@ -46,18 +52,20 @@ export function watch(
   const cleanup = () => {
     if (alive) {
       alive = false
-      cleanups.forEach((clean) => clean())
+      cleanups.forEach(runFnCallback)
       cleanups.clear()
-      subscriptions.forEach((unsubscribe) => unsubscribe())
+      subscriptions.forEach(runFnCallback)
       subscriptions.clear()
     }
   }
 
-  const revalidate = () => {
+  const revalidate = async () => {
     if (!alive) {
       return
     }
-    cleanups.forEach((clean) => clean())
+
+    // run own cleanups before re-subscribing
+    cleanups.forEach(runFnCallback)
     cleanups.clear()
 
     const proxiesToSubscribe = new Set<ProxyObject>()
@@ -69,14 +77,28 @@ export function watch(
 
     // Ensures that the parent is reset if the callback throws an error.
     try {
-      const cleanupReturn = callback((proxyObject) => {
+      const promiseOrPossibleCleanup = callback((proxyObject) => {
         proxiesToSubscribe.add(proxyObject)
+        // in case the callback is a promise and the watch has ended
+        if (alive && !subscriptions.has(proxyObject)) {
+          // subscribe to new proxy immediately -> this fixes problems when Promises are used due to the callstack
+          const unsubscribe = subscribe(proxyObject, revalidate, options?.sync)
+          subscriptions.set(proxyObject, unsubscribe)
+        }
         return proxyObject
       })
+      const couldBeCleanup =
+        promiseOrPossibleCleanup && promiseOrPossibleCleanup instanceof Promise
+          ? await promiseOrPossibleCleanup
+          : promiseOrPossibleCleanup
 
       // If there's a cleanup, we add this to the cleanups set
-      if (cleanupReturn) {
-        cleanups.add(cleanupReturn)
+      if (couldBeCleanup) {
+        if (alive) {
+          cleanups.add(couldBeCleanup)
+        } else {
+          cleanup()
+        }
       }
     } finally {
       currentCleanups = parent
@@ -84,19 +106,10 @@ export function watch(
 
     // Unsubscribe old subscriptions
     subscriptions.forEach((unsubscribe, proxyObject) => {
-      if (proxiesToSubscribe.has(proxyObject)) {
-        // Already subscribed
-        proxiesToSubscribe.delete(proxyObject)
-      } else {
+      if (!proxiesToSubscribe.has(proxyObject)) {
         subscriptions.delete(proxyObject)
         unsubscribe()
       }
-    })
-
-    // Subscribe to new proxies
-    proxiesToSubscribe.forEach((proxyObject) => {
-      const unsubscribe = subscribe(proxyObject, revalidate, options?.sync)
-      subscriptions.set(proxyObject, unsubscribe)
     })
   }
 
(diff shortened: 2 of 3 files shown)


dai-shi:
Thanks for working on it. Been busy, but hopefully, I'll be able to take a look sometime soon.

src/vanilla.ts
```diff
@@ -357,7 +357,7 @@ export function getVersion(proxyObject: unknown): number | undefined {
 
 export function subscribe<T extends object>(
   proxyObject: T,
-  callback: (ops: Op[]) => void,
+  callback: (ops: Op[]) => void | Promise<void>,
```
dai-shi, line 360: I don't think it's necessary. If it does, something seems wrong.
```suggestion
  callback: (ops: Op[]) => void,
```

src/vanilla/utils/watch.ts
```diff
@@ -2,13 +2,19 @@ import { subscribe } from '../../vanilla.ts'
...
 type WatchOptions = {
   sync?: boolean
 }
 
 let currentCleanups: Set<Cleanup> | undefined
 
+function runFnCallback<T extends CallableFunction>(value: T) {
+  value()
+}
+
```
dai-shi, line 17: This is just a refactor. I personally don't prefer it. Please leave them unchanged at least in this PR.
(It might be better to include `.clear`, but that should be a separate PR.)

src/vanilla/utils/watch.ts
```diff
@@ -69,35 +77,40 @@ export function watch(
 
     // Ensures that the parent is reset if the callback throws an error.
     try {
-      const cleanupReturn = callback((proxyObject) => {
+      const promiseOrPossibleCleanup = callback((proxyObject) => {
         proxiesToSubscribe.add(proxyObject)
+        // in case the callback is a promise and the watch has ended
+        if (alive && !subscriptions.has(proxyObject)) {
+          // subscribe to new proxy immediately -> this fixes problems when Promises are used due to the callstack
```
dai-shi, line 84: Hm, sounds complicated.

src/vanilla/utils/watch.ts
```diff
@@ -69,35 +77,40 @@ export function watch(
...
+          const unsubscribe = subscribe(proxyObject, revalidate, options?.sync)
+          subscriptions.set(proxyObject, unsubscribe)
+        }
         return proxyObject
       })
+      const couldBeCleanup =
+        promiseOrPossibleCleanup && promiseOrPossibleCleanup instanceof Promise
+          ? await promiseOrPossibleCleanup
```
dai-shi, line 92: Awaiting here might cause a race condition? Is it somewhat covered by the above code?

src/vanilla/utils/watch.ts
```diff
@@ -69,35 +77,40 @@ export function watch(
...
 
       // If there's a cleanup, we add this to the cleanups set
-      if (cleanupReturn) {
-        cleanups.add(cleanupReturn)
+      if (couldBeCleanup) {
+        if (alive) {
+          cleanups.add(couldBeCleanup)
+        } else {
+          cleanup()
+        }
```
dai-shi, line 101: Is this a bug fix not related with promise support?

src/vanilla/utils/watch.ts
```diff
@@ -69,35 +77,40 @@ export function watch(
...
         subscriptions.delete(proxyObject)
         unsubscribe()
       }
     })
-
-    // Subscribe to new proxies
-    proxiesToSubscribe.forEach((proxyObject) => {
-      const unsubscribe = subscribe(proxyObject, revalidate, options?.sync)
-      subscriptions.set(proxyObject, unsubscribe)
-    })
```
dai-shi, line 100: So, this is moved to L85.

dai-shi (changes requested):
Hey, thanks for working on it and sorry for the delay.
I'm not super confident with the watch code, but if we have good test coverage, it will be mergeable.

Just a small request is to change the diff in this PR as minimal as possible.

src/vanilla.ts
LiamMartens: You're right the promise return value isn't being used anyways

src/vanilla/utils/watch.ts
LiamMartens: Fair enough. Just FYI the reasoning behind is to not create new functions on each revalidate (which can increase GC time)

src/vanilla/utils/watch.ts
LiamMartens: This is related to your previous comment. Since we are awaiting the promise - it is possible for the watcher to already have been cleaned up and alive set to false by the time the promise resolves. In which case the only thing that should be done is the cleanup of the promise result itself.

src/vanilla/utils/watch.ts
```diff
@@ -2,7 +2,9 @@ import { subscribe } from '../../vanilla.ts'
 
 type Cleanup = () => void
 type WatchGet = <T extends object>(proxyObject: T) => T
-type WatchCallback = (get: WatchGet) => Cleanup | void | undefined
+type WatchCallback = (
+  get: WatchGet,
+) => Cleanup | void | Promise<void> | Promise<Cleanup> | undefined
```
dai-shi, line 7: Can this be this?
```suggestion
) => Cleanup | void | Promise<Cleanup | void> | undefined
```

tests/watch.test.tsx
```diff
@@ -2,6 +2,12 @@ import { describe, expect, it, vi } from 'vitest'
 import { proxy } from 'valtio'
 import { watch } from 'valtio/utils'
 
+function wait(ms: number) {
+  return new Promise<void>((resolve) => {
+    setTimeout(resolve, ms)
+  })
+}
```
dai-shi, line 9: Let's follow other tests:
```ts
const sleep = (ms: number) => 
  new Promise<void>((resolve) => {
    setTimeout(resolve, ms)           
  })                                                                      
```

tests/watch.test.tsx
```diff
@@ -96,4 +102,50 @@ describe('watch', () => {
 
     reference.value = 'Update'
   })
+  it('should support promise watchers', async () => {
+    const reference = proxy({ value: 'Example' })
+
+    const callback = vi.fn()
+
+    const waitPromise = wait(1000)
```
dai-shi, line 110: 1000ms is too long. Ideally, we want to have fake timers. We do something in `jotai`, but not in `valtio`. Hmm.

Maybe, it's not very difficult in this case?

```ts
let resolve
const waitPromise = new Promise((r) => { resolve = r })
...
resolve()
```

LiamMartens: Updated with fake timers from vitest - does that suffice? Also just reduced these to `0` but kept the timeout to make sure they are scheduled as a macrotask

dai-shi: > Updated with fake timers from vitest - does that suffice?

Yes! In such a case, we don't need to make it `0`, right? `1000` would be more meaningful. Or, maybe bigger. I set something like 30 sec somewhere, which makes it easy to know if the fake timer works as expected.

LiamMartens: You're right

tests/watch.test.tsx
```diff
@@ -96,4 +109,56 @@ describe('watch', () => {
...
+
+    const waitPromise = sleep(10000)
+    watch(async (get) => {
+      await waitPromise
+      get(reference)
+      callback()
+    })
+
+    vi.runAllTimers()
+    await waitPromise
```
dai-shi, line 125: Do we need this?

src/vanilla/utils/watch.ts
```diff
@@ -2,7 +2,9 @@ import { subscribe } from '../../vanilla.ts'
 
 type Cleanup = () => void
 type WatchGet = <T extends object>(proxyObject: T) => T
-type WatchCallback = (get: WatchGet) => Cleanup | void | undefined
+type WatchCallback = (
+  get: WatchGet,
+) => Cleanup | void | Promise<void | Cleanup> | undefined
```
dai-shi, line 7: Just a tiny preference:
```suggestion
) => Cleanup | void | Promise<Cleanup | void> | undefined
```

tests/watch.test.tsx
LiamMartens: In this instance yes - because whilst the timers will be run - the code is still synchronous and won't wait for the other scheduled tasks to finish - like the promise itself which is scheduled as a macro task so it would resolve after the current function call. It's similar to why you do `await Promise.resolve()`

src/vanilla/utils/watch.ts
LiamMartens: Done

tests/watch.test.tsx
dai-shi: Oh, yeah. Right. My oversight.

dai-shi (approved):
Thanks for your contribution!

Issue #449, opened 2022-04-28
Replacing object that you are subscribed to?


skokenes:
Hi, I have a question about valtio usage.

I have an object that I want to watch for changes. I do this like so:
```javascript
const unsubscribe = subscribe(state.myObject, () => {
  // callback to run whenever object changes
})
```

Occasionally, in response to a change in myObject, I want to reset the object completely to a new state. At first I tried doing this like so:
```javascript
const unsubscribe = subscribe(state.myObject, () => {
  // callback to run whenever object changes
  if(needToReplace) {
    state.myObject = newObjectState
  }
})
```

However, this breaks my subscription as subsequent subscriptions aren't received.
Is there a better way to do this type of scenario with valtio's API?

dai-shi:
In this scenario, you need to watch the parent `state` object
```js
const unsubscribe = subscribe(state, () => {
  ...
})
```

You can use `subscribeKey` if you want.
```js
const unsubscribe = subscribeKey(state, 'myObject', () => {
  ...
})
```

Alternatively, you can try to keep the `state.myObject` reference.
```js
const unsubscribe = subscribe(state.myObject, () => {
  // callback to run whenever object changes
  if(needToReplace) {
    for (const key in state.myObject) {
      delete state.myObject[key]
    }
    Object.assign(state.myObject, newObjectState)
  }
})
```

dai-shi:
Closing as resolved.

rubenapino:
> 

This is exactly what i was looking for , thanks! I didn't think to do Object.assign!

PR #875, opened 2024-03-27
replace listeners on an object that is replaced

Tucker-Eric:

Relates to #449

## Summary

This change allows for listeners to be resubscribed to objects that are replaced.

These changes allow the following code to not lose the subscription to the object that previously would lose the subscription:

```ts
const unsubscribe = subscribe(state.myObject, () => {
  // callback to run whenever object changes
  if(needToReplace) {
    state.myObject = newObjectState
  }
})
```

## Check List


dai-shi:
Thanks for the suggestion.
However, I think this is by design. We want the implementation to be minimal and I don't think this change works for some edge cases.
For example, we still want to use the old object.

PR #881, opened 2024-04-07
feat!: move to ESM first with dual ESM/CJS subpackage exports

userquin:

Talk with Kato and Moha in discord.

## Summary

Move to ESM first and include subpackage exports.

This PR also includes:
- now using rollup-plugin-dts  instead typescript to generate dts files
- remove `useProxy` from `src/macro.ts` module and patch the d.ts and d.cts files
- change `build` to run `build:all`: now we can generate multiples outputs (d.mts, d.cts and d.ts) and also the ts3.4 version, we need to play with the target and maybe lib to fix ts3.4
- ts3.4 now inside dist folder: we should include only dist in files (package.json)
- `postbuild`now includes only `yarn copy`, I've left old `xpostbuild`, we should remove it later
- `copy` script now copy src and esm from root, I've left old `xcopy`, we should remove it later: why copying src folder to remove it in the next script?

## Check List


package.json
```diff
@@ -1,49 +1,67 @@
 {
   "name": "valtio",
+  "type": "module",
```
dai-shi, line 3: ugh, unfortunately, this is not acceptable. we try to keep cjs first. we plan esm first for valtio v3.

userquin: This pr is for main, I can send another pr using commonjs to v2 branch (this PR should be closed)

dai-shi:
> I can send another pr using commonjs to v2 branch (this PR should be closed)

Sounds good. This PR introduces too many changes and can't even see if there are breaking changes or not. I'm pretty sure it breaks some old environments.

I would appreciate if you can send small PRs. For example: #882 (well, it's exceptionally small.)

package.json
```diff
@@ -1,49 +1,67 @@
...
+      "macro": ["./dist/ts3.4/macro.d.ts"],
+      "macro/vite": ["./dist/ts3.4/macro/vite.d.ts"],
+      "react": ["./dist/ts3.4/react.d.ts"],
+      "react/utils": ["./dist/ts3.4/react/utils.d.ts"],
+      "utils": ["./dist/ts3.4/utils.d.ts"],
+      "vanilla": ["./dist/ts3.4/vanilla.d.ts"],
+      "vanilla/utils": ["./dist/ts3.4/vanilla/utils.d.ts"],
+      "*": ["./dist/ts3.4/*"]
     },
-    "./*": {
```
dai-shi, line 34: I'm curious why we need to avoid subpath patterns?

dai-shi:
> small PRs

At least, we should separate for different issues: a) fix the result in `dist` and b) fix for tests, which may depend on a).

PR #891, opened 2024-05-03
fix(react): Change to useLayoutEffect in useSnapshot

endash:
Address spurious consistency check re-renders by using useLayoutEffect inside useSnapshot instead of useEffect

Fixes #890

## Summary

Changing from `useEffect` to `useLayoutEffect` fixes a timing issue in `useSnapshot` where a React consistency check runs before the effect to store the cached snapshot, causing a spurious re-render. 

## Check List


endash:
That makes complete sense. The existing basic test I worked from verifies
the core orthogonality guarantee, these just guard against a performance
regression.

tests/optimization.test.tsx
```diff
@@ -0,0 +1,116 @@
+import { useState } from 'react'
+import { fireEvent, render, waitFor } from '@testing-library/react'
+import { expect, it, vi } from 'vitest'
+import { proxy, snapshot, useSnapshot } from 'valtio'
```
dai-shi, line 4:
```suggestion
import { proxy, useSnapshot } from 'valtio'
```

dai-shi (approved):
Looks great to me. Thanks for your contribution!

endash:
🙌

PR #905, opened 2024-06-01
Update readme.md

cwtuan:
First and foremost, thank you for creating and maintaining the valtio package. I have found it invaluable for managing state in my applications. 

I'd like to propose adding a downloads trend image to the README file. The trend image is generated using a service that visualizes npm's download statistics for valtio. It has been embedded in the README to provide immediate visibility.

I believe this addition will be beneficial for new users and help promote valtio further. I hope you will consider merging this pull request. Thank you for your time and consideration.

Preview:

image


dai-shi:
Thanks for the suggestion, but we don't need it.

PR #919, opened 2024-07-03
Fix linear performance degradation in proxy map and proxy set

kk8787:

Fixes #

## Summary

Hello!
The company I work for uses Valtio in a product, but when handling CAD files we need to store 1,000 to 50,000 elements, and we are experiencing performance issues with the current implementation., so I fixed it.
This is my first time contributing to OSS, so I'm a little nervous.
Thank you.

## Check List

- [✅] `pnpm run prettier` for formatting code and docs

### DIFF
--- a/src/vanilla/utils/proxyMap.ts
+++ b/src/vanilla/utils/proxyMap.ts
@@ -1,9 +1,7 @@
 import { proxy } from '../../vanilla.ts'
 
-type KeyValRecord<K, V> = [key: K, value: V]
-
 type InternalProxyMap<K, V> = Map<K, V> & {
-  data: KeyValRecord<K, V>[]
+  data: boolean
   toJSON: object
 }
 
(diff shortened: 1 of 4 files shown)


dai-shi:
> This is my first time contributing to OSS, so I'm a little nervous.

That's great!

Unfortunately, this change doesn't work.
Proxying Map/Set is very tricky, and what we did is an emulation.
So, it's not for performance, but it just provides Map/Set-like APIs.

That said, proxying Map/Set with performance is technically possible under some reasonable constraints, and for example immer does it.
It would be very nice if anyone would challenge it. It will provide a performance benefit.
We decided not to include it in this repo, but adding such a library in https://github.com/valtiojs is more than welcome.

dai-shi:
(I will run CI again after #920 is merged.)

kk8787:
I see, I might give it a try when I have time.
Thank you for your kind reply.
I think valtio is the simplest way to integrate VanillaJS libraries with React in a natural way.
Thanks for a great library.

PR #927, opened 2024-07-28
Docs

overthemike:
N/A

Fixes #
N/A

## Summary
Added an example in the docs for '@' imports for typescript

## Check List

- [ x] `pnpm run prettier` for formatting code and docs

### DIFF
--- a/docs/how-tos/how-to-easily-access-the-state-from-anywhere-in-the-application.mdx
+++ b/docs/how-tos/how-to-easily-access-the-state-from-anywhere-in-the-application.mdx
@@ -81,6 +81,7 @@ module.exports = {
 ## Using Typescript
 
 Similar to babel above, if you are using typescript in your project, you can add the paths to tsconfig.json
+
 ```json
 {
   {
--- a/docs/how-tos/how-to-easily-access-the-state-from-anywhere-in-the-application.mdx
+++ b/docs/how-tos/how-to-easily-access-the-state-from-anywhere-in-the-application.mdx
@@ -78,6 +78,23 @@ module.exports = {
      (state, useSnapshot, subscribe)
    } from '@state';` from anywhere inside your application.
 
+## Using Typescript
+
+Similar to babel above, if you are using typescript in your project, you can add the paths to tsconfig.json
+```json
+{
+  {
+    "compilerOptions": {
+      // ...
+      "paths": {
+        "@/*": ["./src"]
+      }
+      // ...
+    }
+  }
+}
+```
+
 ## Using a third party library
 
 You can use a third party library to create aliases and achieve the same result.


dai-shi:
Are @roguesherlock and @RikuVan around?

roguesherlock:
I think we don't need a separate typescript example. jsconfig example can serve as typescript example too since it's exactly the same for typescript. Maybe you can just update the line for jsconfig?

Something like,

Create the file `/jsconfig.json` (`/tsconfig.json` if you're using typescript):

overthemike:
> I think we don't need a separate typescript example. jsconfig example can serve as typescript example too since it's exactly the same for typescript. Maybe you can just update the line for jsconfig?
Something like,
Create the file /jsconfig.json (/tsconfig.json if you're using typescript):

It is the same but I do remember looking the docs the first time I was trying this and didn't know that. Ended up having to look somewhere else for it. 

Totally down to make that change, however I was actually thinking today that there are a few other tsconfig properties tha need to align properly in order for the pathing to work. E.g. 

```json
    {
        // ...
        "moduleResolution": "NodeNext",
        "module": "Node,
        "resolveJsonModule": false
        // ...
    }
```
It's not usually a default for ts and without knowing about it, it'll trip you up.

overthemike:
> There's lot of other nuance that goes into tsconfig and tbh I feel like it's a bit out of scope of the valtio the library.

I almost said the exact same thing, but I kind of figured this particular document isn't really valtio related itself but helps those that are wanting to get going with it. I might even think about making a separate document with different scenarios at some point and then we can link something there if that's more in line with what your goals are.

roguesherlock:
> > There's lot of other nuance that goes into tsconfig and tbh I feel like it's a bit out of scope of the valtio the library.
> 
> I almost said the exact same thing, but I kind of figured this particular document isn't really valtio related itself but helps those that are wanting to get going with it. I might even think about making a separate document with different scenarios at some point and then we can link something there if that's more in line with what your goals are.

sorry I might have missed it then. Can we then update to following,
a. Update header for jsconfig to also include tsconfig 
b. Add a note below the code block with a link to tsconfig reference docs or base tsconfig? We can later link it to our internal docs for tsconfig when we have them :)

maybe something like this would help for links?
- https://www.totaltypescript.com/tsconfig-cheat-sheet
- https://github.com/tsconfig/bases
- https://www.typescriptlang.org/tsconfig/

overthemike:
So I got into the documentation zone last night while doing this and went through most the markdown files and tried to update / add as much as I could / know about. I'll go and comment anywhere in the docs where I specifically changed something and explain why.

overthemike:
OH - I just now realized that these docs update the valtio website. I'm not sure if all of these changes are good now. I was mainly focused on markdown for git. My apologies guys. I'm going to go ahead and close this PR and create a new one with fixes I can see working on the site.

PR #939, opened 2024-08-29
docs: improve v2 migration guide

dai-shi:
## Related Bug Reports or Discussions

Fixes #938

## Summary

Clarify about impure proxy().

## Check List

- [x] `pnpm run prettier` for formatting code and docs

PR #941, opened 2024-08-31
feat: Codemods for Upgrading from Valtio v1 to v2

manishjha-04:
**🚀 Feature Proposal**

Building codemods for Upgrading from Valtio v1
Looking forward to add these and help Valtio users have a smooth migration experience.

**Motivation**
The motivation to create codemods is that it would streamline the process for millions of users, and also ensure that the application stays efficient and secure while leveraging the latest advancements in Valtio.

**Example**
These codemods would help migrate from Valtio v1 to v2 real quick as running :

> `npx codemod valtio/v2/migration-recipe `

which would include all the codemods.

### DIFF
--- a/docs/guides/migrating-to-v2.mdx
+++ b/docs/guides/migrating-to-v2.mdx
@@ -26,6 +26,24 @@ Other notable changes to keep things updated and fresh include:
 - Requirement of TypeScript version 4.5 and above
 - The build target updated to ES2018
 
+## Codemods
+
+To assist with the upgrade, a set of codemods is available to automatically update your codebase from Valtio v1. These codemods automate common code transformations to streamline the migration process.
+
+Run all codemods listed in this guide with the Valtio v2 codemod recipe:
+
+  ```bash     
+  npx codemod valtio/v2/migration-recipe
+  ```
+
+This will run the following codemods:
+ - `valtio/v2/use-snapshot-migration`
+ - `valtio/v2/deep-clone-proxy-objects`
+
+Each of these codemods automates the changes listed in this migration guide.
+For a complete list of available Valtio codemods and further details, 
+see the [codemod registry](https://codemod.com/registry?q=valtio).
+
 ## Migration for breaking changes
 
 ### Resolving promises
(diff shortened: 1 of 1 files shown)


docs/guides/migrating-to-v2.mdx
```diff
@@ -26,6 +26,24 @@ Other notable changes to keep things updated and fresh include:
 - Requirement of TypeScript version 4.5 and above
 - The build target updated to ES2018
 
+## Codemods
```
dai-shi, line 29: Thanks for your suggestion! However, I would like to move this section after `## Migration` and remove the change in the migration section. Also, we may need to add some caveats.

Here're some thoughts behind it:
- I don't think most users need migration.
- Only a few users might depend on the subtle behavioral difference, but they should reconsider their code.
- Doing codemods blindly may result in bad performance.
- I haven't looked into the codemode impl, but it may not work for some cases, because we can't statically know what are promises.

dai-shi (changes requested):
Hi, thanks for working on this.

I generally like the idea of codemod, but it feels this is more confusing. The codemod does too much and it fails in edge cases.

Have you tried using this codemod in the large codebase? Maybe we need more experiments before promoting it.

dai-shi:
One failing case:
```js
import { proxy, useSnapshot } from 'valtio';

const state = proxy({ data: fetch(api).then((res) => res.json()) });

const Component = () => {
    const snap = useSnapshot(state);
    return <>{JSON.stringify(snap.data.title)}</>;
};
```

dai-shi:
We don't need codemod for `const state = proxy({ count: 1, obj: { text: 'hi' } });`
Please see #939. Sorry for the confusion.

manishjha-04:
Hi @dai-shi ,

Thanks for the feedback! I understand your concerns. If you have any repositories in mind, I'd be happy to use them to pressure test the codemod and work on refining it to handle edge cases better before merging it to the upgrade doc. Let me know, and we can make improvements based on those tests.

dai-shi:
I don't have anything in mind, but you might be able to find some open source project with GitHub search capability.
In any case, we need some test cases.

manishjha-04:
Closing this for now and will be raising a new PR after pressure testing!

PR #947, opened 2024-09-13
docs - Added a note in how-tos/reset on how to use structuredClone

overthemike:
N/A
Fixes #
N/A
## Summary
Added example on how to use structuredClone in v2 and the reasoning behind why it's different than v1.

## Check List

### DIFF
--- a/docs/how-tos/how-to-reset-state.mdx
+++ b/docs/how-tos/how-to-reset-state.mdx
@@ -66,4 +66,4 @@ const reset = () => {
 }
 ```
 
-See also [title: 'How to Migrate to v2 from v1'](../guides/migrating-to-v2.mdx#impure-proxyobj).
+See also: ['How to Migrate to v2 from v1'](../guides/migrating-to-v2.mdx#impure-proxyobj).
--- a/docs/how-tos/how-to-reset-state.mdx
+++ b/docs/how-tos/how-to-reset-state.mdx
@@ -53,7 +53,7 @@ const reset = () => {
 }
 ```
 
-In valtio v2, the `proxy` function [no longer clones the initial state]('../guides/migrating-to-v2.mdx'). Instead, it uses the original object directly. This means that `structuredClone` is no longer a drop in replacement for `_.cloneDeep` and will throw an error. To fix this, you can use `structuredClone` to clone the initial state that you pass to into `proxy`.
+In valtio v2, the `proxy` function [no longer clones the initial state]('../guides/migrating-to-v2.mdx'). Instead, it uses the original object directly. It is not recommended that you reuse the object that you pass into the proxy function. If you use `structuredClone` that object, it will throw an error. `_.cloneDeep` won't give you an error, but it could produce unwatned results. To fix this, you can clone the intial state object that you pass to into `proxy`. You can do this with `_.cloneDeep`, `structuredClone`, or the utility function `deepClone` from `valtio/utils`. Example below.
 
 ```js
 // v2
@@ -65,3 +65,5 @@ const reset = () => {
   state.obj = structuredClone(initialObj)
 }
 ```
+
+See also [title: 'How to Migrate to v2 from v1']('../guides/migrating-to-v2.mdx').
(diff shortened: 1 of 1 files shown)


docs/how-tos/how-to-reset-state.mdx
```diff
@@ -34,3 +34,34 @@ const reset = () => {
   state.obj = _.cloneDeep(initialObj)
 }
 ```
+
+> [!NOTE]
+> Using `structuredClone()`
+
+In 2022, there was a new global function added called `structuredClone` that is widely available in most modern browsers. While this function and `_.cloneDeep` from lodash seem to do the same thing, their implementations are quite different.
+
+In valtio v1, `proxy` cloned the object used for inital state under the hood. Because the original object isn't used, `structuredClone` works just fine as a drop in replacement for `_.cloneDeep`.
```
dai-shi, line 43: Is this only the issue with `structuredClone`? Can it be the same with `_.cloneDeep`?

overthemike: It doesn't seem to. I haven't tested very thoroughly but from what I've tested so far, it seems to work ok, even modifying the structure of the object doesn't really seem to affect it too much. This example doesn't seem to have any issues

```js
import { cloneDeep } from 'lodash'

const initialState = {
  foo: {
    bar: {
      bat: 'bat initial'
    }
  }
}
const state = proxy({ obj: initialState })
const reset = () => {
  state.obj = cloneDeep(initialState)
}

function App() {
  const snap = useSnapshot(state)

  return (
    <div>
      <p>snap.foo.bar.bat: {snap.obj.foo.bar.bat}</p>
      <button onClick={() => state.obj = { foo: { bar: { asdf: 'bat2', ball: 'ball' } }}}>click</button>
      <button onClick={() => state.obj = { foo: { bar: 'bar2'}}}>click</button>
      <button onClick={() => reset()}>reset</button>
      <pre>{JSON.stringify(snap.obj, null, 2)}</pre>
    </div>
  )
}

export default App

```

overthemike: Outputs as expected:

Initial state
```
snap.foo.bar.bat: bat initial

{
  "foo": {
    "bar": {
      "bat": "bat initial"
    }
  }
}
```

First button click
```
snap.foo.bar.bat:

{
  "foo": {
    "bar": {
      "asdf": "bat2",
      "ball": "ball"
    }
  }
}
```

Second button click
```
snap.foo.bar.bat:

{
  "foo": {
    "bar": "bar2"
  }
}
```

Reset button click
```
snap.foo.bar.bat: bat initial

{
  "foo": {
    "bar": {
      "bat": "bat initial"
    }
  }
}
```

dai-shi: Can you try this?

```jsx
<button onClick={() => { state.obj.foo.bar.bat = 'bat changed' }}>click</button>
```

overthemike: yeah that does mess it up. I'll update the PR once I get a chance.

overthemike: I updated the docs. Let me know if you want anything reworded.

docs/how-tos/how-to-reset-state.mdx
```diff
@@ -34,3 +34,36 @@ const reset = () => {
   state.obj = _.cloneDeep(initialObj)
```
dai-shi, line 34: I mean this is wrong too. L31-35

Can you please fix the doc in this way?
- Fix cloneDeep usage with v2.
- Add a note about structuredClone

Speaking of v2, we now have `deepClone` from valtio/utils, we don't need to use cloneDeep.
So, we could restructure the docs to use `deepClone` first and `structuredClone` second.
(I thought about explaining structuredClone first, but `deepClone` is capable of avoid copying `ref()`'d object.

overthemike: Sure thing. I'll push up the changes soon.

overthemike: Changes have been pushed. Let me know of any feedback.

dai-shi (approved):
Looks perfect!

PR #965, opened 2024-10-13
New implementation of proxyMap and proxySet with performance improvements

overthemike:

#943 

### Improvements:
  - performance improvements from O(n) to O(log n) for both proxyMap and proxySet

### proxySet now as new Set methods:
  - intersection
  - isDisjointFrom
  - isSubsetOf
  - isSupersetOf
  - symmetricDifference
  - union

Fixed some previously unknown bugs with old implementations and added some new tests. 

## Check List
- [x] pnpm run prettier for formatting code and docs


src/vanilla/utils/proxyMap.ts
```diff
@@ -1,111 +1,166 @@
...
+    },
+    get(key: K) {
+      const map = getSnapMap(this) || indexMap
+      const k = maybeProxify(key)
+      const index = map.get(k)
+      if (index === undefined) {
+        if (!isProxy(this)) {
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          this.index
+        }
```
dai-shi, line 64: We could make it `this.index` only, which should be more performant, because if it's a proxy, it's almost nothing, and probably faster than `isProxy()`.

src/vanilla/utils/proxyMap.ts
```diff
@@ -1,111 +1,166 @@
...
+    const cache = snapCache.get(vObject)
+    const latestSnap = cache?.[1]
+    if (latestSnap && !snapMapCache.has(latestSnap)) {
+      const clonedMap = new Map(indexMap)
+      snapMapCache.set(latestSnap, clonedMap)
+      return true
+    }
+    return false
+  }
+  const getSnapMap = (x: any) => snapMapCache.get(x)
```
dai-shi, line 29: It feels a bit repetitive, so we can make it like:
```suggestion
  const getMapForThis = (x: any) => snapMapCache.get(x) || indexMap
```
I'm not sure about the naming.

dai-shi (commented):
Probably the final review.

src/vanilla/utils/proxyMap.ts
```diff
@@ -1,111 +1,162 @@
...
+        this.index
+        return undefined
+      }
+      return this.data[index + 1] as V
+    },
+    has(key: K) {
+      const map = getMapForThis(this)
+      const k = maybeProxify(key)
+      const exists = map.has(k)
+      if (!exists && !isProxy(this)) {
```
dai-shi, line 69:
```suggestion
      if (!exists) {
```

src/vanilla/utils/proxySet.ts
```diff
@@ -1,98 +1,197 @@
...
+      if (!isProxy(this)) {
+        registerSnapMap()
+      }
+      return indexMap.size
+    },
+    has(v: T) {
+      const map = getMapForThis(this)
+      const value = maybeProxify(v)
+      const exists = map.has(value)
+      if (!exists && !isProxy(this)) {
```
dai-shi, line 61:
```suggestion
      if (!exists) {
```

dai-shi (commented):
can do the same.

dai-shi (approved):
🎉

PR #981, opened 2024-10-23
fixed issue with .has() when new value is added and then deleted or c…

overthemike:
…leared

#979
Fixes #
#979 
## Summary
Accessing this.index in the has method doesn't seem to work when a new item is added and then deleted (or cleared) at a later time. Accessing the this.data.length property fixes this issue.

## Check List

### DIFF
--- a/src/vanilla/utils/proxyMap.ts
+++ b/src/vanilla/utils/proxyMap.ts
@@ -87,10 +87,9 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
     },
     has(key: K) {
       const map = getMapForThis(this)
-      const exists = map.has(key)
       // eslint-disable-next-line @typescript-eslint/no-unused-expressions
       this.data.length // touch property for tracking
-      return exists
+      return map.has(key)
     },
     set(key: K, value: V) {
       if (!isProxy(this)) {
--- a/src/vanilla/utils/proxyMap.ts
+++ b/src/vanilla/utils/proxyMap.ts
@@ -88,10 +88,8 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
     has(key: K) {
       const map = getMapForThis(this)
       const exists = map.has(key)
-      if (!exists) {
-        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-        this.index // touch property for tracking
-      }
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.data.length // touch property for tracking
       return exists
     },
     set(key: K, value: V) {
--- a/src/vanilla/utils/proxyMap.ts
+++ b/src/vanilla/utils/proxyMap.ts
@@ -91,7 +91,7 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
       const map = getMapForThis(this)
       const exists = map.has(key)
       // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-      this.epoch
+      this.epoch // touch property for tracking
       return map.has(key)
     },
     set(key: K, value: V) {
--- a/src/vanilla/utils/proxyMap.ts
+++ b/src/vanilla/utils/proxyMap.ts
@@ -6,6 +6,7 @@ const isProxy = (x: any) => proxyStateMap.has(x)
 type InternalProxyObject<K, V> = Map<K, V> & {
   data: Array<V>
   index: number
+  epoch: number
   toJSON: () => Map<K, V>
 }
 
@@ -68,6 +69,7 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
   const vObject: InternalProxyObject<K, V> = {
     data: initialData,
     index: initialIndex,
+    epoch: 0,
     get size() {
       if (!isProxy(this)) {
         registerSnapMap()
@@ -80,15 +82,16 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
       const index = map.get(key)
       if (index === undefined) {
         // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-        this.index // touch property for tracking
+        this.epoch // touch property for tracking
         return undefined
       }
       return this.data[index]
     },
     has(key: K) {
       const map = getMapForThis(this)
+      const exists = map.has(key)
       // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-      this.data.length // touch property for tracking
+      this.epoch
       return map.has(key)
     },
     set(key: K, value: V) {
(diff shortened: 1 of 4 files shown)


src/vanilla/utils/proxyMap.ts
```diff
@@ -88,10 +88,8 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
     has(key: K) {
       const map = getMapForThis(this)
       const exists = map.has(key)
-      if (!exists) {
-        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-        this.index // touch property for tracking
-      }
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.data.length // touch property for tracking
       return exists
```
dai-shi, line 93: Let's do this:
```suggestion
      return map.has(key)
```

dai-shi (commented):
Thanks for working on it. Is it possible to add a test?

src/vanilla/utils/proxyMap.ts
overthemike: Yeah I was planning on it actually. Just thought I'd push this up first to get the patch out. I didn't have a whole lot of time today with other work. Do you want me to push up a different pr for the tests or do you want it in this one?

overthemike: I added the tests and pushed up

src/vanilla/utils/proxyMap.ts
```diff
@@ -87,12 +87,9 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
...
     has(key: K) {
       const map = getMapForThis(this)
-      const exists = map.has(key)
-      if (!exists) {
-        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-        this.index // touch property for tracking
-      }
-      return exists
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.data.length // touch property for tracking
```
dai-shi, line 91: What is the case `this.data.length` changes, but `this.index` does not? It feels like we need another number like `epoch`.

tests/proxyMap.test.tsx
```diff
@@ -375,3 +376,40 @@ describe('snapshot', () => {
...
+describe('ui updates - useSnapshot', async () => {
+  it('should update ui when calling has before and after deleting a key', async () => {
+    const state = proxyMap()
+    const TestComponent = () => {
+      const snap = useSnapshot(state)
+
+      return (
+        <>
+          <p>has key: {`${snap.has('key')}`}</p>
+          <button onClick={() => state.set('key', 'value')}>set key</button>
```
dai-shi, line 389: Try something like this:
```suggestion
          <button onClick={() => { state.set('key', 'value'); state.set('key2', 'value'); }}>set key</button>
```

dai-shi (changes requested):
it probably requires more.

src/vanilla/utils/proxyMap.ts
overthemike: The case I put in the tests was precisely the case that didn't work with this.index. When I wrote the tests, I went back to this.index just to make sure I wasn't missing something and the test didn't pass. When a key is deleted or cleared, `.has()` wasn't updating the ui with `useSnapshot`. The tests we had before was just testing snapshot, which was taken manually.

Out of curiosity, is there a reason you don't like checking this.data.length? It's O(1).

tests/proxyMap.test.tsx
overthemike: This is a good reason 😅😬

overthemike: I added an epoch to keep track of versions instead of using this.data.length and added tests to test for adding and deleting multiple keys as well as clearing the map/set

src/vanilla/utils/proxyMap.ts
dai-shi: `this.data.length` doesn't change if we delete an item (maybe except for the last one.)

overthemike: I originally thought that in order for updates to happen, `this.data.length` had to have changed, but I'm not sure if that's the case. I'm not really sure why it (sort of) works when deleting just one item. I think it's because we're tracking touch the property? I also don't really understand why it would work with a single element, but not with multiple elements. Might only update for the first one. Either way, epoch works with both, and I think it's probably easier to understand what it's doing rather than just accessing the length property randomly on an array....👍😁

overthemike:
Is there anything else that needs doing before this gets merged?

src/vanilla/utils/proxyMap.ts
dai-shi: It's because we use a sparse array.

src/vanilla/utils/proxyMap.ts
```diff
@@ -80,19 +82,17 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
...
     has(key: K) {
       const map = getMapForThis(this)
       const exists = map.has(key)
-      if (!exists) {
-        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-        this.index // touch property for tracking
-      }
-      return exists
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.epoch
```
dai-shi, line 94:
```suggestion
      this.epoch // touch property for tracking
```

tests/proxyMap.test.tsx
```diff
@@ -375,3 +376,145 @@ describe('snapshot', () => {
...
+
+  it('should update ui when calling has before and after settiing and deleting multiple keys', async () => {
+    const state = proxyMap()
+    const TestComponent = () => {
+      const snap = useSnapshot(state)
+
+      return (
+        <>
+          <p>has key: {`${snap.has('key')}`}</p>
+          <p>has key2: {`${snap.has('key2')}`}</p>
```
dai-shi, line 424: I think we should have a test that deletes the first key only. (Did I miss it somewhere?)

dai-shi (commented):
Looks good overall.

tests/proxyMap.test.tsx
overthemike: I'll add it

overthemike: I went ahead and added another test that tests deleting the first key as well as deleting only the second (or last) key. Couldn't hurt.

dai-shi (approved):
This causes extra rerenders with `.has()`, which doesn't seem ideal, but can't be helped. It's ironic that the use of `.has()` should be discouraged, even though proxyMap/Set should exist for performance.

overthemike:
Yeah - `.has()` essentially needs to check if the value exists every time the data changes :\

overthemike:
Are we waiting on anything to get this merged and get a patch version published?

dai-shi:
https://github.com/pmndrs/valtio/releases/tag/v2.1.1

PR #984, opened 2024-10-27
test: calculate coverage for src files only

kretajak:
## Summary

Code coverage should rather take into account files from `src`.

Quote from Vitest docs:

> It's recommended to always define [coverage.include](https://vitest.dev/config/#coverage-include) in your configuration file. This helps Vitest to reduce the amount of files picked by [coverage.all](https://vitest.dev/config/#coverage-all).

## Check List


dai-shi (approved):
Thanks for your contribution!

PR #988, opened 2024-11-03
Fixed iterator reactivity in proxyMap and proxySet

overthemike:
#987
Fixes #
#987
## Summary

proxyMap and proxySet iterator methods lacked having the epoch version being set when they are used. Added them in and added tests to make sure they work.

## Check List


src/vanilla/utils/proxyMap.ts
```diff
@@ -130,24 +130,32 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
...
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.epoch // touch property for tracking
       const map = getMapForThis(this)
       for (const key of map.keys()) {
         yield key
       }
     },
     *values(): IterableIterator<V> {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.epoch // touch property for tracking
```
dai-shi, line 158: I haven't looked at the test and others yet, but this feels wrong...
(If this fixes something, we may have misunderstood something why `this.index` didn't work. that was actually my concern back then.)

overthemike: This is (so far) what I think is going on:
1. Reactivity is triggered when a tracke)d property is accessed (correct me if I'm wrong here)
2. The epoch property changes whenever the proxyMap structure changes
3. By touching epoch in the iterator methods, we ensure they're re-evaluated when there are updates

Without having a tracked property being touched inside of the the methods, the items within any of the iterator methods were not updating when a new key is added or changed. E.G.

```typescript
const state = proxyMap()
const Component = () => {
  const snap = useSnapshot(state)

  return (
    <ul>
      <button onClick={() => addItemToMap()}>Add Item</button>
      {Array.from(snap.values()).map(item => (
        <li key={item.id}>{item.name}</li> {/* this was not updating on click */}
      ))}
    </ul>
  )
}
```

overthemike: Is this the same scenario as using `this.index` inside of the individual mutation methods? This is only calling the epoch property once per full iteration to make sure that changes to data are reflected. I could add in an `!isProxy` check to only access epoch property within those methods within a snapshot.

dai-shi: Yeah, I think I get it. In the previous version (v2.0.0), it touches `this.data.length` behind the scenes, but with v2.1.x, we need to do it manually.

dai-shi (approved):
Will merge later when I prepare the patch.

src/vanilla/utils/proxyMap.ts
```diff
@@ -130,24 +130,32 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
...
     *keys(): IterableIterator<K> {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.epoch // touch property for tracking
       const map = getMapForThis(this)
       for (const key of map.keys()) {
         yield key
       }
     },
     *values(): IterableIterator<V> {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
dai-shi, line 157: Can you disable the rule in "overrides" section in `.eslintrc.json`?
Just for `proxyMap.ts` and `proxySet.ts`.

tests/proxyMap.test.tsx
```diff
@@ -621,4 +621,137 @@ describe('ui updates - useSnapshot', async () => {
...
+          <ul>
+            {Array.from(snap.values()).map((mapitem) => (
+              <li key={mapitem.id}>{`${mapitem.name} ${mapitem.id}`}</li>
+            ))}
+          </ul>
+        </>
+      )
+    }
+
+    const { getByText } = render(
```
dai-shi, line 656: Please follow the #989 change.

dai-shi (changes requested):

.eslintrc.json
```diff
@@ -49,6 +49,7 @@
       "warn",
       { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }
     ],
+    "@typescript-eslint/no-unused-expressions": "off",
```
dai-shi, line 52: Please add the rule in "overrides" and apply it only to `proxySet/Map.ts`.

src/vanilla/utils/proxyMap.ts
```diff
@@ -81,15 +81,13 @@ export function proxyMap<K, V>(entries?: Iterable<[K, V]> | undefined | null) {
       const map = getMapForThis(this)
       const index = map.get(key)
       if (index === undefined) {
-        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
         this.epoch // touch property for tracking
```
dai-shi, line 84: Should this be outside of the "if" block?
Does it cover the case when we delete the item afterwards?

overthemike: Let me check

dai-shi (approved):
Thanks for your work!!

overthemike:
happy to be helpful!

PR #992, opened 2024-11-08
test: incorporate eslint-plugin-testing-library and fix reported issues

kretajak:
## Summary

Testing Library provides [Eslint plugin ](https://testing-library.com/docs/ecosystem-eslint-plugin-testing-library) which encourages to use best practices.

## Check List

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,6 +21,7 @@
     "react-hooks",
     "import",
     "vitest",
+    "testing-library",
     "eslint-plugin-react-compiler"
   ],
   "parser": "@typescript-eslint/parser",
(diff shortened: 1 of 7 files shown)


.eslintrc.json
```diff
@@ -109,8 +110,10 @@
   },
   "overrides": [
     {
+      "extends": ["plugin:testing-library/react"],
       "files": ["tests/**/*.ts", "tests/**/*.tsx"],
       "rules": {
+        "testing-library/no-manual-cleanup": "off",
```
kretajak, line 116: Added as seems like cleanup is not done automatically, and cleanup added in `setup.ts` is still needed.

kretajak: Automatic cleanup is not enabled as current tests configuration does not have `globals`. to `true`. More on that: https://github.com/testing-library/vue-testing-library/issues/296.

I would leave this rule turned off rather than enabling globals.

PR #997, opened 2024-11-11
test: add jest-dom matchers and eslint plugin

kretajak:
## Summary

Testing Library provides nice set of custom matchers that are more declarative and easier to understand. Along with `@testing-library/jest-dom` there is an accompanying `eslint-plugin-jest-dom` which enforce using these selectors. 

## Check List


tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "compilerOptions": {
     "target": "esnext",
     "strict": true,
-    "lib": ["es2023"],
```
kretajak, line 5: This line seems to be problematic with my changes and old TS versions. Zustand supports the same range of TypeScript versions but lacks this line. Maybe it can be safely removed?

dai-shi: It's added in #930. I think it's related to `} catch {` change.
If omitting this line goes well, I think it's fine to remove.

kretajak: Seems like using `"lib": ["dom"]` satisfies all TS versions.

PR #1015, opened 2024-12-16
docs(*): change 'npm i' to 'npm install'

sukvvon:
## Summary

* Since this is an official document, I determined that using `npm install` instead of the abbreviation `npm i` is more appropriate, so I made the change.
* In the official homepage, both `npm install` and `npm i` are used, so I standardizied them.

## Check List


dai-shi (commented):
I don't mind much, but what do you think? @overthemike

overthemike:
I don't mind either way either. They both work just fine.

dai-shi (approved):
two zeros and one plus.

PR #1024, opened 2024-12-19
chore(.github/workflows): improve workflow with removing duplicated configs

sukvvon:
## Summary

* Same with https://github.com/pmndrs/zustand/pull/2918
* `pmpx` is alias version from `pnpm dlx`, i think using full name is right

## Check List


dai-shi (commented):
Should we use `pnpm dlx` like others?

sukvvon:
wait a second please i found that i missed

sukvvon:
@dai-shi Done!

PR #1038, opened 2025-01-20
fix: make `proxy` a no-op when passed a proxy

aleclarson:
## Summary

As far as I can tell, there's no benefit to wrapping a Valtio proxy with `proxy(…)` but it can happen inadvertently. In this case, Valtio should simply return the proxy unchanged, since it's already subscribable.

In [valtio-kit](https://github.com/aleclarson/valtio-kit), I'm using `proxy(new Foo())` as a compiler hint, so the compiler can know to wrap references to the `Foo` instance with `get(…)` inside of `watch` handlers. In my testing of this compiler hint, I found this undesirable Valtio behavior of “proxying a proxy”.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -164,6 +164,9 @@ export function proxy<T extends object>(baseObject: T = {} as T): T {
   if (found) {
     return found
   }
+  if (proxyStateMap.has(baseObject)) {
+    return baseObject
+  }
   let version = versionHolder[0]
   const listeners = new Set<Listener>()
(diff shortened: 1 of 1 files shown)


dai-shi:
I'm not sure if I can explain this well, but it feels like there might be a case where we want wrapping proxies again. Adding the proposed check will prevent such a use case, so this isn't acceptable at the moment. Adding the same check with a wrapper is technically possible.

I assume you have your solution on valtio-kit without changing valtio, and this is just to let us know this should be an expected behavior. Thanks for the suggestion anyway.

PR #1040, opened 2025-01-22
fix(proxySet): fix `.symmetricDifference`,`.isDisjointFrom`; add `.difference`

magicdawn:

N/A

## Summary

- add missing `.difference` method
- fix incorrect implemented `.symmetricDifference` & `.isDisjointFrom`
- copy mdn examples as basic Unit Test
- re-order these methods according to the proposal page

the proposal
https://github.com/tc39/proposal-set-methods?tab=readme-ov-file#proposal

## Check List


overthemike (approved):
Nice work! It's much appreciated. I didn't spend nearly enough time diving deep into these when first implementing. Good looking out and well done.

Verified (quickly) these changes are indeed the correct implementation for these methods.

PR #1050, opened 2025-02-08
docs: add `tauri-plugin-valtio` as a third-party library

ferreira-tb:
## Summary

`tauri-plugin-valtio` is a Tauri plugin that provides a valtio integration. It can be used to save state to disk, share it across multiple windows, or even access it from the Rust backend.

## Check List


dai-shi (approved):
👍

PR #1070, opened 2025-03-09
Expose listeners

anthonyshort:
## Summary

Exposing the internal listeners state via the proxyStateMap. This is for a case where we are using valtio-history with subscriptions at something other than the root. I'm hoping there's an easier way to do this. 

This is allowing us to solve this problem:

```ts
const state = proxy({
  numbers: [] as number[],
  sum: undefined as number | undefined,
});

subscribe(state.numbers, () => {
  state.sum = state.numbers.reduce((acc, n) => acc + n, 0);
}, true)

const history = proxyWithHistory(state);

history.value.numbers.push(1);
expect(history.value.sum).toBe(1);
await Promise.resolve(); // history snapshot

history.value.numbers.push(2);
expect(history.value.sum).toBe(3);
await Promise.resolve(); // history snapshot

history.undo();
expect(history.value.sum).toBe(1);
expect(history.value.numbers).toEqual([1]);

history.value.numbers.push(2);
expect(history.value.numbers).toEqual([1, 2]);
expect(history.value.sum).toBe(3); // <-- Fails
```

We have our own internal wrapper around proxyWithHistory to add some custom functionality (like allowing some state to trigger renders, but not be tracked in history, similar to refs). Maybe valtio-history doesn't need something like this as a core feature, but exposing the internal state allows us to handle it.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -43,6 +43,7 @@ type ProxyState = readonly [
   target: object,
   ensureVersion: (nextCheckVersion?: number) => number,
   addListener: AddListener,
+  listeners: Set<Listener>,
 ]
 
 const canProxyDefault = (x: unknown): boolean =>
@@ -251,7 +252,12 @@ export function proxy<T extends object>(baseObject: T = {} as T): T {
   )
   const proxyObject = newProxy(baseObject, handler)
   proxyCache.set(baseObject, proxyObject)
-  const proxyState: ProxyState = [baseObject, ensureVersion, addListener]
+  const proxyState: ProxyState = [
+    baseObject,
+    ensureVersion,
+    addListener,
+    listeners,
+  ]
   proxyStateMap.set(proxyObject, proxyState)
   Reflect.ownKeys(baseObject).forEach((key) => {
     const desc = Object.getOwnPropertyDescriptor(


anthonyshort:
Realised that this won't actually solve the problem because we're closing over state.

dai-shi:
> Realised that this won't actually solve the problem because we're closing over state.

So, what's your solution?

I'm fine to expose some internals, but I would probably like to refactor a bit (for example, having both `addListener` and `listeners` doesn't seem very nice). In any case, we need to know the solution first.

anthonyshort:
I was trying some things out today. Landed on a solutation that just wraps the proxy:

```ts
export function observable<T extends object>(object: T): T {
  if (cache.has(object)) {
    return cache.get(object) as T;
  }
  if (isObservable(object)) {
    return object;
  }
  const state = valtioProxy(object);
  const listeners = new Set<Listener<T>>();
  observerState.set(state, {
    listeners,
  });
  valtioSubscribe(state, () => {
    listeners.forEach((listener) => listener(state));
  }, true);
  cache.set(object, state);
  return state as T; 
}

export function observe<T extends object>(
  object: T,
  callback: Listener<T>,
): () => void {
  const proxyObject = observable(object);
  const listeners = getListeners(proxyObject);
  if (listeners == null){
    throw new Error('Proxy object does not have listeners');
  }
  listeners.add(callback);
  const removeListener = () => {
    listeners.delete(callback);
  }
  return removeListener;
}
```

Allowing for this API:

```ts
const obj = observable({
  count: 0,
  doubled: 0,
});
observe(obj, (newState) => {
  newState.doubled = newState.count * 2;
});
obj.count = 1;
expect(obj.doubled).toEqual(2);
```

or this:

```ts
const obj = observable({
  count: 0,
});
const derived = computed(obj, {
  doubled: (state) => state.count * 2,
  tripled: (state) => state.count * 3,
});
obj.count = 1;
expect(derived.doubled).toEqual(2);
expect(derived.tripled).toEqual(3);
```

Then I have a `clone` function that I can call that will just create a copy of the object, then copy over the listeners. We have a pretty elaborate wrapper around proxyWithHistory. We have a custom history snapshot function similar to deepClone and I just throw in `isObservable(ob)` and clone it to carry over the listeners.

```ts
const obj = observable({
  count: 0,
  doubled: 0,
});
observe(obj, (newState) => {
  newState.doubled = newState.count * 2;
});
const obj2 = clone(obj); // <--- clone it and copy over the listeners
obj2.count = 1;
expect(obj2.doubled).toEqual(2); // <-- it still works on the clone
```

Overall, I don't actually mind that this is a userland land thing now. I want to see if I can copy what you've done with valtio-reactive, but not entirely sure if it makes sense.

PR #1085, opened 2025-04-18
fix(package): react-native condition

dai-shi:
following https://github.com/pmndrs/jotai/pull/3042


ignaciosantise:
hey @dai-shi 👋 could you also add this fix to valtio v1?

PR #1102, opened 2025-06-07
add umd build option

AB498:

Fixes #

## Check List

- [ ] `pnpm run fix` for formatting and linting code and docs


dai-shi:
We decided to drop UMD support in v2. So, it's intentional.
Our of curiosity, what is your use case?

AB498:
> We decided to drop UMD support in v2. So, it's intentional. Our of curiosity, what is your use case?

Usage with umd version of react.

dai-shi:
React also stopped shipping UMD.
https://react.dev/blog/2024/04/25/react-19-upgrade-guide#umd-builds-removed

> To load React 19 with a script tag, we recommend using an ESM-based CDN such as [esm.sh](https://esm.sh/).

I hope it works for you.

PR #1107, opened 2025-06-26
test(*): add 'expect' in all tests

sukvvon:

Fixes #

## Summary

In the official Testing Library documentation:  
https://github.com/testing-library/jest-dom?tab=readme-ov-file#tobeinthedocument

Using `expect(...).toBeInTheDocument()` is one way to make the test’s intent more explicit — to show that the queried element is expected to exist in the DOM.

While `getBy*` queries already fail when elements are not found, using `expect(...).toBeInTheDocument()` makes the assertion explicit, helping future readers understand that presence in the DOM is intentionally being verified — not just implicitly assumed by the query.

## Check List

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -81,7 +81,6 @@ export default tseslint.config(
     rules: {
       'import/extensions': ['error', 'never'],
       '@typescript-eslint/no-unused-vars': 'off',
-      'vitest/expect-expect': 'off',
       'vitest/consistent-test-it': [
         'error',
         { fn: 'it', withinDescribe: 'it' },


dai-shi:
Can you write down the problem and the benefit of this change in the PR description? Also, there are too many similar PRs. Maybe, combine them all?

sukvvon:
@dai-shi 

> Can you write down the problem and the benefit of this change in the PR description?

As suggested in the official Testing Library documentation:  
https://github.com/testing-library/jest-dom?tab=readme-ov-file#tobeinthedocument

Using `expect(...).toBeInTheDocument()` is the recommended way to explicitly assert that an element exists in the DOM.

Some of the existing tests used this assertion, while others only called `screen.getByText(...)` without an explicit assertion, resulting in inconsistent test styles.

This PR updates all relevant tests to follow the recommended style, improving clarity and consistency by making all assertions explicit.

Additionally, as noted in Kent C. Dodds’ blog:
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-get-variants-as-assertions

> “I personally normally keep the assertion in there just to communicate to readers of the code that it’s not just an old query hanging around after a refactor but that I’m explicitly asserting that it exists.”

Although `getBy*` queries throw helpful errors when the element is not found, making the assertion explicit (e.g., `expect(...).toBeInTheDocument()`) signals clearer intent and avoids any confusion, especially during refactors.

> Also, there are too many similar PRs. Maybe, combine them all?

While the purpose of the changes is consistent across files, I separated the PRs by test file to make the review process more manageable.

Even though the changes are repetitive, splitting them up helps maintain context and reduces the risk of missing subtle issues.

dai-shi:
> in the official Testing Library documentation: https://github.com/testing-library/jest-dom?tab=readme-ov-file#tobeinthedocument
> 
> Using `expect(...).toBeInTheDocument()` is one way to make the test’s intent more explicit — to show that the queried element is expected to exist in the DOM.
> 
> While `getBy*` queries already fail when elements are not found, using `expect(...).toBeInTheDocument()` makes the assertion explicit, helping future readers understand that presence in the DOM is intentionally being verified — not just implicitly assumed by the query.

Cool. Please summarize it in the PR description.

> Some of the existing tests used this assertion, while others only called `screen.getByText(...)` without an explicit assertion, resulting in inconsistent test styles.

Yeah, I've wondered the use of findBy without expect and asked once in one of repos. Didn't get alternative answers back then.

> This PR updates all relevant tests to follow a consistent style by making all assertions explicit, improving clarity and maintainability.
> 
> https://github.com/pmndrs/valtio/blob/main/eslint.config.mjs#L84
> 
> We currently have `vitest/expect-expect` turned `off`, but by following the approach I’m proposing, we can remove that override and fully adopt the rules defined in `vitest.configs.recommended`.

Sounds great.

> > Also, there are too many similar PRs. Maybe, combine them all?
> 
> While the purpose of the changes is consistent across files, I separated the PRs by test file to make the review process more manageable.
> 
> Even though the changes are repetitive, splitting them up helps maintain context and reduces the risk of missing subtle issues.

No, I don't want to review all separated PRs. I just want to review one and trust you that you do the right thing. There's no point of accepting one PR and rejecting others. So, single PR would be better.

sukvvon:
> No, I don't want to review all separated PRs. I just want to review one and trust you that you do the right thing. There's no point of accepting one PR and rejecting others. So, single PR would be better.

@dai-shi 

I reflected your opinion.

dai-shi (approved):
Thanks. Can you send the same PR to zustand and jotai please?

dai-shi:
> We currently have `vitest/expect-expect` turned `off`, but by following the approach I’m proposing, we can remove that override and fully adopt the rules defined in `vitest.configs.recommended`.

I forgot to ask it. Can you do this change too in this PR (and in Z/J too)?

dai-shi (approved):

PR #1116, opened 2025-07-08
test(async): add 'expect' with 'toBeInTheDocument'

sukvvon:
## Related Bug Reports or Discussions

Fixes #

## Summary

* relate #1107 
* I missed `async.test.tsx` file

## Check List

- [x] `pnpm run fix` for formatting and linting code and docs

PR #1144, opened 2025-08-03
chore(eslint): correctly apply rules in test files and update related tests

sukvvon:

Fixes #

## Summary

* relate https://github.com/pmndrs/valtio/pull/1116#issuecomment-3050714059

## Check List

### DIFF
--- a/tests/async.test.tsx
+++ b/tests/async.test.tsx
@@ -112,6 +112,7 @@ it.skipIf(typeof use === 'undefined')(
       )
     }
 
+    // eslint-disable-next-line testing-library/no-unnecessary-act
     await act(() =>
       render(
         <StrictMode>
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -9,7 +9,7 @@ import tseslint from 'typescript-eslint'
 
 export default tseslint.config(
   {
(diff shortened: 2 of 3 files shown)


tests/async.test.tsx
```diff
@@ -112,6 +112,7 @@ it.skipIf(typeof use === 'undefined')(
       )
     }
 
+    // eslint-disable-next-line testing-library/no-unnecessary-act
```
sukvvon, line 115: Unlike other tests in `async.test.tsx`, this test uses a `state` extracted as a constant via `valtio`'s `proxy`, which updates asynchronously and requires wrapping with `act` to properly track its changes. Since this is the only test with such behavior in our suite, we disable ESLint for it.

dai-shi:
I would just like a third-party review. It can be a human.

dai-shi (approved):
Thanks for working on it. Is this all done in Valtio?

sukvvon:
> Thanks for working on it. Is this all done in Valtio?

@dai-shi 

I believe that, overall, everything is done.

dai-shi:
Cool. Let's move on to Zustand and Jotai.

PR #1153, opened 2025-08-16
fix(vanilla): snapshot should change with subscription after deleting nested props

pilaoda:

Fixes #1152

## Summary

remove listeners.size check in `ensureVersion` to make sure version changed.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -200,7 +200,7 @@ export function proxy<T extends object>(baseObject: T = {} as T): T {
   }
   let checkVersion = versionHolder[1]
   const ensureVersion = (nextCheckVersion = ++versionHolder[1]) => {
-    if (checkVersion !== nextCheckVersion && !listeners.size) {
+    if (checkVersion !== nextCheckVersion) {
       checkVersion = nextCheckVersion
       propProxyStates.forEach(([propProxyState]) => {
         const propVersion = propProxyState[1](nextCheckVersion)
--- a/tests/subscribe.test.tsx
+++ b/tests/subscribe.test.tsx
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
-import { proxy, ref, subscribe } from 'valtio'
+import { proxy, ref, snapshot, subscribe } from 'valtio'
 import { subscribeKey } from 'valtio/utils'
 
 describe('subscribe', () => {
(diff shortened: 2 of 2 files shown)


dai-shi (approved):
Thanks for the test and the fix.
It's surprising that such a fundamental bug still existed.

pilaoda:
It seems this fix cause a performance degradation, because a deep nested object will call `ensureVersion` recursively many times, Maybe we should figure out a better way to fix this. @dai-shi

dai-shi:
Yes, I know I added `&& !listeners.size` for performance.
Hope we find a better solution.

dai-shi:
If subscribed, notifyUpdate should happen and we should able to skip "ensureVersion". That was the intention of ` && !listeners.size`.

dai-shi:
Okay, I found a solution.

```diff
diff --git a/src/vanilla.ts b/src/vanilla.ts
index 0035053..46cec3b 100644
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -197,9 +197,10 @@ export function proxy<T extends object>(baseObject: T = {} as T): T {
       version = nextVersion
       listeners.forEach((listener) => listener(op, nextVersion))
     }
+    checkVersion = versionHolder[1]
   }
   let checkVersion = versionHolder[1]
-  const ensureVersion = (nextCheckVersion = ++versionHolder[1]) => {
+  const ensureVersion = (nextCheckVersion = versionHolder[1]++) => {
     if (checkVersion !== nextCheckVersion) {
       checkVersion = nextCheckVersion
       propProxyStates.forEach(([propProxyState]) => {
```

dai-shi:
This might be safer:
```diff
diff --git a/src/vanilla.ts b/src/vanilla.ts
index 0035053..91f3b00 100644
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -197,9 +197,10 @@ export function proxy<T extends object>(baseObject: T = {} as T): T {
       version = nextVersion
       listeners.forEach((listener) => listener(op, nextVersion))
     }
+    checkVersion = versionHolder[1]
   }
-  let checkVersion = versionHolder[1]
-  const ensureVersion = (nextCheckVersion = ++versionHolder[1]) => {
+  let checkVersion = versionHolder[1] - 1
+  const ensureVersion = (nextCheckVersion = versionHolder[1]++) => {
     if (checkVersion !== nextCheckVersion) {
       checkVersion = nextCheckVersion
       propProxyStates.forEach(([propProxyState]) => {
```

pilaoda:
This is my current thinking, there may be some misunderstanding.

Solution 1: We just need a `reverse reference map`, for child object to notify parent object to update version.

Solution 2: addPropListener also insert into child `listeners` Set, listen to child object prop update instead of just child object itself.

pilaoda:
I know snapshot is lazy evaluated. Both of my solutions are not conflict with lazy evaluated. I can have a try with solution 2 seems easier.

dai-shi:
> We should be able to write a memory leak test.

👉 #1155 

> Writing performance test would be a challenge.

👉 #1156

PR #1156, opened 2025-08-16
test: performance

dai-shi:
- snapshot with subscription

https://github.com/pmndrs/valtio/pull/1153#issuecomment-3193670394

> Writing performance test would be a challenge.


pilaoda:
Cool! In my project, the `version` is correct, the performance is cool, man you are the hero!

PR #1157, opened 2025-08-16
fix(vanilla): snapshot should change with subscription after deleting nested props (ensureVersion() if listeners size zero)

pilaoda:

Fixes #1152

## Summary

Pre trying fix in #1153 cause performance issue.

Finally I come up with a simple solution:
we just need to `ensureVersion` when `listeners.size` condition changed, that is when `subscribe` or more precisely `addListener` being called.

## Check List


dai-shi (commented):
Oh, this is interesting.
But `ensureVersion` calls recursively which is bad on performance, right?

pilaoda:
Maybe you are right, my test environment can always have `listeners.size` to prevent recursive `ensureVersion`.

PR #1158, opened 2025-08-16
fix(test): make performance test more stable

dai-shi:
I wasn't sure why the test was unstable in #1156.


tests/performance.test.tsx
```diff
@@ -17,22 +17,22 @@ const measurePerformance = (
     times.push(performance.now() - start)
   }
   tearDown()
-  times.slice().sort((a, b) => a - b)
+  times.sort((a, b) => a - b)
```
dai-shi, line 20: It was noop... 😓

PR #1175, opened 2025-11-02
refactor(utils): changes to proxySet for new typescript impl

overthemike:

Fixes #

## Check List

### DIFF
--- a/src/vanilla/utils/proxySet.ts
+++ b/src/vanilla/utils/proxySet.ts
@@ -91,7 +91,7 @@ export function proxySet<T>(initialValues?: Iterable<T> | null) {
     }
   }
 
-  const hasIterator = (o: unknown): o is Iterable<unknown> =>
+  const isIterable = (o: unknown): o is Iterable<unknown> =>
     typeof o === 'object' && o !== null && Symbol.iterator in (o as object)
 
   const hasForEach = <U>(
@@ -100,7 +100,7 @@ export function proxySet<T>(initialValues?: Iterable<T> | null) {
     typeof (o as { forEach?: unknown }).forEach === 'function'
 
   const asIterable = <U>(other: RSetLike<U> | Set<U>): Iterable<U> => {
-    if (hasIterator(other)) return other as Iterable<U>
+    if (isIterable(other)) return other as Iterable<U>
     if (hasForEach(other)) {
       const acc: U[] = []
       other.forEach((v) => acc.push(v))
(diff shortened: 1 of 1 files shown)


src/vanilla/utils/proxySet.ts
```diff
@@ -84,6 +91,98 @@ export function proxySet<T>(initialValues?: Iterable<T> | null) {
     }
   }
 
+  const hasIterator = (o: unknown): o is Iterable<unknown> =>
```
dai-shi: I prefer to name it `isIterable`.

src/vanilla/utils/proxySet.ts
```diff
@@ -130,31 +229,31 @@ export function proxySet<T>(initialValues?: Iterable<T> | null) {
       if (!isProxy(this)) {
         throw new Error('Cannot perform mutations on a snapshot')
       }
-      this.data.length = 0 // empty array
```
dai-shi: Can we keep the comment?

src/vanilla/utils/proxySet.ts
```diff
@@ -130,31 +229,31 @@ export function proxySet<T>(initialValues?: Iterable<T> | null) {
...
+      this.data.length = 0
       this.index = 0
       this.epoch++
       indexMap.clear()
     },
-    forEach(cb) {
-      this.epoch // touch property for tracking
```
dai-shi: Can we keep the comment?

dai-shi (commented):
Thanks for working on this. Comments below.

src/vanilla/utils/proxySet.ts
```diff
@@ -4,18 +4,24 @@ const { proxyStateMap, snapCache } = unstable_getInternalStates()
 const maybeProxify = (x: any) => (typeof x === 'object' ? proxy({ x }).x : x)
 const isProxy = (x: any) => proxyStateMap.has(x)
 
+type RSetLike<T> = { has(value: T): boolean }
+
 type InternalProxySet<T> = Set<T> & {
   data: T[]
-  toJSON: object
+  toJSON: () => object
```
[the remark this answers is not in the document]
overthemike: Actually, this kind of brings up a bit of a question...

When you run `JSON.stringify` on a Set, it will always output an empty object...like so...

```ts
const s = new Set([1,2,3])
console.log(JSON.stringify(s)) // '{}'
```

We did mimic this with proxySet...but is this actually behavior that we want? The reason `JSON.stringify` does this is because of the internal data slots which are the original cause of needing to create proxyMap/Set in the first place. Should we instead actually output the values inside the proxySet instead of just mimicing native `Set` behavior?

```ts
toJSON(): Set<T> {
  return new Set(this.values())
},
```
could be this instead
```ts
toJSON(): Set<T> {
  return Array.from(this.values())
},
```

Thoughts? This would also need to be done with proxyMap if we went that direction.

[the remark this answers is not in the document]
dai-shi: I think the goal here is to mimic `Set`'s behavior, regardless of being useful or not.

PR #1176, opened 2025-11-03
refactor(react): avoid eslint exception

dai-shi:
It passes all the tests, but I'm not very confident if tests are comprehensive.


zardoy:
im pretty sure after this commit I've started noticing in an edge-case (when there is a lot of updates), but an easy-to-reproduce bug leading to desync with just captured snapshot data on init render and actual proxy data, downgrading from 2.2.0 to 2.1.8 seems do the thing (but im not confident either)

I'd like to elaborate, but for me it is extremely hard to debug Valtio, can't get the sense of where to put breakpoints at what time

dai-shi:
I feel like I should revert this. It would be really great if we have a failing test.

PR #1189, opened 2025-12-21
feat(vanilla): make ops callback opt-in with unstable_enableOp

dai-shi:
close #1188

This is a breaking change for an unstable feature.
If a user or a library uses `unstable_ops` in a callback, they have to enable in advance with `unstable_enableOp(true)`.

### DIFF
--- a/src/vanilla/utils/devtools.ts
+++ b/src/vanilla/utils/devtools.ts
@@ -1,8 +1,4 @@
-import {
-  unstable_enableOp as enableOp,
-  snapshot,
-  subscribe,
-} from '../../vanilla.ts'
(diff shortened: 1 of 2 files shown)


dai-shi:
This will be a breaking change for valtio-yjs and valtio-y. 
cc @agcty

dai-shi:
@ScreamZ 
Thanks for your feedback.

No in rush, but could you try some experiments, if performance changes between these?

- unstable_enableOp(true) + using `ops`
- unstable_enableOp(false) + using `deep-object-diff`

My intuition is that if we don't have performance benefit, we should consider dropping `ops` completely. Though, we need to also consider `valtio-yjs` and `valtio-y` that are using `ops` now.

PR #1193, opened 2026-01-10
docs: add how to update values inside arrays

pooya-badiee:
https://github.com/pmndrs/valtio/discussions/755

## Summary
Add documentation on how to update values inside arrays while avoiding unnecessary re-renders.
I tried to keep it short, let me know if I can add anything

## Check List

### DIFF
--- a/docs/how-tos/how-to-update-values-inside-arrays.mdx
+++ b/docs/how-tos/how-to-update-values-inside-arrays.mdx
@@ -77,6 +77,7 @@ function CounterList() {
       <h1>{snap.title}</h1>
       {/* Only the length is accessed from snap, so only updated children re-render */}
       {Array.from({ length: snap.items.length }, (_, index) => (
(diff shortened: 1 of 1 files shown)


overthemike:
This is a fantastic start. I think it might be beneficial to add something at the end that explains _why_ it avoids re-renders. Proxy vs Snapshot is the probably the most difficult to understand for newcomers to library. Maybe something like this?

Because the proxy (in this case: `state.items[index]`) is a `proxy` object instead of a `snapshot` object, accessing it does not track usage. This prevents the parent component from "subscribing" to the item's internal changes, effectively isolating updates to the child component.

pooya-badiee:
Thank you @overthemike. I implemented the feedback. Since the FAQ answers were too close together, I merged them into a 'Why It Works' section.
Let me know if it makes more sense

overthemike (approved):

PR #1209, opened 2026-03-07
test(ssr): add SSR and hydration tests for 'getServerSnapshot'

sukvvon:
## Summary

- Add `ssr.test.tsx` with tests for `getServerSnapshot` callback in `useSnapshot`
- Test initial snapshot and latest snapshot after proxy state changes
- Add hydration mismatch test to verify `getServerSnapshot` prevents hydration error
- Covers previously uncovered line 161 in `src/react.ts`

## Check List

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -1,5 +1,6 @@
 import { StrictMode, useEffect, useLayoutEffect, useState } from 'react'
 import { act, fireEvent, render, screen } from '@testing-library/react'
+import { renderToString } from 'react-dom/server'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { proxy, snapshot, useSnapshot } from 'valtio'
 import { useCommitCount } from './utils'
@@ -11,6 +12,24 @@ describe('basic', () => {
 
   afterEach(() => {
     vi.useRealTimers()
+  })
+
+  it('should return snapshot via getServerSnapshot with renderToString', () => {
+    const obj = proxy({ count: 0 })
+
+    const Counter = () => {
+      const snap = useSnapshot(obj)
+      return <div>count: {snap.count}</div>
+    }
+
+    const view = renderToString(
+      <StrictMode>
+        <Counter />
+      </StrictMode>,
+    )
+
+    expect(view).toContain('count:')
+    expect(view).toContain('0')
   })
 
   it('simple counter', async () => {
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -1,6 +1,5 @@
 import { StrictMode, useEffect, useLayoutEffect, useState } from 'react'
 import { act, fireEvent, render, screen } from '@testing-library/react'
-import { renderToString } from 'react-dom/server'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { proxy, snapshot, useSnapshot } from 'valtio'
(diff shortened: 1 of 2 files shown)


tests/basic.test.tsx
```diff
@@ -1,5 +1,6 @@
 import { StrictMode, useEffect, useLayoutEffect, useState } from 'react'
 import { act, fireEvent, render, screen } from '@testing-library/react'
+import { renderToString } from 'react-dom/server'
```
dai-shi: Let's move this case for a separate file tests/ssr.test.tsx?

tests/basic.test.tsx
```diff
@@ -13,6 +14,24 @@ describe('basic', () => {
...
+    }
+
+    const view = renderToString(
+      <StrictMode>
+        <Counter />
+      </StrictMode>,
+    )
+
+    expect(view).toContain('count:')
+    expect(view).toContain('0')
```
dai-shi: I think this test still passes even if we remove getServerSnapshot.

dai-shi (changes requested):
Unless I don't misunderstand, this is not a valid test.

tests/basic.test.tsx
sukvvon: @dai-shi I verified by removing `getServerSnapshot` — the test fails with `Missing getServerSnapshot, which is required for server-rendered content`. So `renderToString` does invoke `getServerSnapshot`, and this test validates its presence.

dai-shi: Okay, but this test doesn't cover the behavior of getServerSnapshot. For example, if we have the getServerSnapshot implementation identical to that of getSnapshot, the test may pass.

sukvvon: @dai-shi You're right — this test doesn't distinguish the behavior of `getServerSnapshot` from `getSnapshot`. However, it serves as a regression test to ensure `getServerSnapshot` is provided, since `renderToString` throws if it's missing. I also added a test (3aea407) that verifies the latest snapshot is returned after proxy state changes.

dai-shi (changes requested):
I think I've got a better explanation.

The purpose of getServerSnapshot is to avoid hydration error.
So, the test should cover that. Testing if getServerSnapshot is defined is too trivial (only to increase the test coverage without much benefit.)

We've never tested if the current implementation can avoid hydration error.
https://github.com/pmndrs/valtio/blob/383ccf185997cd6ec69ea136bcaccece9d6a7eb7/src/react.ts#L161
If the test reveals it, we need to fix the implementation.

tests/basic.test.tsx
sukvvon: @dai-shi Added a hydration mismatch test (a5c4d42) that verifies `getServerSnapshot` prevents hydration error by rendering with `renderToString` and hydrating with `hydrateRoot`, asserting no `console.error` is called.

tests/ssr.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <StrictMode>
+        <Counter />
+      </StrictMode>,
+    )
+
+    expect(view).toContain('count:')
+    expect(view).toContain('5')
+  })
+
+  it('should not cause hydration mismatch', async () => {
```
dai-shi, line 51: This is too naive. The AI doesn't understand the underlying problem.
Try to create a bug so that such a hydration-mismatch test fails.
I'm not sure if it's possible. But, otherwise, it's not valid for `getServerSnapshot` test.

PR #1211, opened 2026-03-07
test(proxySet): add test for 'intersection' with Set-like object using 'forEach' fallback

sukvvon:
## Summary

- Add test case for `intersection` with a Set-like object that has `forEach` but is not iterable
- Covers previously uncovered `hasForEach` function (line 92) and `forEach` fallback path (lines 96-99) in `src/vanilla/utils/proxySet.ts`

## Check List

### DIFF
--- a/tests/proxySet.test.tsx
+++ b/tests/proxySet.test.tsx
@@ -854,6 +854,24 @@ describe('proxySet', () => {
       expect(result).toEqual(proxySet([1, 9]))
     })
 
+    it('.intersection with Set-like object that has forEach but is not iterable', () => {
+      const odds = proxySet([1, 3, 5, 7, 9])
+      const setLike = {
+        size: 3,
+        has(v: number) {
+          return [1, 4, 9].includes(v)
+        },
+        keys() {
+          return [1, 4, 9].values()
+        },
+        forEach(cb: (v: number) => void) {
+          ;[1, 4, 9].forEach(cb)
+        },
+      }
+      const result = odds.intersection(setLike)
+      expect(result).toEqual(proxySet([1, 9]))
+    })
+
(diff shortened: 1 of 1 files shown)


tests/proxySet.test.tsx
```diff
@@ -854,6 +854,24 @@ describe('proxySet', () => {
       expect(result).toEqual(proxySet([1, 9]))
     })
 
+    it('.intersection with Set-like object that has forEach but is not iterable', () => {
+      const odds = proxySet([1, 3, 5, 7, 9])
+      const setLike = {
```
dai-shi, line 859: What's the reason for this instead of native `Set`?

sukvvon: @dai-shi Native `Set` and `proxySet` have `Symbol.iterator`, so they pass the `isIterable` check and skip the `hasForEach` path (line 92, 96-99). A Set-like object without `Symbol.iterator` is needed to cover the `forEach` fallback in `asIterable`.

dai-shi: I see. I'll defer the review to copilot and @overthemike .

tests/proxySet.test.tsx
```diff
@@ -854,6 +854,24 @@ describe('proxySet', () => {
...
+        size: 3,
+        has(v: number) {
+          return [1, 4, 9].includes(v)
+        },
+        keys() {
+          return [1, 4, 9].values()
+        },
+        forEach(cb: (v: number) => void) {
+          ;[1, 4, 9].forEach(cb)
```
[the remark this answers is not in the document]
dai-shi: Copilot doesn't know prettier. 😢

tests/proxySet.test.tsx
overthemike: Sorry for the wait. I see what you're going for here. Is there a reason we're only testing for interesection?

PR #1220, opened 2026-04-13
fix(vanilla): remove Object.preventExtensions from createSnapshotDefault for Hermes compatibility

manakuro:

Fixes #874 

## Summary
Removes `Object.preventExtensions(snap)` from `createSnapshotDefault` to fix a `TypeError` that occurs on React Native (Hermes).

## Problem

Hermes enforces strict Proxy invariant checks. When a Proxy target is non-extensible, Hermes requires that the `ownKeys` trap returns every key present on the target — any mismatch throws:

```
TypeError: ownKeys target is non-extensible but key is missing from trap result
```

`createSnapshotDefault` calls `Object.preventExtensions(snap)` at the end of snapshot creation. Because `Object.preventExtensions` is irreversible, there is no way to undo it after the fact. The only viable fix is to not call it in the first place.

This is particularly pronounced with arrays (e.g. `InformationItem[]`), since their dynamic key structure (`'0'`, `'1'`, `'length'`, etc.) makes Hermes's invariant check especially strict.

The error surfaces at render time rather than at the point of assignment, because `createSnapshotDefault` is only called when `useSnapshot` runs during the render phase — not when the value is written to the store.

Note: A fix on the Hermes side is not planned. The React Native team has indicated that the `Proxy` API is internally discouraged at Meta. See: https://github.com/facebook/hermes/issues/1609

## Change

- Removed `Object.preventExtensions(snap)` from `createSnapshotDefault`

## Trade-off

Removing `Object.preventExtensions` means snapshots are no longer guaranteed to be non-extensible. New properties could technically be added to a snapshot object at runtime. However, since snapshots are intended to be consumed as read-only values within a render cycle, this is unlikely to cause issues in practice.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -115,6 +115,7 @@ const createSnapshotDefault = <T extends object>(
     }
     Object.defineProperty(snap, key, desc)
   })
+  // Object.preventExtensions is removed. ref: https://github.com/pmndrs/valtio/pull/1220
   return snap
 }
 
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -115,7 +115,7 @@ const createSnapshotDefault = <T extends object>(
     }
     Object.defineProperty(snap, key, desc)
   })
-  return Object.preventExtensions(snap)
+  return snap
 }
 
 const createHandlerDefault = <T extends object>(
--- a/tests/snapshot.test.ts
+++ b/tests/snapshot.test.ts
@@ -24,26 +24,6 @@ describe('snapshot', () => {
     expect(snap1).toBe(snap2)
   })
 
-  it('should make the snapshot immutable', () => {
-    const state = proxy<{ foo: number; bar?: string }>({ foo: 1 })
-    const snap = snapshot(state)
-
-    // Overwriting existing property
-    expect(() => {
-      ;(snap as typeof state).foo = 100
-    }).toThrow()
-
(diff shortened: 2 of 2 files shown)


src/vanilla.ts
```diff
@@ -115,7 +115,7 @@ const createSnapshotDefault = <T extends object>(
     }
     Object.defineProperty(snap, key, desc)
   })
-  return Object.preventExtensions(snap)
+  return snap
```
dai-shi, line 119: Let me leave a comment so that I don't forget.
```suggestion
  // Object.preventExtensions is removed. ref: https://github.com/pmndrs/valtio/pull/1220
  return snap
```

manakuro:
The Vercel deployment is failing, but I don't have permission to view the logs. Could you please take a look?

dai-shi:
> Build Failed
> No Next.js version detected. Make sure your package.json has "next" in either "dependencies" or "devDependencies". Also check your Root Directory setting matches the directory of your package.json file.

It's not your fault obviously.

dai-shi:
Still trying to figure out #1221. But maybe I can merge this without waiting for it.

manakuro:
> Still trying to figure out https://github.com/pmndrs/valtio/pull/1221. But maybe I can merge this without waiting for it.

Thanks! I appreciate your help.

dai-shi (approved):

PR #1224, opened 2026-05-01
chore(deps): update dev dependencies

dai-shi:
for v2.3.2


dai-shi:
```
> valtio@2.3.1 test:types /home/runner/work/valtio/valtio
> tsc --noEmit

Error: src/vanilla/utils/proxyMap.ts(83,9): error TS2322: Type '{ data: V[]; index: number; epoch: number; readonly size: number; get(key: K): V | undefined; has(key: K): boolean; set(key: K, value: V): Map<K, V> & { ...; }; ... 8 more ...; toJSON(): Map<...>; }' is not assignable to type 'InternalProxyObject<K, V>'.
  Type '{ data: V[]; index: number; epoch: number; readonly size: number; get(key: K): V | undefined; has(key: K): boolean; set(key: K, value: V): Map<K, V> & { ...; }; ... 8 more ...; toJSON(): Map<...>; }' is missing the following properties from type 'Map<K, V>': getOrInsert, getOrInsertComputed
 ELIFECYCLE  Command failed with exit code 2.
```

@overthemike Would you be able to help fixing types?

overthemike:
Yeah I'll take a look

overthemike:
@dai-shi So this looks like a bit more than just types. `getOrInsert` and `getOrInsertComputed` are new in the Map interface. I'm not sure yet what they entail as far as implementation goes. We could probably omit those methods for the time being, although we won't be to spec with map anymore.

dai-shi:
Got it. Thanks!

